PR #115, opened 2024-07-26
[WIP] - SDK for lumigator with binary file and wheel to use

binaryaaron:
this is a work in progress of an sdk and CLI for lumigator. it is _partially_ tested and the full SDK isn't done, but if you want to add more to it it should be pretty straightforward from here. I tried to leave good notes in the main module for it. I was testing it with a local notebook.

the new pants targets,`lumigator:lumigator_bin` and `lumigator:lumigator_sdk` are a pex binary and wheel target.

## Pex
Use it like a single-file binary.  i've been using the pex and would recommend using that if you want as a cli tool. 

Build it:

```
pants package lumigator:lumigator_bin
```

Run it via pants (useful if you're debugging) 
```
pants run lumigator:lumigator_bin
```

when you build or run the above target, pants puts it in `./dist/lumigator`. you can run it like

```
lumigator/lumigator_bin.pex --help
```

@veekaybee , @aittalam , @macaab26 - if you launch the app via docker-compose, you can use this file to run commands like:

```
LUMIGATOR_SERVICE_HOST=127.0.0.1 \
RAYCLUSTER_KUBERAY_HEAD_SVC_PORT_8265_TCP_ADDR=localhost \
  dist/lumigator/lumigator_bin.pex groundtruth --list
```

There's a makefile target to do this that's a little buggy - see notes in the makefile - but basically do this if the app isn't running locally:

```
make local-up; sleep 5
make run-sdk-test
```

Run against local via pants:
```
> make local-up
> LUMIGATOR_SERVICE_HOST=127.0.0.1 \
 RAYCLUSTER_KUBERAY_HEAD_SVC_PORT_8265_TCP_ADDR=localhost  \
    pants run lumigator:lumigator_bin -- groundtruth --list
```

## Wheel
you can use the wheel and import it in a nb (!!). It's not perfect; the sources need to get changed a bit before publishing to pypi or something but you can build and use the cli like:

```
pants package lumigator:lumigator_sdk
pip install  dist/lumigator_sdk-0.1-py311-none-any.whl
python -m mzai.schemas.main --help
```

or import it and use `main` like

```
from mzai.schemas.main import LumigatorClient

lc = LumigatorClient(ray_host=<ray host>, api_host=<api host>, s3_base_path=<s3>)
lc.experiments.info()

```


chainlink:
Diff: https://github.com/mozilla-ai/lumigator/compare/main...agonzales/a-real-sdk

PR #201, opened 2024-09-09
[LUMI-21] Make required fields as `required` in chart

javiermtorres:
Chart values that are needed for the Deployment env vars should be made `required`, and show an error if they are null-valued (following the Go template definition).

### DIFF
--- a/lumigator/infra/mzai/helm/lumigator/templates/deployment.yaml
+++ b/lumigator/infra/mzai/helm/lumigator/templates/deployment.yaml
@@ -29,7 +29,7 @@ spec:
         - name: {{ .Chart.Name }}
           securityContext:
             {{- toYaml .Values.securityContext | nindent 12 }}
-          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          image: "{{ required "The repository for the Lumigator image is missing" .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           ports:
             - name: http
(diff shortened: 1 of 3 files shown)


lumigator/infra/mzai/helm/lumigator/templates/deployment.yaml
```diff
@@ -37,25 +37,25 @@ spec:
               protocol: TCP
           env:
```
macaab26, line 38: Do you only add it for the env vars? Is there a reason why?

chainlink:
Tag and repository are also required

chainlink (approved):
Added command on repo and tag, but otherwise pre-approving

javiermtorres:
> Tag and repository are also required

But shouldn't we have default values for these in the `values.yaml` file? The user wouldn't need to add these in their own values.

macaab26:
FYI, I created a [story](https://mzai.atlassian.net/browse/LUMI-52) to check all the default values for the variables, and also check if we miss some required variable.

macaab26 (approved):

agpituk (approved):

agpituk:
> > Tag and repository are also required

> But shouldn't we have default values for these in the `values.yaml` file? The user shouldn't need to add these in their own values.

I'm in favor of what Javier says, but, in that case, shouldn't we do add a default for the repository as well?

agpituk:
> > > Tag and repository are also required
> 
> > But shouldn't we have default values for these in the `values.yaml` file? The user shouldn't need to add these in their own values.
> 
> I'm in favor of what Javier says, but, in that case, shouldn't we do add a default for the repository as well?

So I just had a chat with Mario which highlighted a few [things](https://mzlai.slack.com/archives/C07LJP9Q5KJ/p1726217228011579?thread_ts=1726209826.507089&cid=C07LJP9Q5KJ). After that I think required + defaults is the way to go

PR #202, opened 2024-09-09
[LUMI-16] Allow a secret for the Mistral API to be auto-generated if needed

javiermtorres:
Lumigator needs the `MISTRAL_API_KEY` environment variable if the Mistral API is to be used. The user can provide either of these properties in the `values.yaml` file to set the variable:

* The name of a `Secret` in property `mistralSecretName`
* The Mistral API key directly in property `mistralKey`; in this case, a Secret will also be generated by the template

If neither is set, an error will be shown.


lumigator/infra/mzai/helm/lumigator/test/custom-mistral-key.yaml
```diff
@@ -0,0 +1 @@
+mistralKey: abcdefgh
```
macaab26, line 1: I think you added the files in this folder by mistake, could you check it?

lumigator/infra/mzai/helm/lumigator/templates/deployment.yaml
```diff
@@ -57,13 +58,14 @@ spec:
...
-            {{- if .Values.mistralSecretName }}
-            - name: MISTRAL_API_KEY # kubectl create secret generic mistral-key --from-literal MISTRAL_API_KEY=value
+            - name: MISTRAL_API_KEY
               valueFrom:
                 secretKeyRef:
-                  name: {{ .Values.mistralSecretName }}
-                  key: {{ .Values.mistralSecretKey }}
-            {{- end }}
+                  {{- /* Note that a one-line comments needs THIS precise layout {{- }} to avoid breaking flow... */}}
+                  {{- /* Note the lack of - at the if and end on each line */}}
```
macaab26, line 65: Is this really needed?

lumigator/infra/mzai/helm/lumigator/templates/gen_secrets.yaml
```diff
@@ -0,0 +1,13 @@
+{{- include "lumigator.consts" . -}}
+{{- if not .Values.mistralSecretName -}}
+{{- if not .Values.mistralKey -}}
+{{- fail "Neither an existing Mistral API Secret name nor Mistral API key details have been provided" -}}
```
macaab26, line 4: For me, it's not an error to not include a secret with the key or the key instead. Maybe some users don't have a Mistral API key because they don't use it

lumigator/infra/mzai/helm/lumigator/test/custom-mistral-key.yaml
javiermtorres: Hmm, I'm using this directory to hold test values for the moment. The idea is that the test tool for helm templates will use these values files. But if you feel that these test values files are better placed somewhere else, just let me know.

You can check for example that `helm template -f test/custom-mistral-key.yaml` will generate the secret whereas the one with the Secret name will not.

lumigator/infra/mzai/helm/lumigator/templates/deployment.yaml
javiermtorres: Given the amount of time that I have spent fixing indentation and whitespacing in helm templates, I think it is 😆 I'd put it in some doc page too; maybe we can remove it from here at this point. But helm template indentation is imho tricky enough that a newcomer may find these comments helpful.

lumigator/infra/mzai/helm/lumigator/templates/gen_secrets.yaml
javiermtorres: Hmmmm, good point. In that case, should we add the MISTRAL_API_KEY to the pod environment, or should we skip it?

lumigator/infra/mzai/helm/lumigator/templates/deployment.yaml
macaab26: Yep, agree with adding it in Confluence, but not in the code. I would rather have it as documentation, not as a comment.

lumigator/infra/mzai/helm/lumigator/templates/gen_secrets.yaml
macaab26: The actual behavior is not include it. But I'm not 100% sure that we expect that in the code. [Here](https://github.com/mozilla-ai/lumigator/blob/main/lumigator/python/mzai/backend/services/completions.py#L19) is where it's used, and we rely in the library to rise the error (if I'm not wrong). Maybe @veekaybee can help us with this 😄

macaab26:
@javiermtorres please, add some text in the README describing the new behavior and the options that the user has.

lumigator/infra/mzai/helm/lumigator/templates/gen_secrets.yaml
veekaybee: We expect the Mistral completion service to work when the user is interested in performing a ground truth evaluation using the Mistral API, that is it's entirely open to the user if they want to use the component or not. 

Right now this passed at the "SDK", aka [notebook level](https://github.com/mozilla-ai/lumigator/blob/4ff0acffecbcdd5a9394e3097f1a6ed2c45bd4c5/notebooks/lumigator_demo.py#L342). So not having the key will not stop the app from working, and if you want to perform an eval, you will need to inject a key at eval time. How we implement this depends more on the SDK side.  Generally in the LLM space, the expected user behavior is for the user to provide the key in the application they're developing by injecting the token. As an example: 

https://github.com/run-llama/llama_index?tab=readme-ov-file#-example-usage

lumigator/infra/mzai/helm/lumigator/test/custom-mistral-key.yaml
chainlink: agree with adding to a separate folder (so it doesn't get shipped with the chart

lumigator/infra/mzai/helm/lumigator/templates/deployment.yaml
chainlink: It might be cleaner to define this in `_helpers` and call it here

chainlink: To Mario's point, I think some of this could be captured in confluence,  https://mzai.atlassian.net/wiki/spaces/SD/pages/13271041/Helm+Best+Practices Specifically the callouts on using the dash `-` with `{{` and `}}` as it's more of a general helm-ism.

lumigator/infra/mzai/helm/lumigator/test/custom-mistral-key.yaml
javiermtorres: Ok, removing these files right now

lumigator/infra/mzai/helm/lumigator/templates/deployment.yaml
javiermtorres: > It might be cleaner to define this in `_helpers` and call it here

I'd usually agree, but the lack of proper functions in Helm and being a one line (however long) make me prefer having this piece here.

lumigator/infra/mzai/helm/lumigator/templates/gen_secrets.yaml
javiermtorres: Ok, I can try to include some way of signalling that the Mistral API key is not used (post install hook?) so the user is aware that this is not available.

macaab26: For me it's enough with what you did @javiermtorres, great work!

lumigator/infra/mzai/helm/lumigator/templates/_helpers.tpl
```diff
@@ -49,3 +49,8 @@ Selector labels
 app.kubernetes.io/name: {{ include "lumigator.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
+
+{{- define "lumigator.consts" -}}
```
chainlink, line 53: This is an interesting approach, but I think it might make sense to use a 'regular' helper variable here

lumigator/infra/mzai/helm/lumigator/values.yaml
```diff
@@ -20,8 +20,8 @@ serviceAccountName: ""
 
 summarizerWorkDir: ""
 
-mistralSecretName: mistral-key
-mistralSecretKey: MISTRAL_API_KEY
+mistralSecretName: ""
```
chainlink, line 23: Perhaps change Mistralsecretname to existingMistralSecret to make it more clear

lumigator/infra/mzai/helm/lumigator/values.yaml
```diff
@@ -20,8 +20,8 @@ serviceAccountName: ""
...
+mistralKey: ""
```
chainlink, line 24: Change to 'mistralAPIKey'

chainlink (changes requested):

lumigator/infra/mzai/helm/lumigator/templates/gen_secrets.yaml
```diff
@@ -0,0 +1,13 @@
+{{- include "lumigator.consts" . -}}
+{{- if .Values.mistralKey -}}
+{{- if .Values.mistralSecretName -}}
+{{- fail "A Mistral key and a Secret name cannot be provided at the same time" -}}
+{{- end -}}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "lumigator.fullname" . }}
```
chainlink, line 9: the secret name should be `{{ include "lumigator.fullname" . }}-mistral`

chainlink: Also please include:

```
  labels:
    {{- include "lumigator.labels" . | nindent 4 }}
```

lumigator/infra/mzai/helm/lumigator/templates/_helpers.tpl
chainlink: We could also use `envFrom` and avoid having to set this variable in two places https://gist.github.com/troyharvey/4506472732157221e04c6b15e3b3f094

javiermtorres: The issue with regular variables is that the value will need to be defined in each template, since variables don't cross the `include` scope afaict. This ensures that the same value is used in both templates (gen_secret and deployment). I'm using a special root property here because I don't want values from Values to leak into our constants.

Another trick I tend to use instead of setting the values manually is doing a `fromJson` or `fromYaml` for constants and then assigning that to the `.Consts` property, although that one is trickier for the new reader.

lumigator/infra/mzai/helm/lumigator/values.yaml
javiermtorres: Ok

lumigator/infra/mzai/helm/lumigator/values.yaml
javiermtorres: Ok

lumigator/infra/mzai/helm/lumigator/templates/gen_secrets.yaml
javiermtorres: Ok to both

lumigator/infra/mzai/helm/lumigator/templates/_helpers.tpl
javiermtorres: If you mean using a define as, well, as a define, _my_ issue with the define/include mechanism in Helm is that it mixes dealing with values and dealing with format. In this way the value (with its proper type and structure, in case of maps for example) is set in one place, and then the receiver can decide how to best format this. Including simple snippets is ok with the base mechanism, but for example if we want to start adding constants or "scoped" properties (`mistral` as a map with subproperties) it gets awkward with even the second or third property.

If you think a plain define is clearer in this case, I'll replace it.

lumigator/infra/mzai/helm/lumigator/templates/deployment.yaml
macaab26: I maintain that this type of information should be in Confluence not in the code

javiermtorres: Yes, it should be removed (if it's not, I just forgot, sorry in advance...). What I mean is...

> It might be cleaner to define this in _helpers and call it here

Which I'm *assuming* to be about https://github.com/mozilla-ai/lumigator/blob/20382c07ad36146529aab4b40a3d392f145e33b5/lumigator/infra/mzai/helm/lumigator/templates/deployment.yaml#L67. But maybe I got this wrong 🤔

chainlink: Yep, that one line probably should be a template helper since it's quite long. Within a given manifest it's better to stick with functions/pipelines, or calling named templates.

javiermtorres: Ok, an include should be fine here.

chainlink (approved):
Pre-approving. Last piece is to rename the file `gen_secrets.yaml` to `mistral-secret.yaml`

PR #206, opened 2024-09-11
Move the postgres password into a secret and allow for passing an existing secret

chainlink:


lumigator/infra/mzai/helm/lumigator/templates/_helpers.tpl
```diff
@@ -49,3 +49,14 @@ Selector labels
 app.kubernetes.io/name: {{ include "lumigator.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
+
+{{/*
+Postgresql password secret name
+*/}}
+{{- define "lumigator.postgresqlPasswordSecretName" -}}
+{{- if .Values.existingPostgresPasswordSecretName }}
```
javiermtorres, line 57: What about using default directly?

lumigator/infra/mzai/helm/lumigator/templates/db-secret.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.postgresPassword -}}
```
javiermtorres, line 1: We could make this as a define, and add the needed values (pwd, existing secret name, key and env name, etc) in the dict params. And then we have the one for OAPI too ^_^

lumigator/infra/mzai/helm/lumigator/templates/_helpers.tpl
javiermtorres: Yes, but I mean, is it different from using the `default` function from Helm?

chainlink: Yeah, I could use `default` or `ternary` here

lumigator/infra/mzai/helm/lumigator/templates/db-secret.yaml
chainlink: Sorry, I don't follow. Could you provide a code example?

javiermtorres: Basically my Mistral secret is:
```
{{- include "lumigator.mistral-default-secret" . -}}
{{- if .Values.mistralAPIKey -}}
{{- if .Values.existingMistralAPISecret -}}
{{- fail "A Mistral key and an existing Mistral Secret name cannot be provided at the same time" -}}
{{- end -}}
apiVersion: v1
kind: Secret
metadata:
  name: {{ include "lumigator.mistral-secret-name" . }}
  labels:
    {{- include "lumigator.labels" . | nindent 4 }}
type: Opaque
data:
  {{ .Consts.mistralSecretKey }}: {{ .Values.mistralAPIKey | b64enc | quote }}
{{- end -}}
```
Which is quite similar. Maybe we could make a template like this...
```
{{ define "lumigator-secret" }}
{{- if .secretToken -}}
{{- if .existingSecretName -}}
{{- fail .onlyOneMsg -}}
{{- end -}}
apiVersion: v1
kind: Secret
metadata:
  name: .secretName
  labels:
    {{- include "lumigator.labels" . | nindent 4 }}
type: Opaque
data:
  {{ .secretTokenKey }}: {{ .secretTokenValue | b64enc | quote }}
{{- end -}}
```
And then use it like...
```
{{- include lumigator-secret (dict "secretToken" .Values.postgresPassword ".existingSecretName" .Values.existingPostgresPasswordSecretName "secretTokenKey" .Values.postgresPasswordSecretKey "secretTokenValue" .Values.mistralAPIKey "secretName" (include "lumigator.postgresqlPasswordSecretName" .) ) -}}
```
(Fixing details, I'd need to check the includes there). As I mentioned, one could also do:
```
{{ define "postgress-data" }}
secretToken: {{ .Values.postgresPassword }}
existingSecretName: {{ .Values.existingPostgresPasswordSecretName }}
secretTokenKey: {{ .Values.postgresPasswordSecretKey }}
secretTokenValue: {{ .Values.mistralAPIKey }}
secretName: {{ include "lumigator.postgresqlPasswordSecretName" . }}
{{ end }}
{{ $params := fromYaml (include "postgress-data" .) }}
{{- include lumigator-secret $params }}
```
Of course, this depends on how much we want to keep templates consistent. Maybe the secrets will not be homogeneous, maybe they will so we'd need to make the same changes across all secrets.

javiermtorres: If you see this as too complicated, then just resolve the issue and keep the secret as it is 👍

agpituk (approved):

lumigator/infra/mzai/helm/lumigator/values.yaml
```diff
@@ -10,8 +10,18 @@ postgresDb: ""
...
 postgresPort: ""
 postgresUser: ""
+
+# The password of the postgres database to connect to.
+# A secret will be created for you.
+# Set this OR 'existingPostgresPasswordSecretName'
 postgresPassword: ""
 
+# Use an existing secret, set this value OR use 'postgresPassword'
+existingPostgresPasswordSecretName: ""
```
macaab26, line 20: We should clarify that the user has to use only of the two options (provide the key or the secret), like we did with the [keys](https://github.com/mozilla-ai/lumigator/blob/main/lumigator/infra/mzai/helm/lumigator/README.md#mistral-api-key-management)

PR #208, opened 2024-09-12
[LUMI-19] Coreweave s3 support

agpituk:
- Modified settings.py to support s3,coreweave and localstack:
Currently if `AWS_HOST` value is set to any string our settings.py will concatenate `AWS_HOST` and `AWS_PORT`. This concatenation is only required for localstack(afaik) so the change I propose is to make it explicit, because to use coreweave s3 storage we need to set AWS_HOST to a specific coreweave URL 
 - Added Env Vars to the helm chart
 - Download doesn't fully work, but it's related to a bug that's being worked on in [207](https://github.com/mozilla-ai/lumigator/pull/207) . I've checked that the URL is created with the right format for coreweave
 - Currently S3_ENDPOINT_URL and AWS_ENDPOINT_URL. have the default same values, but they are used differently, one is passed onto ray and another one used to connect to s3 in lumigator, but I think that work should be done in [this](https://mzai.atlassian.net/browse/LUMI-53) ticket

### DIFF
--- a/.devcontainer/docker-compose-local.yaml
+++ b/.devcontainer/docker-compose-local.yaml
@@ -79,12 +79,11 @@ services:
       - POSTGRES_USER=admin
       - POSTGRES_PASSWORD=password
       - POSTGRES_DB=lumigator
-      - AWS_HOST=localstack
+      - S3_ENDPOINT_URL=http://localstack:4566
       - AWS_ACCESS_KEY_ID=test
       - AWS_SECRET_ACCESS_KEY=test
       - AWS_DEFAULT_REGION=us-east-2
       - AWS_ENDPOINT_URL=http://localstack:4566
-      - S3_PORT=4566
       - S3_BUCKET=lumigator-storage
       - PYTHONPATH=/mzai/lumigator/python/mzai:/mzai/lumigator/python
       - RAY_DASHBOARD_PORT=8265
--- a/lumigator/python/mzai/backend/settings.py
+++ b/lumigator/python/mzai/backend/settings.py
@@ -18,8 +18,7 @@ class BackendSettings(BaseSettings):
     POSTGRES_DB: str | None = None
 
     # AWS
-    AWS_HOST: str | None = None
-    S3_PORT: int = 4566
+    S3_ENDPOINT_URL: str | None = None
     S3_BUCKET: str = "lumigator-storage"
     S3_URL_EXPIRATION: int = 3600  # Time in seconds for pre-signed url expiration
     S3_DATASETS_PREFIX: str = "datasets"
@@ -53,13 +52,6 @@ class BackendSettings(BaseSettings):
 
     # Summarizer
     SUMMARIZER_WORK_DIR: str | None = None
-
-    @computed_field
-    @property
-    def S3_ENDPOINT_URL(self) -> str | None:  # noqa: N802
-        # Live AWS doesn't require this but LocalStack testing does, so it's optional
-        if self.AWS_HOST == "localstack":
-            return f"http://{self.AWS_HOST}:{self.S3_PORT}"
 
     @computed_field
     @property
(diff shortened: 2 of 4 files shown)


lumigator/python/mzai/backend/settings.py
```diff
@@ -58,7 +58,7 @@ class BackendSettings(BaseSettings):
     @property
     def S3_ENDPOINT_URL(self) -> str | None:  # noqa: N802
         # Live AWS doesn't require this but LocalStack testing does, so it's optional
-        if self.AWS_HOST is not None:
+        if self.AWS_HOST == "localstack":
```
chainlink, line 61: Why the change here?

agpituk: hi @chainlink  I left it as a draft because I planned to add more context and a readme update! Currently if you write anything in the AWS_HOST variable it will add the port
```
            return f"http://{self.AWS_HOST}:{self.S3_PORT}"
```
Because we have to actually set the AWS_HOST to `object.lga1.coreweave.com` (or whatever region/endpoint we use when we create the buckets) when using coreweave it would go into that if statement and add the port. I changed so when you set AWS_HOST to localstack, is when you actually need it and it will add the port

agpituk: I was just thinking. Probably I'll add context into the ticket (so we have it documented) but avoid doing any updates into the repo as Coreweave is very specific to our use case. With the current configuration and values in the docker-compose-local.yaml localstack works out of the box, so probably no other changes needed here

chainlink: I think we can actually simplify this a bunch.

I believe for Coreweave, we'll need to use `https` and no port (ie. `https://object.lga1.coreweave.com`

The `S3_ENDPOINT_URL` variable is used to setup the [s3 client](https://github.com/mozilla-ai/lumigator/blob/helm-unit-test/lumigator/python/mzai/backend/api/deps.py#L31)

The only use of `AWS_HOST` and `S3_PORT` is to compute this value.

I propose the following
- Change `S3_ENDPOINT_URL` to be a regular variable in settings that gets passed in directly via env var (ie. not computed)

- Update the env vars in [docker-compose-local.yaml](https://github.com/mozilla-ai/lumigator/blob/helm-unit-test/.devcontainer/docker-compose-local.yaml#L75) with the full `S3_ENDPOINT_URL` for localstack `http://localstack:4566`

- Remove all references to `AWS_HOST` and `S3_PORT`

There's some other AWS ENV Var cleanup we can do, but I'll add that to another ticket

agpituk: That sounds like a good idea! I'll make the changes and ask for approval again!

agpituk: Changes applied & more context added to the PR description, ready for review!

chainlink (approved):
LGTM!

.devcontainer/docker-compose-local.yaml
```diff
@@ -79,12 +79,11 @@ services:
       - POSTGRES_USER=admin
       - POSTGRES_PASSWORD=password
       - POSTGRES_DB=lumigator
-      - AWS_HOST=localstack
+      - S3_ENDPOINT_URL=http://localstack:4566
       - AWS_ACCESS_KEY_ID=test
       - AWS_SECRET_ACCESS_KEY=test
       - AWS_DEFAULT_REGION=us-east-2
       - AWS_ENDPOINT_URL=http://localstack:4566
```
veekaybee, line 86: We have S3_ENDPOINT_URL and AWS_ENDPOINT_URL pointing to the same address, are these different variables?

veekaybee (approved):
Left a question around S3 URL for the env variables, otherwise looks good to go once that's addressed

.devcontainer/docker-compose-local.yaml
veekaybee: Yup, works for me and serves me right for not re-reading 🤦‍♀️

agpituk: to be fair it was a very long description :D I'll keep them smaller if possible! thanks!!

veekaybee: No, no, this is good context and will be good to keep for reference, please keep writing these 🙏

PR #209, opened 2024-09-13
Direct port of lm-buddy to Lumigator

veekaybee:
## What's changing

[lm-buddy](https://github.com/mozilla-ai/lm-buddy) is a Python library that we developed to be a collection of jobs for fine-tuning and evaluating open-source (large) language models. The library makes use of YAML-based configuration files as inputs to CLI commands for each job and provides a uniform interface to access different sources of data and models.

lm-buddy was initially developed to fill a need for finetuning and evaluation, and we are continuing to use the evaluation component as a core for Lumigator. However, as a result of starting on Lumigator work in parallel, we now have a fair amount code duplication. There are a number of issues we could resolve by merging lm-buddy code into Lumigator. 

**1. Simplifying our dependency chain:** Right now the lm-buddy code is shipped as a pip dependency (i.e. for every new lm-buddy release, we build/publish a package on pypi and this is downloaded at setup time within Lumigator. ) We wouldn't need to do this anymore with lm-buddy contained within lumigator and shipped with the repo as-is. 
2. **Remove redundant components:** we have a chance to make lm-buddy simpler and smaller as a part of this port, and as a result, take up less of a footprint within lumgiator as we ship it. We no longer use finetuning, so can remove all these components. We additionally can probably remove ragas evaluation + prometheus, but will leave this for discussion. 
3. **Remove redundant routes:** Duplication of inference against OpenAI/Mistral for evaluation: we do this within Lumigator via ground truth and in lm-buddy via offline evaluation. We have a chance to fold these two code pieces together. 

There are several pieces to this change: 

1.  A straight port of lm-buddy as part of the lumigator backend to make sure all functionality still works.  <<- this PR
1a. Investigate removing ability to use CLI with lm-buddy and call modules directly as part of simplification
2. Removing redundant components <-- will be addressed once this PR is approved
3. Removing redundant routes.  <-- will be addressed once this PR is approved
4. Unit tests and documentation  <-- will be addressed once this PR is approved (will need to modify tests to run lm-buddy in addition to `pants test lumigator/python/mzai/backend/tests:backend_tests@parametrize=linux_cpu`)

## How to test it

To test this branch, pull it and run:

1. `make local-up` and rebuild Lumigator
2.  Add a new sample dataset using the OpenAPI docs page endpoint or via cURL or via notebook helper commands. 
That dataset will have a UUID that you can copy.  Screenshot 2024-09-13 at 1 31 06 PM.

3. Run the following within a new Jupyter notebook connected to the local environment:

```python
import io
import os
from pathlib import Path
from typing import Any, Dict  # noqa: UP035
from uuid import UUID
import json
import numpy as np

import pandas as pd
import requests

# APP URL
API_HOST = os.environ["LUMIGATOR_SERVICE_HOST"]
API_URL = f"http://{API_HOST}/api/v1"

# Ray URL
RAY_HEAD_HOST = os.environ["RAYCLUSTER_KUBERAY_HEAD_SVC_PORT_8265_TCP_ADDR"]
RAY_SERVER_URL = f"http://{RAY_HEAD_HOST}:8265"

# base S3 path
S3_BASE_PATH = "lumigator-storage/experiments/results/"


# EVAL_METRICS is a dict we use to show / refer to eval metrics
# in pandas dataframes and when filtering by metric name.
# The dict format is
# "column name": [list of keys to get val in nested results dict]
EVAL_METRICS = {
    "Meteor": ["meteor", "meteor_mean"],
    "BERT Precision": ["bertscore", "precision_mean"],
    "BERT Recall": ["bertscore", "recall_mean"],
    "BERT F1": ["bertscore", "f1_mean"],
    "ROUGE-1": ["rouge", "rouge1_mean"],
    "ROUGE-2": ["rouge", "rouge2_mean"],
    "ROUGE-L": ["rouge", "rougeL_mean"],
}

# the following file contains information about different models
# (currently just max RAM consumption on the TB inference task)
MODEL_INFO_FILE = "assets/model_info.csv"

def make_request(
    url: str,
    method: str = "GET",
    params: Dict[str, Any] = None,  # noqa: UP006
    data: Dict[str, Any] = None,  # noqa: UP006
    files: Dict[str, Any] = None,  # noqa: UP006
    headers: Dict[str, str] = None,  # noqa: UP006
    json_: Dict[str, str] = None,  # noqa: UP006
    timeout: int = 10,
    verbose: bool = True,
    *args,
    **kwargs,
) -> requests.Response:
    """Make an HTTP request using the requests library.

    Args:
        url (str)
        method (str, optional): The HTTP method to use. Defaults to "GET".
        params (Dict[str, Any], optional): URL parameters to include in the request.
        data (Dict[str, Any], optional): Data to send in the request body.
        files (Dict[str, Any], optional): Files to send in the request body.
        headers (Dict[str, str], optional): Headers to include in the request.
        timeout (int, optional): Timeout for the request in seconds. Defaults to 10.

    Returns:
        requests.Response: The response object from the request.

    Raises:
        requests.RequestException
    """
    try:
        response = requests.request(
            method=method.upper(),
            url=url,
            params=params,
            data=data,
            files=files,
            headers=headers,
            timeout=timeout,
            json=json_,
            *args,
            **kwargs,  # noqa: B026
        )
        response.raise_for_status()
        if verbose:
            print(f"{json.dumps(response.json(), indent = 2)}")
    except requests.RequestException as e:
        print(f"Request failed: {e}")
        raise
    return response


def experiments_submit(
    model_name: str,
    name: str,
    description: str | None,
    dataset_id: UUID,
    max_samples: int | None = None,
    system_prompt: str | None = None,
) -> requests.Response:
    if system_prompt is None and (
        model_name.startswith("oai://") or model_name.startswith("http://")
    ):
        system_prompt = "You are a helpful assistant, expert in text summarization. For every prompt you receive, provide a summary of its contents in at most two sentences."  # noqa: E501

    payload = {
        "name": name,
        "description": description,
        "model": model_name,
        "dataset": dataset_id,
        "max_samples": max_samples,
        "system_prompt": system_prompt,
    }

    response = make_request(f"{API_URL}/experiments", method="POST", data=json.dumps(payload))
    return response
 
 # set this value to limit the evaluation to the first max_samples items (0=all)
max_samples = 0
# or your model of choice from the choices
# https://github.com/mozilla-ai/lumigator/blob/273f3ba0ec042fbed4f532ce535c7e855d3cc862/lumigator/python/mzai/backend/config_templates.py#L64
model = 'oai://gpt-4o-mini' 
dataset_id = "dbe05bc5-4956-43fb-9afc-f2df46e35cc3" # the ID of the dataset you uploaded

# now look for the dataset on lumigator
descr = f"Testing {model} summarization model on your csv"
experiments_submit(model, "lumi", descr, dataset_id, max_samples)
```

## Related Jira Ticket

https://mzai.atlassian.net/browse/LUMI-45

This will be part of a series of PRs that we approve sequentially as part of the larger change. 

## I already...

cc @ividal @macaab26 @javiermtorres @agpituk @davidmanzanoai


.devcontainer/docker-compose-local.yaml
```diff
@@ -58,7 +58,8 @@ services:
       # NOTE: the following path contains a hash which needs to be updated whenever
       # the list of libraries installed in backend/services/experiments.py changes.
       # This is a temporary solution, see https://github.com/mozilla-ai/lumigator/pull/186
-      - LD_PRELOAD=/tmp/ray/session_latest/runtime_resources/pip/96de31139e548d3e8200d1e4b2575b77d5695312/virtualenv/lib/python3.11/site-packages/scikit_learn.libs/libgomp-d22c30c5.so.1.0.0
+      # Need to change hash if pip installs change, hash generated from Ray logs
+      - LD_PRELOAD=/tmp/ray/session_latest/runtime_resources/pip/2b258edbe79c9be2bb813e301619caa20219d310/virtualenv/lib/python3.11/site-packages/scikit_learn.libs/libgomp-d22c30c5.so.1.0.0
```
veekaybee, line 62: it turns out we still need this because we import scikit, but I bumped the hash

lumigator/python/mzai/backend/jobs/submission.py
```diff
@@ -25,9 +24,8 @@ class RayJobEntrypoint(ABC):
 
     @property
     def command(self) -> str:
-        """The command ran by the ray job, which is passed a JSON-serialized config."""
-        return f"lm-buddy evaluate huggingface --config '{json.dumps(self.config.args)}'"
-
+        """The command passed to the Ray job, which includes a JSON-serialized config. We currently use lm-buddy as a built-in."""
+        return f"python -m lumigator.python.mzai.backend.lm_buddy evaluate huggingface --config '{json.dumps(self.config.args)}'"
```
veekaybee, line 28: Note the change here to executing the cli module at the root of lm-buddy locally within the lumigator at this point

lumigator/python/mzai/backend/services/experiments.py
```diff
@@ -135,10 +135,9 @@ def create_experiment(self, request: ExperimentCreate) -> ExperimentResponse:
         else:
             worker_gpus = float(os.environ.get(settings.RAY_WORKER_GPUS_ENV_VAR, 1.0))
 
-        # NOTE: Whenever the list of libraries below change, the LD_PRELOAD line in
-        # `docker-compose-local.yaml` has to be updated with a new hash value
         runtime_env = {
-            "pip": ["nltk==3.8.1", "lm-buddy[jobs]==0.12.1"],
+            "pip": ["wandb", "evaluate","pydantic_yaml", "torch", "peft","s3fs","loguru","mistralai","openai","lm-eval","langchain_community","langchain_openai","ragas","trl"],
```
veekaybee, line 139: this is fairly clunky at this point - we can pass a requirements.txt file, and I know we need to add nltk back in, but I was wondering if people thought we should have these reqs as part of the base requirements of lumigator or keep them separately specifically for this module.

veekaybee: I also notice a lot of calls we're doing for components I don't think we need anymore, including langchain and ragas here

veekaybee:
Note that tests are currently failing - will start fixing but still happy to take first-pass comments at this point.

lumigator/python/mzai/backend/services/experiments.py
veekaybee: Note that we'll no longer need `"lm-buddy[jobs]==0.12.1"` 🎉

ruff.toml
```diff
@@ -70,7 +70,15 @@ ignore = [
   "N817", # imported as acryonym
   "B023", # doesn't bind loop var, we do this a lot in torch
   "D100", # module-level docstrings
+  "D417", # documentation for every function parameter.
+  "N806", # ignore uppercased variables
+  "N812", # import as uppercased
+  "N803", # lowercased args
   "N805", # first param needs to be self; pydantic breaks this sometimes
+  "D415", # First line should end with a period, question mark, or exclamation point
```
veekaybee, line 78: added some of these to bypass numerous changes on import from lm-buddy, cross-checked with the ruff file there as well.

lumigator/python/mzai/backend/lm_buddy/__init__.py
```diff
@@ -0,0 +1,3 @@
+from lumigator.python.mzai.backend.lm_buddy.buddy import LMBuddy
```
chainlink, line 1: I think we should move this out of the `backend` folder and make it it's own module so we can have independent deps between the backend API and lm-buddy jobs.

veekaybee: That works for me, initially I debated between having it at `lumigator.python.mzai.backend` and `lumigator.python.mzai` like we did for the summarizer and felt that where the summarizer was, was in a very weird place in relation to the code, but perhaps it makes more sense logically since lm-buddy is larger and "further" away functionality-wise

veekaybee: Another question is whether we see this retaining the lm-buddy name. I'm happy to keep it but feel like it might be less confusing to be named evaluator or something more in line with the names of lumigator components.

chainlink: I'm also good to re-name

lumigator/python/mzai/backend/lm_buddy/cli/finetune.py
```diff
@@ -0,0 +1,13 @@
+import click
```
chainlink, line 1: Do we want to trim some things out like finetuning?

veekaybee: Yup, that will be step 2 described in the PR. I want to do this direct port first, test to make sure functionality is basically the same, and then the second PR following this will be removing: finetuning, ragas and prometheus if we decide we don't need them anymore, and perhaps some of the other pieces.

lumigator/python/mzai/backend/lm_buddy/__init__.py
aittalam: If we want to use it as just an evaluator, it makes sense to rename it that way and avoid confusion by removing non-eval entrypoints (namely finetuning).

lumigator/python/mzai/backend/lm_buddy/tests/conftest.py
aittalam, line 1: Just a heads up that tests should be ported too.
We might want to take advantage of this to remove some of the `cache-*` files under `datasets` (see individual folders there e.g. `tiny_shakespeare`, `xyz`) we mistakenly pushed in the past

lumigator/python/mzai/backend/services/experiments.py
veekaybee: Yup, all good points. I'll use a lm-buddy requirements.txt at the root of the repo, and in the next PR remove some of them.

lumigator/python/mzai/backend/lm_buddy/__init__.py
veekaybee: Will rename in next PR + remove finetuning

lumigator/python/mzai/backend/lm_buddy/tests/conftest.py
veekaybee: Tests are all ported, files removed

lumigator/python/mzai/backend/jobs/submission.py
```diff
@@ -25,8 +25,11 @@ class RayJobEntrypoint(ABC):
 
     @property
     def command(self) -> str:
-        """The command ran by the ray job, which is passed a JSON-serialized config."""
-        return f"lm-buddy evaluate huggingface --config '{json.dumps(self.config.args)}'"
+        # lm-buddy passed as a module to Ray using a JSON-serialized config.
+        return (
+            f"python -m lumigator.python.mzai.backend.lm_buddy evaluate huggingface "
```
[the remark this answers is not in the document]
veekaybee: It's based on Ray's architecture. We pass this command,

```lm-buddy evaluate huggingface --config '{json.dumps(self.config.args)}****```,

Because [Ray entrypoints require](https://github.com/ray-project/ray/blob/a1be06346e532e656c128ffd2230f06a4d72679e/python/ray/dashboard/modules/job/sdk.py#L149) a CLI for jobs, whether that's a Python script or a bash script, or some combination.  We also need to pass configuration into this CLI call as args.

File: .pre-commit-config.yaml [hunk 8]
```diff
@@ -24,7 +24,7 @@ repos:
     hooks:
       - id: pants-fmt-lint
         name: pants fmt lint
-        always_run: false
+        always_run: true
```
veekaybee, line 27: I thought this might fix the precommit flakey linter but no dice, so I filed an issue https://github.com/mozilla-ai/lumigator/issues/213

aittalam (approved):
Thank you Vicki!

I think it is a great first step, I tested the branch locally and could run an evaluation job on Ray by directly calling the API.

chainlink:
I'd like to try running this on coreweave (ideally before the merge if you don't mind) I think the `work_dir` approach might need some tuning (I was getting 'request too large' error when trying it before)

veekaybee:
Yup, works for me, let's pair if you have some time over the next few days to debug

PR #210, opened 2024-09-16
chore: ray image for local container depends on OS

ividal:
## What's changing
Users on a local Linux currently need to manually change the docker compose file to point to the right Ray image. This small change does it for them, so neither Mac nor Linux users need to do anything other than cloning, make bootstrap..., make local-up.

## How to test it
On a Linux machine, follow the procedure stated in the updated documentation. Ideally, having Pants installed, the usual
```
make bootstrap-dev-environment && source mzaivenv/bin/activate
make local-up
```

## Related Jira Ticket
[LUMI-36](https://mzai.atlassian.net/browse/LUMI-36)

## I already...

- [ x] updated the documentation.

[LUMI-36]: https://mzai.atlassian.net/browse/LUMI-36?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.devcontainer/docker-compose-local.yaml
+++ b/.devcontainer/docker-compose-local.yaml
@@ -34,7 +34,7 @@ services:
       - localstack-data:/var/lib/localstack
 
   ray:
-    image: rayproject/ray:2.30.0-py311-cpu${ARCH}
+    image: rayproject/ray:2.30.0-py311-cpu${RAY_ARCH_SUFFIX}
     ports:
       - "6379:6379"
       - "8265:8265"
--- a/Makefile
+++ b/Makefile
@@ -7,10 +7,11 @@ CUDA_AVAILABLE := $(shell nvidia-smi &> /dev/null; echo $$?)
 PANTS_INSTALLED := $(shell pants --version &> /dev/null; echo $$?)
 PYTHON :=
 #used in docker-compose to choose the right Ray image
-ARCH := 
+ARCH := $(shell uname -m)
+RAY_ARCH_SUFFIX :=
 
-ifeq ($(UNAME), Darwin)
-	ARCH := -aarch64
+ifeq ($(ARCH), arm64)
+	RAY_ARCH_SUFFIX := -aarch64
 endif
 
 ifndef PYTHON
@@ -77,7 +78,7 @@ LOCAL_DOCKERCOMPOSE_FILE:= .devcontainer/docker-compose-local.yaml
 	pants run 3rdparty/python:gen_requirements_python_linux_cpu
 
 local-up: 3rdparty/python/requirements_python_linux_cpu.txt
-	ARCH=$(ARCH) docker compose -f $(LOCAL_DOCKERCOMPOSE_FILE) up -d --build
+	RAY_ARCH_SUFFIX=$(RAY_ARCH_SUFFIX) docker compose -f $(LOCAL_DOCKERCOMPOSE_FILE) up -d --build
 	rm 3rdparty/python/requirements_python_linux_cpu.txt
 
 local-down:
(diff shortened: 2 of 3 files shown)


veekaybee (approved):
Looks good! Note that we'll want to sync with @javiermtorres on x-platform compatiblity

README.md
```diff
@@ -80,17 +80,18 @@ It consists of:
...
++ [Docker](https://docs.docker.com/engine/install/)
+    + On Linux, please also follow the [post-installation steps](https://docs.docker.com/engine/install/linux-postinstall/)
++ System Python (that is: no version manager, such as pyenv, should be active).
 
-## Local Development Setup (Currently targeting Mac)
+## Local Development Setup (either Mac or Linux)
 1. `git clone git@github.com:mozilla-ai/lumigator.git`
-2.  Install pants using homebrew `brew install pantsbuild/tap/pants` . For more on using Pants, read the [Pants guide](PANTS_GUIDE.md).
-3. `make bootstrap-dev-environment` and `source mzaivenv/bin/activate` to activate the virtualenv
+2.  Install [Pants](https://www.pantsbuild.org) using the [official instructions for your system](https://www.pantsbuild.org/2.21/docs/getting-started/installing-pants). For more on using Pants, read the [Pants guide](PANTS_GUIDE.md).
```
aittalam, line 93: +1e9 thanks for adding this!

aittalam (approved):
Looks good, thank you!

.devcontainer/docker-compose-local.yaml
```diff
@@ -34,7 +34,7 @@ services:
       - localstack-data:/var/lib/localstack
 
   ray:
-    image: rayproject/ray:2.30.0-py311-cpu-aarch64
+    image: rayproject/ray:2.30.0-py311-cpu${ARCH}
```
javiermtorres, line 37: Maybe `-${ARCH}` and remove the dash later on?

Makefile
```diff
@@ -6,7 +6,12 @@ PY_VERSION := 3.11.9
 CUDA_AVAILABLE := $(shell nvidia-smi &> /dev/null; echo $$?)
 PANTS_INSTALLED := $(shell pants --version &> /dev/null; echo $$?)
 PYTHON :=
+#used in docker-compose to choose the right Ray image
+ARCH := 
 
+ifeq ($(UNAME), Darwin)
+	ARCH := -aarch64
```
javiermtorres, line 13: ...so this becomes `ARCH := aarch64` 
Also, `UNAME` actually captures the OS. IMHO we should have something like `ARCH := $(shell uname -m)` above, which captures the arch itself (so we don't mix up the arch and the os).

javiermtorres (approved):

.devcontainer/docker-compose-local.yaml
aittalam: As the non-aarch64 image does not have any suffix, this would break things. As I am not 100% sure this is a standard (e.g. are suffixes starting with `_` used too?) perhaps adding the dash explicitly in the suffix will work better. WDYT?

Makefile
aittalam: Given that:
- different images use different architecture suffixes (e.g. `-aarch64` and `-arm64`) so you cannot directly append arch to an image and expect to find it
- for some images, there is no `-suffix` image for the default architecture (see e.g. [ray](https://hub.docker.com/r/rayproject/ray/tags?page_size=&ordering=&name=latest-py311-cpu))
- we are only interested in capturing the aarch64 suffix for ray right now

... perhaps we might just rename `ARCH` in `RAY_ARCH_PREFIX` so it is semantically accurate, and make it depend on either os being Darwin (`$(UNAME)`) or arch being arm64 (`$(shell uname -m)`)?
I would opt for reading the os as we are not supporting any arm64 platform.

javiermtorres (approved):

PR #211, opened 2024-09-16
Added updated notebook and deps

aittalam:
## What's changing

Main change is the notebook itself. It now relies on dialogsum for the walkthrough.
A few days ago `datasets` deprecated the use of `use_auth_token` which now breaks our current code, this is fixed by freezing `datasets` version in lumigator (we can then update lm-buddy deps when the task of including it into lumigator is completed)

## How to test it
Run the notebook. Note that you'll need mistral api if you want to generate GT with it

## Related Jira Ticket
https://mzai.atlassian.net/browse/LUMI-40

## I already...

### DIFF
--- a/.devcontainer/docker-compose-local.yaml
+++ b/.devcontainer/docker-compose-local.yaml
@@ -58,7 +58,7 @@ services:
       # NOTE: the following path contains a hash which needs to be updated whenever
       # the list of libraries installed in backend/services/experiments.py changes.
       # This is a temporary solution, see https://github.com/mozilla-ai/lumigator/pull/186
-      - LD_PRELOAD=/tmp/ray/session_latest/runtime_resources/pip/96de31139e548d3e8200d1e4b2575b77d5695312/virtualenv/lib/python3.11/site-packages/scikit_learn.libs/libgomp-d22c30c5.so.1.0.0
+      - LD_PRELOAD=/tmp/ray/session_latest/runtime_resources/pip/43b9b8b429ea9ef2b7b388371c223b0dacf13906/virtualenv/lib/python3.11/site-packages/scikit_learn.libs/libgomp-d22c30c5.so.1.0.0
     volumes:
       - ../lumigator/python/mzai/summarizer/summarizer.py:/home/ray/summarizer.py
 
--- a/lumigator/python/mzai/backend/services/experiments.py
+++ b/lumigator/python/mzai/backend/services/experiments.py
@@ -138,7 +138,7 @@ class ExperimentService:
         # NOTE: Whenever the list of libraries below change, the LD_PRELOAD line in
         # `docker-compose-local.yaml` has to be updated with a new hash value
         runtime_env = {
-            "pip": ["nltk==3.8.1", "lm-buddy[jobs]==0.12.1"],
+            "pip": ["nltk==3.8.1", "datasets==2.20.0", "lm-buddy[jobs]==0.12.1"],
             "env_vars": runtime_env_vars,
         }
         entrypoint = RayJobEntrypoint(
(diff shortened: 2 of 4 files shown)


.devcontainer/docker-compose-local.yaml
```diff
@@ -58,7 +58,7 @@ services:
       # NOTE: the following path contains a hash which needs to be updated whenever
       # the list of libraries installed in backend/services/experiments.py changes.
       # This is a temporary solution, see https://github.com/mozilla-ai/lumigator/pull/186
-      - LD_PRELOAD=/tmp/ray/session_latest/runtime_resources/pip/96de31139e548d3e8200d1e4b2575b77d5695312/virtualenv/lib/python3.11/site-packages/scikit_learn.libs/libgomp-d22c30c5.so.1.0.0
+      - LD_PRELOAD=/tmp/ray/session_latest/runtime_resources/pip/43b9b8b429ea9ef2b7b388371c223b0dacf13906/virtualenv/lib/python3.11/site-packages/scikit_learn.libs/libgomp-d22c30c5.so.1.0.0
```
aittalam, line 61: Note that as we updated deps this is the new hash. @veekaybee I know you have something similar in your PR right now, just a heads up so we can sync with rebase / merge

lumigator/python/mzai/backend/services/experiments.py
```diff
@@ -138,7 +138,7 @@ def create_experiment(self, request: ExperimentCreate) -> ExperimentResponse:
         # NOTE: Whenever the list of libraries below change, the LD_PRELOAD line in
         # `docker-compose-local.yaml` has to be updated with a new hash value
         runtime_env = {
-            "pip": ["nltk==3.8.1", "lm-buddy[jobs]==0.12.1"],
+            "pip": ["nltk==3.8.1", "datasets==2.20.0", "lm-buddy[jobs]==0.12.1"],
```
aittalam, line 141: @veekaybee a heads up that instead of updating `evaluate` to 0.4.3  in lm-buddy, I preferred to freeze `datasets` here so we won't touch lm-b's code until it is merged into lumigator

.devcontainer/docker-compose-local.yaml
veekaybee: Yup, thanks for the heads up on this. We'll really need to solve this problem at a higher-level soon :\

notebooks/walkthrough.ipynb
veekaybee, line 1: Worth noting for DialogSum, "This dataset is under [CC BY-NC-SA 4.0](https://creativecommons.org/licenses/by-nc-sa/4.0/) license. You may not use it for commercial use without permission." - Not the case for us now but in the future we'll need to evaluate.  Do we also need to cite this dataset now that we're using it?

veekaybee, line 1: in 

```
# inspect our data
df[:3]
```
could we add a note about why we look at these rows specifically?

veekaybee, line 1: in

```
models = [
    enc_dec_models[0],
]

models
```
since it's now only one value, makes sense to simplify the list

aittalam: That's a great point, TY Vicki! I think it is fair to cite them. I have already put a link to their github page (which points to data, code, and paper) in the notebook, if that's not enough we can add a paper citation in the README.

aittalam: I used this instead of `head` mainly to save vertical space :-) I'll use head with a parameter to make it clear that it's mostly about a size constraint rather than specific rows being selected

.devcontainer/docker-compose-local.yaml
aittalam: We can! I'll send a PR with a fix that gets that path from the worker's ENV

veekaybee (approved):
LGTM, thanks so much for these changes!

PR #214, opened 2024-09-18
Davide/datasets fix

aittalam:
## What's changing

- added s3 filesystem to deps (similarly to what we had with s3_client)
- fixed dataset upload issue (when S3 broke a record was still saved on the DB)
- fixed dataset deletion for full directories, added exception handling
- fixed dataset download for multiple-file datasets, now a list of signed URLs is returned to the client for downloading

## How to test it

- different methods can be tested from the API

## Related Jira Ticket
https://mzai.atlassian.net/browse/LUMI-41

## I already...

- [x] added some tests for any new functionality (updated unit tests to work with current code)

### DIFF
--- a/.devcontainer/docker-compose-local.yaml
+++ b/.devcontainer/docker-compose-local.yaml
@@ -20,7 +20,7 @@ services:
     image: localstack/localstack:3.4.0
     platform: linux/amd64
     ports:
-      - "0.0.0.0:4566:4566"
+      - 4566:4566
     environment:
       - SERVICES=s3:4566
       - CREATE_BUCKETS=lumigator-storage
@@ -61,6 +61,13 @@ services:
       - LD_PRELOAD=/tmp/ray/session_latest/runtime_resources/pip/3aad1b9393df933f711013a7d9650bb9821e891c/virtualenv/lib/python3.11/site-packages/scikit_learn.libs/libgomp-d22c30c5.so.1.0.0
     volumes:
       - ../lumigator/python/mzai/summarizer/summarizer.py:/home/ray/summarizer.py
+    # NOTE: to keep AWS_ENDPOINT_URL as http://localhost:4566 both on the host system
+    #       and inside containers, we map localhost to the host gateway IP.
+    #       This currently works properly, but might be the cause of networking
+    #       issues down the line. This should be used only for local, development
+    #       deployments.
+    extra_hosts:
+      - "localhost:host-gateway"
 
   backend:
     build:
@@ -97,6 +104,11 @@ services:
       - RAY_WORKER_GPUS_FRACTION=0
     volumes:
       - ../:/mzai
+    # NOTE: to keep AWS_ENDPOINT_URL as http://localhost:4566 both on the host system
+    #       and inside containers, we map localhost to the host gateway IP.
+    #       This currently works properly, but might be the cause of networking
+    #       issues down the line. This should be used only for local, development
+    #       deployments.
     extra_hosts:
       - "localhost:host-gateway"
 
--- a/lumigator/python/mzai/backend/api/deps.py
+++ b/lumigator/python/mzai/backend/api/deps.py
@@ -6,16 +6,17 @@ from fastapi import Depends
 from mypy_boto3_s3.client import S3Client
 from ray.dashboard.modules.serve.sdk import ServeSubmissionClient
 from ray.job_submission import JobSubmissionClient
+from s3fs import S3FileSystem
 from sqlalchemy.orm import Session
 
 from mzai.backend.db import session_manager
 from mzai.backend.repositories.datasets import DatasetRepository
 from mzai.backend.repositories.experiments import ExperimentRepository, ExperimentResultRepository
 from mzai.backend.repositories.groundtruth import GroundTruthDeploymentRepository
+from mzai.backend.services.completions import MistralCompletionService, OpenAICompletionService
 from mzai.backend.services.datasets import DatasetService
 from mzai.backend.services.experiments import ExperimentService
 from mzai.backend.services.groundtruth import GroundTruthService
-from mzai.backend.services.completions import MistralCompletionService, OpenAICompletionService
 from mzai.backend.settings import settings
 
 
@@ -34,9 +35,19 @@ def get_s3_client() -> Generator[S3Client, None, None]:
 S3ClientDep = Annotated[S3Client, Depends(get_s3_client)]
 
 
-def get_dataset_service(session: DBSessionDep, s3_client: S3ClientDep) -> DatasetService:
+def get_s3_filesystem() -> Generator[S3FileSystem, None, None]:
+    return S3FileSystem()
+
+
+S3FileSystemDep = Annotated[S3FileSystem, Depends(get_s3_filesystem)]
+
+
+def get_dataset_service(
+    session: DBSessionDep, s3_client: S3ClientDep, s3_filesystem: S3FileSystemDep
+) -> DatasetService:
     dataset_repo = DatasetRepository(session)
-    return DatasetService(dataset_repo, s3_client)
+    dataset_repo = DatasetRepository(session)
+    return DatasetService(dataset_repo, s3_client, s3_filesystem)
 
 
 DatasetServiceDep = Annotated[DatasetService, Depends(get_dataset_service)]
--- a/lumigator/python/mzai/backend/services/datasets.py
+++ b/lumigator/python/mzai/backend/services/datasets.py
@@ -1,4 +1,5 @@
 import csv
+import json
 from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import BinaryIO
@@ -94,6 +95,9 @@ class DatasetService:
     def _raise_not_found(self, dataset_id: UUID) -> None:
         raise HTTPException(status.HTTP_404_NOT_FOUND, f"Dataset '{dataset_id}' not found.")
 
+    def _raise_unhandled_exception(self, e: Exception) -> None:
+        raise HTTPException(status.HTTP_500_INTERNAL_SERVER_ERROR, str(e)) from e
+
     def _get_dataset_record(self, dataset_id: UUID) -> DatasetRecord:
         record = self.dataset_repo.get(dataset_id)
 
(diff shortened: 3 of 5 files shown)


agpituk:
Great work! I've got a question about the download URL it's currently generated for localstack. As an example, with this branch, I tried to download a dataset and it generated this URL:
`http://localstack:4566/lumigator-storage/datasets/9dcd61bd-24b5-4ff8-9b49-066c524f42b1/thunderbird_gt_bart.csv/data-00000-of-00001.arrow?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=test%2F20240918%2Fus-east-2%2Fs3%2Faws4_request&X-Amz-Date=20240918T123952Z&X-Amz-Expires=3600&X-Amz-SignedHeaders=host&X-Amz-Signature=931e05e485e5b0f5a2b645606da9602a1e6bc690853334229178c517439d4f32`

I've used the default values. That link doesn't work for me, I need to change it to `http://localhost:4566 ......` to work. Is this something you handled with local changes to `/etc/hosts` ? or is this commented somewhere in the platform? Maybe we need an update on the env vars to make it work out of the box?

aittalam:
> Great work! I've got a question about the download URL it's currently generated for localstack. As an example, with this branch, I tried to download a dataset and it generated this URL: `http://localstack:4566/lumigator-storage/datasets/9dcd61bd-24b5-4ff8-9b49-066c524f42b1/thunderbird_gt_bart.csv/data-00000-of-00001.arrow?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=test%2F20240918%2Fus-east-2%2Fs3%2Faws4_request&X-Amz-Date=20240918T123952Z&X-Amz-Expires=3600&X-Amz-SignedHeaders=host&X-Amz-Signature=931e05e485e5b0f5a2b645606da9602a1e6bc690853334229178c517439d4f32`
> 
> I've used the default values. That link doesn't work for me, I need to change it to `http://localhost:4566 ......` to work. Is this something you handled with local changes to `/etc/hosts` ? or is this commented somewhere in the repo? Maybe we need an update on the env vars to make it work out of the box?

That's a great point! Overall it boils down to whether we consider this more as a self-contained platform or a tool that runs in its users' environment. The ENV vars currently point to the S3 server as a domain name that should be correctly resolved both by the different services running within lumigator (e.g. the backend and ray) and by the client connecting to the API. As this does not happen for the local installation (we cannot point Lumigator services to 127.0.0.1 as long as they run in separate containers, and `localstack` has no meaning outside of docker compose) we need to fix this somehow.

So the TL/DR is: I am just using an updated `/etc/hosts` right now but am happy to discuss how to tackle this.
My intuition is that the API should be agnostic of whether the setup is local or remote and just point downloads to the S3 endpoint URL/IP. Perhaps the SDK or the UI, being closer to the client, could make sense of the URL and perhaps update it with a custom one? WDYT?

lumigator/python/mzai/backend/api/deps.py
```diff
@@ -34,9 +35,19 @@ def get_s3_client() -> Generator[S3Client, None, None]:
 S3ClientDep = Annotated[S3Client, Depends(get_s3_client)]
 
 
-def get_dataset_service(session: DBSessionDep, s3_client: S3ClientDep) -> DatasetService:
+def get_s3_filesystem() -> Generator[S3FileSystem, None, None]:
```
javiermtorres, line 38: I guess we'll end up with file-like objects regardless of where they're placed, either in local FS or in S3, right?

lumigator/python/mzai/backend/services/datasets.py
```diff
@@ -153,29 +171,51 @@ def get_dataset_s3_path(self, dataset_id: UUID) -> str:
...
         record = self._get_dataset_record(dataset_id)
 
-        # Delete from S3
-        # S3 delete is called first, since if this fails the DB delete won't take place
-        dataset_key = self._get_s3_key(record.id, record.filename)
-        self.s3_client.delete_object(Bucket=settings.S3_BUCKET, Key=dataset_key)
+        try:
+            # Delete from S3
+            # S3 delete is called first => if this fails the DB delete won't take place
+            dataset_key = self._get_s3_key(record.id, record.filename)
```
javiermtorres, line 177: could these details be hidden somewhere so the dataset service itself is not aware of what kind of FS is being used for this file?

agpituk:
> > Great work! I've got a question about the download URL it's currently generated for localstack. As an example, with this branch, I tried to download a dataset and it generated this URL: `http://localstack:4566/lumigator-storage/datasets/9dcd61bd-24b5-4ff8-9b49-066c524f42b1/thunderbird_gt_bart.csv/data-00000-of-00001.arrow?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=test%2F20240918%2Fus-east-2%2Fs3%2Faws4_request&X-Amz-Date=20240918T123952Z&X-Amz-Expires=3600&X-Amz-SignedHeaders=host&X-Amz-Signature=931e05e485e5b0f5a2b645606da9602a1e6bc690853334229178c517439d4f32`
> > I've used the default values. That link doesn't work for me, I need to change it to `http://localhost:4566 ......` to work. Is this something you handled with local changes to `/etc/hosts` ? or is this commented somewhere in the repo? Maybe we need an update on the env vars to make it work out of the box?
> 
> That's a great point! Overall it boils down to whether we consider this more as a self-contained platform or a tool that runs in its users' environment. The ENV vars currently point to the S3 server as a domain name that should be correctly resolved both by the different services running within lumigator (e.g. the backend and ray) and by the client connecting to the API. As this does not happen for the local installation (we cannot point Lumigator services to 127.0.0.1 as long as they run in separate containers, and `localstack` has no meaning outside of docker compose) we need to fix this somehow.
> 
> So the TL/DR is: I am just using an updated `/etc/hosts` right now but am happy to discuss how to tackle this. My intuition is that the API should be agnostic of whether the setup is local or remote and just point downloads to the S3 endpoint URL/IP. Perhaps the SDK or the UI, being closer to the client, could make sense of the URL and perhaps update it with a custom one? WDYT?

My personal take on this is that we should support running Lumigator out of the box locally without having to change `/etc/hosts `when possible. I assume playing with the env vars (or adding part of the code I removed in 208`https://github.com/mozilla-ai/lumigator/pull/208`) should be enough. I'm happy to tackle this in a separate PR myself if you want and agree with the above!

lumigator/python/mzai/backend/api/deps.py
aittalam: Yes, file-like or directory-like (as in the case of HF datasets).
I realise there is a bit of redundancy with both `s3_client` and `S3FileSystem`: these are used together to implement different s3-based tasks but I am quite sure we could consolidate them into just one if we choose to continue writing our own data storage system

lumigator/python/mzai/backend/services/datasets.py
aittalam: Definitely! At the moment, the DatasetService only works with S3. I think it'd be great to have an abstraction that allows us to be agnostic of the underlying storage. However, I am not sure we want to implement it from scratch (see my point re: artifact storage / experiment tracking). WDYT?

aittalam:
> > > Great work! I've got a question about the download URL it's currently generated for localstack. As an example, with this branch, I tried to download a dataset and it generated this URL: `http://localstack:4566/lumigator-storage/datasets/9dcd61bd-24b5-4ff8-9b49-066c524f42b1/thunderbird_gt_bart.csv/data-00000-of-00001.arrow?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=test%2F20240918%2Fus-east-2%2Fs3%2Faws4_request&X-Amz-Date=20240918T123952Z&X-Amz-Expires=3600&X-Amz-SignedHeaders=host&X-Amz-Signature=931e05e485e5b0f5a2b645606da9602a1e6bc690853334229178c517439d4f32`
> > > I've used the default values. That link doesn't work for me, I need to change it to `http://localhost:4566 ......` to work. Is this something you handled with local changes to `/etc/hosts` ? or is this commented somewhere in the repo? Maybe we need an update on the env vars to make it work out of the box?
> > 
> > 
> > That's a great point! Overall it boils down to whether we consider this more as a self-contained platform or a tool that runs in its users' environment. The ENV vars currently point to the S3 server as a domain name that should be correctly resolved both by the different services running within lumigator (e.g. the backend and ray) and by the client connecting to the API. As this does not happen for the local installation (we cannot point Lumigator services to 127.0.0.1 as long as they run in separate containers, and `localstack` has no meaning outside of docker compose) we need to fix this somehow.
> > So the TL/DR is: I am just using an updated `/etc/hosts` right now but am happy to discuss how to tackle this. My intuition is that the API should be agnostic of whether the setup is local or remote and just point downloads to the S3 endpoint URL/IP. Perhaps the SDK or the UI, being closer to the client, could make sense of the URL and perhaps update it with a custom one? WDYT?
> 
> My personal take on this is that we should support running Lumigator out of the box locally without having to change `/etc/hosts `when possible. I assume playing with the env vars (or adding part of the code I removed in 208`https://github.com/mozilla-ai/lumigator/pull/208`) should be enough. I'm happy to tackle this in a separate PR myself if you want and agree with the above!

I 100% agree we should not require users to manually change their `/etc/hosts`. Happy to discuss how to implement a solution for this in a subsequent PR!

agpituk (approved):

.devcontainer/docker-compose-local.yaml
```diff
@@ -20,7 +20,7 @@ services:
     image: localstack/localstack:3.4.0
     platform: linux/amd64
     ports:
-      - 4566:4566
+      - "0.0.0.0:4566:4566"
```
chainlink, line 23: Can you add some comments on mapping this to localhost? (and the extra hosts section below)

.devcontainer/docker-compose-local.yaml
```diff
@@ -97,6 +97,8 @@ services:
       - RAY_WORKER_GPUS_FRACTION=0
     volumes:
       - ../:/mzai
+    extra_hosts:
```
chainlink, line 100: I think we should add an extra note of caution here. (Since this modifies localhost networking)

.devcontainer/docker-compose-local.yaml
aittalam: Sure! I actually removed this change as it was not required for localstack to work properly, but I added comments below

.devcontainer/docker-compose-local.yaml
aittalam: +100

aittalam:
Added fixes and comments, tested all the pipeline from dataset upload to evaluation results, system works properly

PR #220, opened 2024-09-20
Resolving pathing and build dependencies between lm-buddy and Lumigator

veekaybee:
## What's changing

Now that we've imported lm-buddy into Lumigator, we need to make sure it builds correctly both locally on Mac and on remote servers with Linux distributions. This PR addresses two issues: 

1) The relative pathing is incrroect for lm buddy to be imported into Lumigator, this change is from `lumigator.python.mzai.lm_buddy to `mzai.lm_buddy` across all lm-buddy artifacts. 

2) The second issue is that now that we have two sets of dependencies, it's not clear how they should be resolved in light of Pants and there are import resolution issues because now both packages use, for example, Ray. 

## How to test it

To reproduce this problem:

A. Fetch branch
B. run `pants test ::` at the root of the repo, you should see issues resolving both ray and PyTorch (although this might be unrelated)

In order to get these tests to pass, build a new container and test on Coreweave and merge this PR, we need to decide: 

1. How to do dependency resolution between packages in the short-term
2. Whether we should keep PyTorch as a top-level platform dependency or relegate to the list of dependencies shipped to Ray
3. Whether to keep Pants  <--- this can perhaps be folded into a next PR

## I already...

cc @javiermtorres @ividal @aittalam


3rdparty/python/linux_cpu.lock
veekaybee, line 1: Note that these get regenerated since we have new dependencies coming in from  lm-buddy. `pants generate-lockfiles --resolve=linux_cuda`

3rdparty/python/pyproject.toml
```diff
@@ -3,23 +3,6 @@ name = "lumigator-3rdparty-python-dependencies"
 version = "0.0.1"
 requires-python = "==3.11.9"
 
-## local lm-buddy can be made with the following url spec.
```
veekaybee, line 6: removing this as we no longer build lm-buddy separately

lumigator/python/mzai/lm_buddy/BUILD.pants
```diff
@@ -0,0 +1,11 @@
+python_sources(
```
veekaybee, line 1: Adding a build file so lm buddy will build with the rest of the platform, but not compile to a PEX

lumigator/python/mzai/lm_buddy/__init__.py
```diff
@@ -1,3 +1,3 @@
-from lumigator.python.mzai.lm_buddy.buddy import LMBuddy
+from mzai.lm_buddy.buddy import LMBuddy
```
veekaybee, line 1: This is probably the source of the issues you were having in pex, @chainlink

lumigator/python/mzai/lm_buddy/tests/BUILD.pants
```diff
@@ -0,0 +1,9 @@
+python_sources(
```
veekaybee, line 1: Adding this file to enable lm-buddy tests to run in Pants

veekaybee:
Note that tests pass on linux but not on Mac, so we'll want to check that as well.

chainlink (approved):
Looks good!

veekaybee:
Note that app will still not bundle for linux due to path resolution issues for lm-buddy requirements , dependent on pex solution

PR #221, opened 2024-09-20
Add documentation for environment variables in README and docker compose

javiermtorres:
## What's changing

Documentation on the environment variables is added to the README and the docker compose file.

## How to test it

N/A

## Related Jira Ticket

LUMI-28

## I already...

### DIFF
--- a/.devcontainer/README.md
+++ b/.devcontainer/README.md
@@ -51,3 +51,10 @@ In order to use the API-based vendor ground truth generation and evaluation, you
 + `MISTRAL_API_KEY`
 + `OPENAI_API_KEY`
 into the docker container.
+
+## Image documentation
+
+The following documentation on the images used in the containers is available:
+
+* PostgreSQL: https://hub.docker.com/_/postgres
+* LocalStack: https://docs.localstack.cloud/references/configuration/
--- a/README.md
+++ b/README.md
@@ -142,17 +142,37 @@ If the S3 storage service is used, the endpoint, key and secret are needed. The 
 
 | Environment variable name | Default value | Description |
 | --- | :-: | --- |
-| LOCAL_FSSPEC_S3_ENDPOINT_URL | "" | Endpoint URL for the S3 data storage service |
-| LOCAL_FSSPEC_S3_KEY | "" | Key for the S3 data storage service |
-| LOCAL_FSSPEC_S3_SECRET | "" | Secret for the S3 data storage service |
-| LOCALSTACK_AUTH_TOKEN | "" | Authentication token for the LocalStack service |
+| AWS_HOST | "" | Host where S3 is located. Currently pointing at `services.localstack`. |
+| AWS_ENDPOINT_URL | "" | Endpoint URL for the S3 data storage service. |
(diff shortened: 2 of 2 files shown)


.devcontainer/docker-compose-local.yaml
```diff
@@ -9,6 +9,7 @@ services:
     ports:
       - 5432:5432
     environment:
+      # https://hub.docker.com/_/postgres
```
veekaybee, line 12: I'm not sure we need this comment since folks find the docker image here https://github.com/mozilla-ai/lumigator/blob/caa5b80ea8c982264f28fadc3bdabb7aa7c71bad/.devcontainer/docker-compose-local.yaml#L5C9-L5C30

.devcontainer/docker-compose-local.yaml
```diff
@@ -47,6 +49,7 @@ services:
           cpus: '2'
           memory: '5g'
     environment:
+      # ref???
```
veekaybee, line 52: I'm not sure we need this

.devcontainer/docker-compose-local.yaml
```diff
@@ -73,27 +76,60 @@ services:
     ports:
       - 80:80
     environment:
+      #
```
veekaybee, line 79: remove?

veekaybee (commented):
Thanks for adding the documentation. For the dockerfile, many of the variables are self-explanatory, so I'm not sure that adding comments would be helpful here. I've noted the ones that I think we could keep

.devcontainer/docker-compose-local.yaml
```diff
@@ -22,6 +23,7 @@ services:
     ports:
       - 4566:4566
     environment:
+      # https://docs.localstack.cloud/references/configuration/
```
veekaybee, line 26: This might be a good comment to leave in this readme instead https://github.com/mozilla-ai/lumigator/blob/main/.devcontainer/README.md

javiermtorres: I can move the list of compose'd image docs to this file, wdyt?

.devcontainer/docker-compose-local.yaml
javiermtorres: This would hold the description of `DEPLOYMENT_TYPE`. Right now it's just `local`, right?

.devcontainer/docker-compose-local.yaml
chainlink: Always a fan of docs, but I think this is redundant

.devcontainer/docker-compose-local.yaml
chainlink: remove for now, maybe let's get his PR shipped and iterate?

chainlink (approved):
Overall, I think this is a good incremental improvement and good to merge

.devcontainer/docker-compose-local.yaml
javiermtorres: I removed all comments in docker compose.

PR #224, opened 2024-09-23
Single image holding Lumigator and auxiliary services

javiermtorres:
## What's changing

A new dockerfile has been written including all necessary services, plus auxiliary start scripts.

## How to test it

From the root, buld the image:

```
docker build -t lumigator-test:0.1.0-alpha.1 -f .devcontainer/Dockerfile.single .
```

And then run the image in a new container passing a host directory in which the necessary data will be created, and mapping the necessary ports:

```
docker container run -it --init -v <host datadir>:/etc/lumigator/data --env-file .devcontainer/single-env.list -p 15432:5432 -p 14566:4566 -p 16379:6379 -p 18265:8265 -p 20001:10001 -p 18080:80 --entrypoint /bin/bash lumigator-test:0.1.0-alpha.1
```

## Related Jira Ticket

LUMI-56

Maybe some of the lumigator demo code needs to be changed, so the appropriate names for the services are all pointing to localhost.

Port mapping is the only way to access containers in MacOS. Additional information can be found in https://docs.docker.com/desktop/networking/#known-limitations and https://docs.docker.com/desktop/networking/#use-cases-and-workarounds. Host networking does not seem to work on Mac OS.

## I already...


javiermtorres:
Closed as the approach to individual deployments is changed.

PR #226, opened 2024-09-23
Change lm-buddy to evaluator and remove finetuning module

veekaybee:
## What's changing

As a continued piece of the lm-buddy refactor (see https://github.com/mozilla-ai/lumigator/pull/209 and https://github.com/mozilla-ai/lumigator/pull/220, we're renaming lm-buddy to `evaluator` since it's no longer a stand-alone library and the task-based naming is more idiomatic to other lumigator functionality (jobs, summarizer, deployment, etc) and will be helpful to developers working with the Lumigator codebase.

We're also removing the option to fine-tune models since Lumigator focuses specifically on evaluation.  Finally, we've added a README. 

This will be part of a stacked PR that will also include removal of Prometheus and RAGAS.  (related PR here: https://github.com/mozilla-ai/lumigator/pull/227)

**1. Simplifying our dependency chain:** Right now the lm-buddy code is shipped as a pip dependency (i.e. for every new lm-buddy release, we build/publish a package on pypi and this is downloaded at setup time within Lumigator. ) We wouldn't need to do this anymore with lm-buddy contained within lumigator and shipped with the repo as-is.  <--- done
2. **Remove redundant components:** we have a chance to make lm-buddy simpler and smaller as a part of this port, and as a result, take up less of a footprint within lumgiator as we ship it. We no longer use finetuning, so can remove all these components. <---- we are here
3. **Remove redundant routes:** Duplication of inference against OpenAI/Mistral for evaluation: we do this within Lumigator via ground truth and in lm-buddy via offline evaluation. We have a chance to fold these two code pieces together. 

## Related Jira Ticket

LUMI-69

Note that we'll need to test the evaluation functionality separately. 

## I already...

- [X] updated the documentation.

FYI @javiermtorres @agpituk


lumigator/python/mzai/backend/settings.py
```diff
@@ -88,7 +88,7 @@ def RAY_WORKER_GPUS_FRACTION(self) -> float:  # noqa: N802
     def PIP_REQS(self) -> list:  # noqa: N802
         return [
             line.strip()
-            for line in Path.open("lumigator/python/mzai/lm_buddy/requirements.txt")
+            for line in Path.open("lumigator/python/mzai/evaluator/requirements.txt")
```
veekaybee, line 91: Note that the way these requirements are currently pulled in means we build a pex executable that can't find them in Coreweave, modifications to come later. cc @chainlink

lumigator/python/mzai/evaluator/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+def command(self) -> str:
+        # evaluator passed as a module to Ray using a JSON-serialized config.
+        return (
+            f"python -m lumigator.python.mzai.evaluator evaluate huggingface "
+            f"--config '{json.dumps(self.config.args)}'"
+        )
+
+```
+and are populated from [these config templates.](https://github.com/mozilla-ai/lumigator/blob/main/lumigator/python/mzai/backend/config_templates.py)
```
veekaybee, line 30: Let me know if this piece is clear, happy to iterate on it.

lumigator/python/mzai/evaluator/tests/integration/test_lm_harness.py
```diff
@@ -24,7 +24,7 @@ def job_config(llm_model_path) -> LMHarnessJobConfig:
 
 
 def test_lm_harness_job(job_config):
-    buddy = LMBuddy()
+    buddy = Evaluator()
     result = buddy.evaluate(job_config)
```
veekaybee, line 28: Note that this will also be addressed here https://github.com/mozilla-ai/lumigator/compare/LUMI-69...LUMI-69-prometheus?expand=1#diff-10e0316f3138c092d3e1790c2bdff4259d65b18d70f854239f41482ceb37ae5bR28

.devcontainer/docker-compose-local.yaml
```diff
@@ -58,7 +58,7 @@ services:
       # NOTE: the following path contains a hash which needs to be updated whenever
       # the list of libraries installed in backend/services/experiments.py changes.
       # This is a temporary solution, see https://github.com/mozilla-ai/lumigator/pull/186
-      - LD_PRELOAD=/tmp/ray/session_latest/runtime_resources/pip/3aad1b9393df933f711013a7d9650bb9821e891c/virtualenv/lib/python3.11/site-packages/scikit_learn.libs/libgomp-d22c30c5.so.1.0.0
+      - LD_PRELOAD=/tmp/ray/session_latest/runtime_resources/pip/3aad1b9393df933f711013a7d9650bb9821e891/virtualenv/lib/python3.11/site-packages/scikit_learn.libs/libgomp-d22c30c5.so.1.0.0
```
aittalam, line 61:
```suggestion
      - LD_PRELOAD=/tmp/ray/session_latest/runtime_resources/pip/3aad1b9393df933f711013a7d9650bb9821e891c/virtualenv/lib/python3.11/site-packages/scikit_learn.libs/libgomp-d22c30c5.so.1.0.0
```

lumigator/python/mzai/backend/services/experiments.py
```diff
@@ -134,7 +134,7 @@ def create_experiment(self, request: ExperimentCreate) -> ExperimentResponse:
 
         runtime_env = {
             "pip": settings.PIP_REQS,
```
aittalam, line 136: Sorry I missed this earlier - why don't we just pass the `requirements.txt` file that comes in the lm-buddy (sorry, evaluator) directory? According to [ray docs](https://docs.ray.io/en/latest/ray-core/handling-dependencies.html#api-reference) this should be possible and

- allows ppl to provide requirements together with their jobs
- does not require settings to know the position of lm-buddy's requirements (removing the need to update this if anything's moved or added)

WDYT?

lumigator/python/mzai/backend/settings.py
aittalam, line 91: see comment above

lumigator/python/mzai/backend/jobs/submission.py
```diff
@@ -12,9 +12,9 @@
 class RayJobEntrypoint(ABC):
     """A generic command which is passed a config and submitted as a ray job.
 
-    Currently the only command we run is `lm-buddy evaluate`, but this can
-    be parametrised to support both different parameters (e.g. `lm-buddy
-    evaluate lm-harness`, `lm-buddy finetune`, etc) or even different commands.
+    Currently the only command we run is `evaluator evaluate`, but this can
```
aittalam, line 15: We are about to develop a batch inference job. Will this be "evaluator predict" or do we want to get back to a more general name so we do e.g. "rayjob evaluate" and "rayjob predict"? @veekaybee @chainlink

lumigator/python/mzai/evaluator/README.md
aittalam: Thanks for adding it!
I think it is as clear as it can be given the current setup :-)
We'll need to iterate on the new job development pipeline as soon as we gather feedback from the creation of extra tasks (lm-harness, batch inference, systemchat). I'll document the process while I work on lm-h!

lumigator/python/mzai/evaluator/constants.py
```diff
@@ -0,0 +1,12 @@
+import os
+from pathlib import Path
+
+EVALUATOR_HOME_PATH: str = os.getenv(
+    "EVALUATOR_HOME",
+    str(Path.home() / ".evaluator"),
+)
+
+EVALUATOR_RESULTS_PATH: str = os.getenv(
```
aittalam, line 9: In the code there is still LM_BUDDY_RESULTS_PATH (will try to capture those I see here, there might be more around)

lumigator/python/mzai/evaluator/jobs/evaluation/hf_evaluate.py
```diff
@@ -6,23 +6,23 @@
 
 import s3fs
 from loguru import logger
-from mzai.lm_buddy.configs.jobs.hf_evaluate import HuggingFaceEvalJobConfig
-from mzai.lm_buddy.configs.vllm import VLLMCompletionsConfig
-from mzai.lm_buddy.constants import LM_BUDDY_RESULTS_PATH
-from mzai.lm_buddy.jobs.asset_loader import (
+from mzai.evaluator.configs.jobs.hf_evaluate import HuggingFaceEvalJobConfig
+from mzai.evaluator.configs.vllm import VLLMCompletionsConfig
+from mzai.evaluator.constants import LM_BUDDY_RESULTS_PATH
```
aittalam, line 11:
```suggestion
from mzai.evaluator.constants import EVALUATOR_RESULTS_PATH
```

lumigator/python/mzai/evaluator/jobs/evaluation/prometheus.py
```diff
@@ -9,18 +9,18 @@
 
 from datasets import Dataset
 from loguru import logger
-from mzai.lm_buddy.configs.huggingface import AutoTokenizerConfig
-from mzai.lm_buddy.configs.jobs.prometheus import PrometheusJobConfig
-from mzai.lm_buddy.constants import LM_BUDDY_RESULTS_PATH
-from mzai.lm_buddy.jobs.asset_loader import (
+from mzai.evaluator.configs.huggingface import AutoTokenizerConfig
+from mzai.evaluator.configs.jobs.prometheus import PrometheusJobConfig
+from mzai.evaluator.constants import LM_BUDDY_RESULTS_PATH
```
aittalam, line 14:
```suggestion
from mzai.evaluator.constants import EVALUATOR_RESULTS_PATH
```

lumigator/python/mzai/evaluator/jobs/evaluation/ragas.py
```diff
@@ -7,15 +7,15 @@
 from ragas import evaluate as ragas_evaluate
 from ragas.metrics import answer_relevancy, context_precision, context_recall, faithfulness
 
-from mzai.lm_buddy.configs.jobs.ragas import RagasJobConfig
-from mzai.lm_buddy.constants import LM_BUDDY_RESULTS_PATH
-from mzai.lm_buddy.jobs.asset_loader import (
+from mzai.evaluator.configs.jobs.ragas import RagasJobConfig
+from mzai.evaluator.constants import LM_BUDDY_RESULTS_PATH
```
aittalam, line 11:
```suggestion
from mzai.evaluator.constants import EVALUATOR_RESULTS_PATH
```

lumigator/python/mzai/evaluator/README.md
javiermtorres: should the templates end up eventually as json schemas?

veekaybee: Kind of, they get passed as [Pydantic models](https://docs.pydantic.dev/latest/concepts/models/), which can be converted to/from JSON.

lumigator/python/mzai/backend/services/experiments.py
veekaybee: this is what we're doing, but passing through settings because I wanted to consolidate all files, variables, etc in one place. I wouldn't want the filepath to be in the code directly since that could change. In the case of the direct filepath, this module would have to know where the filepath is relative to the `services` directory anyway. It's a tricky issue to resolve because most Ray installs assume you're running the script from the same directory the code is.

lumigator/python/mzai/backend/settings.py
veekaybee: We'll still need to eventually point to the requirements path in the evaluator directory, unless I'm misunderstanding

lumigator/python/mzai/evaluator/constants.py
veekaybee: thank you!

lumigator/python/mzai/backend/services/experiments.py
aittalam: Gotcha! One possible suggestion we discussed: what if we set `working_dir` to the actual evaluator dir and just consider it as some independent package? Eventually rn it has no lumigator deps and when we run it on ray it'll just be its own thing. This allows us to only reference its dir once (in working_dir) and pip install requirements.txt because it'll be directly accessible in the working directory

veekaybee: works for now!

aittalam (approved):

chainlink (approved):

PR #227, opened 2024-09-23
Remove Ragas and Prometheus from evaluation suite

veekaybee:
## What's changing

In the short-to-medium term, we'd like to cut down on technical debt and since we're no longer using  Ragas and Prometheus for our evaluation suite, remove them from the evaluation options and cut down on their dependencies.  This PR depends on this one being approved: https://github.com/mozilla-ai/lumigator/pull/226

## I already...


aittalam (approved):
TY for this first pass!

If I recall correctly sentencepiece was a dep for ragas/prometheus, but I am not sure whether the following updates also required. Did you have a chance to test the current code without it?

A few possible next-steps after this are:

- configs: after this simplification step (and perhaps before re-increasing the complexity by adding new evals), we should be able to employ lm-buddy's config schemas more widely. This should give us some typechecking for free in the code that relies on it (e.g. job submission if we replace the dictionaries with our configs and perhaps the SDK itself could rely on this)
- lm-eval job update and integration with lumigator
- update breaking tests (is the paths issue related to the pex/packaging issue we are having right now?)

WDYT? I am happy to add those as issues if they are not tracked yet

chainlink:
A note on the configs, we should pull them to a separate module shared by lumigator and lm-buddy. 
image

veekaybee:
> A note on the configs, we should pull them to a separate module shared by lumigator and lm-buddy.

Thanks for this, are you thinking at this level? 

Screenshot 2024-09-25 at 9 09 27 AM

More specifically, we also have config_templates in the backend that we'd also need to move right?

veekaybee:
> A few possible next-steps after this are:
> 
>     configs: after this simplification step (and perhaps before re-increasing the complexity by adding new evals), we should be able to employ lm-buddy's config schemas more widely. This should give us some typechecking for free in the code that relies on it (e.g. job submission if we replace the dictionaries with our configs and perhaps the SDK itself could rely on this)
>     lm-eval job update and integration with lumigator
>     update breaking tests (is the paths issue related to the pex/packaging issue we are having right now?)
> 

Yep, let's please add issues here. The next tickets will be:

1. Split out config once we clarify what we'd like where 
2. Fix the pex tests - dependent on @chainlink monorepo experiments
3. lm-eval job 

1+3 are not dependent on getting Lumigator working on coreweave since it works locally but would be a nice-to-have.

PR #230, opened 2024-09-24
Tests for the SDK draft

javiermtorres:
Adding tests to the draft


lumigator/python/mzai/backend/backend/tests/BUILD.pants
```diff
@@ -6,4 +6,8 @@ python_sources(
 python_tests(
     name="backend_tests",
     sources=["**/test_*.py"],
+    dependencies=[
```
veekaybee, line 9: we don't need this

lumigator/python/mzai/backend/schemas/deployments.py
```diff
@@ -28,3 +28,7 @@ class DeploymentEvent(BaseModel):
     deployment_type: DeploymentType
     status: DeploymentStatus
     detail: str | None = None
+
+
+class DeploymentEventList(RootModel):
```
veekaybee, line 33: we should only need to inherit from `BaseModel `. https://docs.pydantic.dev/latest/concepts/models/#model-methods-and-properties

lumigator/python/mzai/sdk/tests/data/job.json
```diff
@@ -0,0 +1,59 @@
+{
```
veekaybee, line 1: let's add these to a `static` resources folder like in evaluator to keep things consistent in the module https://github.com/mozilla-ai/lumigator/tree/main/lumigator/python/mzai/evaluator/tests/resources

javiermtorres: Is there anything else needed apart from changing the name of the directory?

lumigator/python/mzai/backend/schemas/deployments.py
javiermtorres: Ok, thanks. This was an attempt at loading lists of X, until I noticed https://github.com/mozilla-ai/lumigator/blob/b1971690e0b9abf45280c69831f141be7164d468/lumigator/python/mzai/backend/schemas/extras.py#L21

lumigator/python/mzai/backend/backend/tests/BUILD.pants
javiermtorres: Let me recheck. The tests were not finding some source code before adding these dependencies. But maybe it's not the right way to make pytest find them.

javiermtorres: Ok, I still have issues with these tests but they don't seem the previous errors complaining that schemas could not be resolved. AFAICT this is not present in the new SDK PRs.

javiermtorres:
Closed to be split up in several PRs addressing specific endpoints

PR #236, opened 2024-09-27
Use custom sql init (sqlite)

javiermtorres:
## What's changing

The whole DB URL is passed through the environment. This allows a custom DB to be used without any changes.

## How to test it

Start the lumigator image using the provided URL, and removing the postgresql service from the docker compose. SQLite should be used instead locally in the lumigator image.

## Related Jira Ticket

LUMI-86

## I already...


lumigator/python/mzai/backend/tests/conftest.py
```diff
@@ -28,7 +28,8 @@ def postgres_container():
 @pytest.fixture(scope="session", autouse=True)
 def db_engine(postgres_container: PostgresContainer):
     """Initialize a DB engine bound to the Postres container, and create tables."""
-    engine = create_engine(postgres_container.get_connection_url(), echo=True)
+    # engine = create_engine(postgres_container.get_connection_url(), echo=True)
```
agpituk, line 31: Why leave it commented?  I am personally in favor of removing it from the code if we're not using it (we have git history if we need this again)

.devcontainer/docker-compose-local.yaml
```diff
@@ -81,6 +81,9 @@ services:
       - 80:80
     environment:
       - DEPLOYMENT_TYPE=local
+      # The local file needs to be available through a mount,
+      # if persistence is needed
+      - SQLALCHEMY_DATABASE_URL=sqlite:///local.db
```
agpituk, line 86: Can't we have postgres vars removed & postgres removed from the docker-compose? If you connect to an external postgres changing the SQLALCHEMY_DATABSE_URL it would still work, right?

agpituk (changes requested):
I think we should remove postgres dependencies from the docker-compose and simplify the variables  as well

agpituk:
> I think we should remove postgres dependencies from the docker-compose and simplify the variables as well

Also, can we add local.db to .gitignore?

lumigator/python/mzai/backend/tests/conftest.py
javiermtorres: Ok, removing

.devcontainer/docker-compose-local.yaml
javiermtorres: IIRC there are dependencies to the postgres variables somewhere inside. I can make a quick test and try to find out where they are used. They were used only to create the db url, but I found errors when I removed them. I'll give it a check.

agpituk (approved):

chainlink (approved):

PR #237, opened 2024-09-30
Check test race conditions

javiermtorres:
## What's changing

The test cases sometimes run into race conditions. Even though `pytest` itself is sequential, pants will run each test in parallel as shown [here](https://www.pantsbuild.org/dev/docs/python/goals/test). However, the only effect we have come across for the moment is an error creating the buckets, when the fixtures for different tests overlap. The fix focuses then on ignoring errors creating the buckets when the bucket already exists.

## How to test it

Just run the `pants` test.

## Related Jira Ticket

- 

## I already...

### DIFF
--- a/lumigator/python/mzai/backend/backend/tests/conftest.py
+++ b/lumigator/python/mzai/backend/backend/tests/conftest.py
@@ -15,6 +15,8 @@ from backend.main import create_app
 from backend.records.base import BaseRecord
 from backend.settings import settings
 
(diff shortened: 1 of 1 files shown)


aittalam (approved):
Thank @javiermtorres, I had this issue on github just recently and could not replicate it locally. I appreciate the fix!

veekaybee (approved):

PR #238, opened 2024-09-30
CLI :Remove Ragas and Prometheus from CLI

peteski22:
## What's changing

CLI code contained references to Ragas and Prometheus that are no longer supported, this PR removes them from the evaluator.

## How to test it

Ran through the (currently internal) manual test via cURL commands.


veekaybee (approved):
Thanks for this, looks good!

PR #239, opened 2024-09-30
ran: `ruff check` for linting

peteski22:
## What's changing

Running `ruff check` from the lumigator root gives these results.

Change seem to mostly be reordering of imports to follow: https://peps.python.org/pep-0008/#imports

> Imports should be grouped in the following order:
>
> * Standard library imports.
> * Related third party imports.
> * Local application/library specific imports.
> 
> You should put a blank line between each group of imports.

Two things that stood out:

1. [List => list](https://github.com/mozilla-ai/lumigator/compare/linting/ruff-check?expand=1#diff-c68d57f3b4b654e42bf8551142c047800cdcbd4d2a08c0ea1aa857f9dc26335aL41) - [List deprecated](https://docs.python.org/3/library/typing.html#typing.List)
2. [Optional[str] => None](https://github.com/mozilla-ai/lumigator/compare/linting/ruff-check?expand=1#diff-a91541141b02fd65436aa3a336b17b04b35de436dddb8b7216e27b21254a6a61L35-L43) - seems to be the preferred way [PEP-0604](https://peps.python.org/pep-0604/)

### DIFF
--- a/lumigator/python/mzai/backend/api/deployments/summarizer_config_loader.py
+++ b/lumigator/python/mzai/backend/api/deployments/summarizer_config_loader.py
@@ -1,8 +1,8 @@
 from typing import Any
+from uuid import UUID
 
 from loguru import logger
 from pydantic import BaseModel
-from uuid import UUID
 
 from mzai.backend.api.deployments.configloader import ConfigLoader
 from mzai.backend.settings import settings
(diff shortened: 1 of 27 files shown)


peteski22:
Closing as we use currently `ruff` via `pants` for linting so this might not be required.

chainlink:
I just spotted this too working on another PR, I think pants fmt isn't working correctly

PR #242, opened 2024-10-02
Lumi 69 fix

aittalam:
## What's changing

Pickin' up @veekaybee's LUMI-69-fix (with my changes merged) and bringing it to main.
This solves issues with evaluator paths, fixes tests, removes pants issues on evaluator, updates lockfiles.

## How to test it

Tested with
- pants test (for lumigator)
- pytest specifically on the evaluator dir
- integration test with CURL


veekaybee (approved):
Approving as we tested and walked through this branch this morning, thanks so much Davide!

PR #245, opened 2024-10-03
Schema discussion draft

javiermtorres:
## What's changing

Initial proposal for schema handling

## I already...

### DIFF
--- a/lumigator/python/mzai/schemas/__init__.py
+++ b/lumigator/python/mzai/schemas/__init__.py
@@ -1,0 +1,15 @@
+# Are these schemas used in both API and SDK?
+# As long as they're not used for other purposes
+# like DB storage, it should be ok
+
+# We could start with an OpenAPI spec and then generate the models
+# using https://docs.pydantic.dev/latest/integrations/datamodel_code_generator/
+
+# Otherwise, we can generate manually the openapi spec using
+# https://www.speakeasy.com/openapi/frameworks/pydantic
+
+# One important item is versioning: the `info` item in an OpenAPI
+# spec contains the version of the application API itself.
+# We'd need to check how to add the info to the openapi spec, but
+# here are detailed instructions that we could explore:
+# https://www.speakeasy.com/openapi/frameworks/pydantic
--- a/lumigator/python/mzai/schemas/experiments.py
+++ b/lumigator/python/mzai/schemas/experiments.py
@@ -16,7 +16,11 @@ class ExperimentCreate(BaseModel):
     system_prompt: str | None = None
     config_template: str | None = None
 
+class ExperimentResult(BaseModel):
+    download_url: str
 
+# The experiment data is retrieved like .../experiments/{experiment_id}
+# The experiment result is retrieved like .../experiments/{experiment_id}/result
 class ExperimentResponse(BaseModel, from_attributes=True):
     id: UUID
     name: str
@@ -24,13 +28,5 @@ class ExperimentResponse(BaseModel, from_attributes=True):
     status: JobStatus
     created_at: datetime.datetime
     updated_at: datetime.datetime | None
+    result: ExperimentResult | None
 
-
-class ExperimentResultResponse(BaseModel, from_attributes=True):
-    id: UUID
-    experiment_id: UUID
-
-
-class ExperimentResultDownloadResponse(BaseModel):
-    id: UUID
-    download_url: str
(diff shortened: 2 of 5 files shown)


lumigator/python/mzai/schemas/__init__.py
```diff
@@ -0,0 +1,15 @@
+# Are these schemas used in both API and SDK?
```
veekaybee, line 1: not used for DB storage, those are in the repositories.  https://github.com/mozilla-ai/lumigator/blob/main/lumigator/python/mzai/backend/repositories/experiments.py

lumigator/python/mzai/schemas/experiments.py
```diff
@@ -16,21 +16,17 @@ class ExperimentCreate(BaseModel):
     system_prompt: str | None = None
     config_template: str | None = None
 
+class ExperimentResult(BaseModel):
```
veekaybee, line 19: this would also change right? https://github.com/mozilla-ai/lumigator/blob/6b89a1520ead58b15b04ca2f5eefe44d851e9146/lumigator/python/mzai/backend/services/experiments.py#L178

lumigator/python/mzai/schemas/__init__.py
javiermtorres: Ok, thanks. Would it make sense to merge modules `repositories` and `records` since they're both ORM?

lumigator/python/mzai/schemas/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+# As long as they're not used for other purposes
+# like DB storage, it should be ok
+
+# We could start with an OpenAPI spec and then generate the models
+# using https://docs.pydantic.dev/latest/integrations/datamodel_code_generator/
```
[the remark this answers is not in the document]
javiermtorres: TL;DR yes :) 

The idea is that since it's a remote invocation, and we're using python at both ends of the openapi, we should be using the same (or very similar) objects in both ends, the SDK and the server implementation. In that case there are never issues of mismatch between data sent and data received.

Usually one starts with an interface spec (CORBA, grpc, for example) and then derives code for the client and server (stubs and skeletons). In our case, we've specified the interface in pydantic, which allows us to walk in both directions: either from spec to code (the datamodel code generator) or from code to spec (model_to_json_schema or something like it). It's fine in both cases (as long as we provide the openapi spec in the end to external devs to be able to derive clients in other languages).

Sometimes the comms-related features (like serialization) in such data types are hidden behind an interface. In this case we derive from BaseModel, which is enough to provide serialization, and the SDK user will not need to know or use anything about it.

lumigator/python/mzai/schemas/experiments.py
javiermtorres: Yes, of course. The server impl will need to be adjusted.

Note that, in this case, we're leaking the FK solution in the DB into the API, which is not desirable because using two tables with FKs is actually an implementation aspect of our 1:1 relation between experiments and results (and since it's a 1:1, one could actually think about storing the results inside the experiment itself, but as it is we have the possibility to change to 1:M later on, if an experiment provides several results).

javiermtorres: _Actually_ one could think why a join is not done in this case. The original is (crossing class boundaries, etc etc):

```
record = self.experiment_repo.get(experiment_id)
...
result_record = self.result_repo.get_by_experiment_id(experiment_id)
```

Which I'd assume to result in two separate SQL statements. Maybe we could provide another method that combines both like this:

```
    def get_by_experiment_id(self, experiment_id: UUID) -> ExperimentResultRecord | None:
        return (
            self.session.query(ExperimentRecord, ExperimentResultRecord)
            .where(ExperimentResultRecord.experiment_id == experiment_id)
            .where(ExperimentRecord.experiment_id == ExperimentResultRecord.experiment_id)
            .first()
        )
```

This ensures transactional behaviour and can also be reused later on to retrieve sets of experiments+results (by dropping the first `where`).

javiermtorres:
You can generate a sample OpenAPI schema using
```
pants run lumigator/python/mzai/schemas/openapi.py:schemas_sources@parametrize=darwin
```
We can discuss what pieces would still be missing in here (mainly the endpoints) and how to get them more or less automatically from live inspection.

javiermtorres:
Obsolete.

PR #248, opened 2024-10-04
Parametrized job submission

aittalam:
## What's changing
Minor update to parametrize job submission. jobs/submission.py now gets a `command` parameter (string) and optional `args` (dictionary with k = --parameter_name and v = parameter_value) so it is possible to run jobs either by directly passing the command as a single string or just the base command + params (eventually everything becomes one string anyway at submission time but perhaps it saves a bit of hassle, especially if we decide to move back to structured configs and bring serialization into submission.py so developers won't have to be bothered by that).

## How to test it
curl directly hitting the API to run an experiment

## Related Jira Ticket
https://mzai.atlassian.net/browse/LUMI-87

### DIFF
--- a/lumigator/python/mzai/backend/backend/jobs/submission.py
+++ b/lumigator/python/mzai/backend/backend/jobs/submission.py
@@ -16,11 +16,9 @@ class RayJobEntrypoint(ABC):
     memory: int | float | None = None
 
     @property
-    def command(self) -> str:
+    def command_with_params(self) -> str:
         """A generic command which is passed some optional args and submitted as a ray job.
 
-        It likely will be a ptyhon module used via cli (e.g. `python -m evaluator evaluate ...`)
-        or a different kind of command as long as it is executable.
         Note that parameters can either be passed as config.args (a dictionary containing
         parameter names as keys and parameter values as values) or directly with the command
         string.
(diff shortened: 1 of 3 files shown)


lumigator/python/mzai/backend/backend/jobs/submission.py
```diff
@@ -26,11 +17,21 @@ class RayJobEntrypoint(ABC):
 
     @property
     def command(self) -> str:
```
veekaybee, line 19:
```suggestion
    def build_ray_job_params
```

or similar since we now build up to the full command

lumigator/python/mzai/backend/backend/jobs/submission.py
```diff
@@ -26,11 +17,21 @@ class RayJobEntrypoint(ABC):
...
-        # evaluator entrypoint passed as a module to Ray using a JSON-serialized config.
-        return (
-            f"python -m evaluator evaluate huggingface "
-            f"--config '{json.dumps(self.config.args)}'"
-        )
+        """A generic command which is passed some optional args and submitted as a ray job.
+
+        It likely will be a ptyhon module used via cli (e.g. `python -m evaluator evaluate ...`)
```
veekaybee, line 22:
```suggestion
    
```

we can remove since we won't use CLI

lumigator/python/mzai/backend/backend/services/experiments.py
```diff
@@ -114,13 +114,19 @@ def create_experiment(self, request: ExperimentCreate) -> ExperimentResponse:
             # (which works with seq2seq models too except it does not use pipeline)
             config_template = config_templates.causal_template
 
-        eval_config_dict = json.loads(config_template.format(**config_params))
+        eval_config_args = {
```
veekaybee, line 117: let's add a comment here about what this does

veekaybee (approved):
approved after a few comments

veekaybee (approved):

PR #251, opened 2024-10-07
Add docker backend

chainlink:
## What's changing
Adds a new dockerfile that works via UV

This allows for file upload, and works the summarizer via docker compose, but it does not work for eval just yet (will be done in a follow up PR)

## I already...

### DIFF
--- a/lumigator/python/mzai/backend/Dockerfile.uv
+++ b/lumigator/python/mzai/backend/Dockerfile.uv
@@ -1,0 +1,12 @@
+FROM python:3.11-slim-bookworm
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /bin/uv
+
+# Copy the project into the image
+ADD . /app
+
+WORKDIR /app
+
+# Sync the project into a new environment, using the frozen lockfile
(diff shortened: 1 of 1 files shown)


lumigator/python/mzai/backend/Dockerfile.uv
```diff
@@ -0,0 +1,12 @@
...
+
+# Copy the project into the image
+ADD . /app
+
+WORKDIR /app
+
+# Sync the project into a new environment, using the frozen lockfile
+RUN uv sync --frozen
+
+CMD ["uv", "run", "uvicorn",  "--host", "0.0.0.0", "--port", "8000", "backend.main:app"]
```
javiermtorres, line 12: Maybe make the port configurable vía env var?

javiermtorres: Ideally, at runtime. But this means using a script as CMD and moving the current starting point there, we can leave that for later.

macaab26 (approved):

chainlink:
Lint is failing, but this has been a persistent flakey issue with pants 
image

PR #254, opened 2024-10-08
Davide/lumi 100 fix ld preload

aittalam:
## What's changing

- `LD_PRELOAD` is removed from docker compose
- `LD_PRELOAD` is prepended, at runtime, before calling the ray job. This allows us to set it as relative to `$VIRTUAL_ENV` which has the correct pip path
- minor fix to evaluator to log job configuration

## How to test it

`make local-up` + start a ray job by hitting the API via curl

## Related Jira Ticket

https://mzai.atlassian.net/browse/LUMI-100

This assumes that preloading the library is not breaking under different architectures. I do not think this is the case but I am happy to set up a test on an amd64/x86-64 environment.


lumigator/python/mzai/backend/backend/services/experiments.py
```diff
@@ -122,13 +122,17 @@ def create_experiment(self, request: ExperimentCreate) -> ExperimentResponse:
             "--config": config_template.format(**config_params),
         }
 
+        # as our command relies on libgomp which gives issues on aarch64, we
```
veekaybee, line 125:
```suggestion
Pre-loading libgomp before all other python imports resolves dependency errors on aarch64. Since we send a hashed list of imports to Ray, we need to change the python prefix path dynamically. 

see LD_PRELOAD_PREFIX in settings.py 
```

lumigator/python/mzai/evaluator/entrypoint.py
```diff
@@ -51,7 +51,7 @@ def evaluate(self, config: EvaluationJobConfig) -> EvaluationResult:
         The underlying evaluation framework is determined by the configuration type.
         """
         print("evaluating...")
-        logger.info("evaluating config: %s", config)
+        logger.info(f"evaluating config: {config}")
```
veekaybee, line 54: We'll need to log with print statements here since these logs are surfaced through Ray, which uses print for default

lumigator/python/mzai/backend/backend/settings.py
```diff
@@ -1,11 +1,10 @@
...
-from sqlalchemy.engine import URL
-from pathlib import Path
 import os
+from collections.abc import Mapping
 
+from pydantic import ByteSize, computed_field
+from pydantic_settings import BaseSettings
 from schemas.extras import DeploymentType
-from collections.abc import Mapping
+from sqlalchemy.engine import URL
```
veekaybee, line 7: possibly unused?

veekaybee (approved):
lgtm after a few changes

lumigator/python/mzai/backend/backend/services/experiments.py
```diff
@@ -122,13 +122,17 @@ def create_experiment(self, request: ExperimentCreate) -> ExperimentResponse:
...
+        # preload it with LD_PRELOAD_PREFIX (see settings.py for details)
+        eval_command = f"{settings.LD_PRELOAD_PREFIX} python -m evaluator evaluate huggingface"
```
chainlink, line 127: Add a note as to why we're setting here and not in env_vars (presumably because VIRTUAL_ENV hasn't been resolved in env vars yet) 

Also, this should only get set in certain circumstances (only on mac) not in all environments. This was implicitly handled before by only setting it the local docker compose

lumigator/python/mzai/evaluator/entrypoint.py
aittalam: Ah ok, that's good to know!
Il will fix the print adding the % then

lumigator/python/mzai/backend/backend/settings.py
aittalam: keeping because of `  def inherit_ray_env(self, runtime_env_vars: Mapping[str, str]): `

lumigator/python/mzai/backend/backend/services/experiments.py
aittalam: Thanks!
- added rationale here
- added architecture-specific setting of the variable (in settings.py for readability, with explanation)

PR #255, opened 2024-10-08
SDK: add `datasets`

javiermtorres:
## I already...

### DIFF
--- a/lumigator/python/mzai/backend/backend/main.py
+++ b/lumigator/python/mzai/backend/backend/main.py
@@ -13,6 +13,14 @@ import sys
 import os
 
 
+LUMIGATOR_APP_TAGS = {
+    "title": "Lumigator Backend",
+    "description": "Backend server",
+    "version": "0.0.1",
+    "openapi_tags": TAGS_METADATA,
+}
+
+
 def create_app(engine: Engine) -> FastAPI:
     @contextlib.asynccontextmanager
     async def lifespan(app: FastAPI):
@@ -20,7 +28,7 @@ def create_app(engine: Engine) -> FastAPI:
         BaseRecord.metadata.create_all(engine)
         yield
 
-    app = FastAPI(title="Lumigator Backend", lifespan=lifespan, openapi_tags=TAGS_METADATA)
+    app = FastAPI(**(LUMIGATOR_APP_TAGS | {"lifespan": lifespan}))
 
     main_log_level = os.getenv("LOG_LEVEL", "INFO").upper()
     logger.remove()
--- a/lumigator/python/mzai/sdk/client.py
+++ b/lumigator/python/mzai/sdk/client.py
@@ -50,7 +50,7 @@ def _make_request(
             **kwargs,  # noqa: B026
         )
         response.raise_for_status()
-        if verbose:
+        if verbose and response.content:
             logger.info(f"{json.dumps(response.json(), indent=2)}")
     except requests.RequestException as e:
         logger.error(f"Request failed: {e}")
@@ -69,6 +69,7 @@ class ApiClient:
         api_path,
         method: HTTPMethod = HTTPMethod.GET,
         data=None,
+        files=None,
         json_data=None,
         verbose: bool = True,
     ) -> requests.Response:
@@ -78,7 +79,9 @@ class ApiClient:
         path = f"{self._api_url.rstrip('/')}/{api_path.lstrip('/')}"
 
         try:
-            response = _make_request(path, method, data=data, json_=json_data, verbose=verbose)
+            response = _make_request(
+                path, method, data=data, files=files, json_=json_data, verbose=verbose
+            )
             # Support returning a response for 200-204 status codes.
             # NOTE: Other status codes that are returned without an HTTP error aren't supported.
             # e.g. 307 - Temporary Redirect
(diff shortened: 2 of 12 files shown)


lumigator/python/mzai/sdk/completions.py
```diff
@@ -1,3 +1,9 @@
+"""
+Completions SDK
+
+Provides a class to manipulate completions in Lumigator.
```
veekaybee, line 4:
```suggestion
Provides a class to perform LLM vendor-based API inference from within in Lumigator.
```

lumigator/python/mzai/sdk/lm_datasets.py
```diff
@@ -0,0 +1,91 @@
...
+        return [DatasetResponse(**args) for args in response.json()]
+
+    def get_dataset(self, id: str) -> DatasetResponse:
+        """Returns information on a specific dataset.
+        Args:
+            id (str): the id of the dataset to retrieve
+        Returns:
+            DatasetResponse: the dataset information for the provided id.
+        """
+        UUID(id)
```
veekaybee, line 42: we likely don't want to reformat this as a uuid since it will get sent as a string to the response anyway and in the current implementation it's not assigned, aka `id = UUID(id)`

lumigator/python/mzai/sdk/lm_datasets.py
```diff
@@ -0,0 +1,91 @@
...
+        response = self.client.get_response(f"{self.DATASETS_ROUTE}/{id}", HTTPMethod.DELETE)
+
+        if not response:
+            return []
+
+        return DatasetResponse(**(response.json()))
+
+    def create_dataset(self, dataset: IOBase, format: DatasetFormat) -> DatasetResponse:
```
veekaybee, line 50: no need to use IObase here, we can do something as simple as

`files = {'upload_file': open('file.txt','rb')}`
and simply pass in the string of the dataset filename

lumigator/python/mzai/sdk/tests/conftest.py
```diff
@@ -26,7 +27,14 @@ def mock_requests(mock_requests_response):
 
 @pytest.fixture(scope="session")
 def lumi_client() -> LumigatorClient:
-    return LumigatorClient(LUMI_HOST)
+    lumi_c = None
```
veekaybee, line 30: nit: rename to lumi_client so it's clearer what this is?

lumigator/python/mzai/sdk/tests/helpers.py
```diff
@@ -0,0 +1,24 @@
...
+import importlib.resources
+import json
+from pathlib import Path
+from importlib.resources.abc import Traversable
+
+
+LUMI_HOST = "localhost"
+
+
+def load_json(datafile: Traversable) -> str:
```
veekaybee, line 11: same suggestion as in other prs around Traversable, datafile can just be a string or Path

lumigator/python/mzai/sdk/lm_datasets.py
javiermtorres: The idea was to perform client side validation, but I guess a much more logical move is to replace `id: str` with `id: UUID` and force the user to provide something that is really a UUID. Wdyt?

lumigator/python/mzai/sdk/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""
```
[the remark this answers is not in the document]
javiermtorres: ACK, will move this 👍🏻 I was reading that pydoc used this for package docs, but this must not be common practice now.

lumigator/python/mzai/sdk/lm_datasets.py
javiermtorres: Using IOBase allows us to load content from other sources like memory byte buffers. I would like the options open and not force the user to use a file necessarily.

javiermtorres: (Admittedly this is done in tests, but it is still something a user might want to do)

lumigator/python/mzai/sdk/lm_datasets.py
veekaybee: sure, sounds good.

lumigator/python/mzai/sdk/lm_datasets.py
javiermtorres: Thanks a lot :)

veekaybee (approved):
👍

veekaybee:
Looks good once tests pass and main is merged in

lumigator/python/mzai/backend/backend/main.py
```diff
@@ -13,14 +13,23 @@
 import os
 
 
+LUMIGATOR_APP_TAGS = {
+    "title": "Lumigator Backend",
+    "description": "Backend server",
+    # "summary": "TBD",
```
veekaybee, line 19: we can take out summary

lumigator/python/mzai/sdk/client.py
```diff
@@ -50,7 +50,7 @@ def _make_request(
             **kwargs,  # noqa: B026
         )
         response.raise_for_status()
-        if verbose:
+        if verbose and (len(response.content) > 0):
```
veekaybee, line 53: suggestion `is response.content`

PR #258, opened 2024-10-08
Initial commit on LUMI-99- created basic structure

agpituk:
THIS PR IS A DRAFT AND IT WILL CHANGE
## I already...


lumigator/infra/docker-compose/docker-compose.yaml
```diff
@@ -0,0 +1,125 @@
+name: lumi
```
chainlink, line 1: Why are we creating a duplicate docker-compose instead of using the one in .devcontainers?

aittalam: +1 to this, if there is not too much difference and as you are parametrizing this, perhaps we can have one single compose file and two different envs?

PR #259, opened 2024-10-08
SDK: Add 'experiments'

peteski22:
## What's changing

This PR adds support for the Experiments API in the SDK.

## How to test it

Run the PyTest tests locally in your IDE, or `pants test lumigator/python/mzai/sdk:: -- -s`

## I already...

### DIFF
--- a/lumigator/python/mzai/sdk/experiments.py
+++ b/lumigator/python/mzai/sdk/experiments.py
@@ -1,0 +1,77 @@
+from uuid import UUID
+from http import HTTPMethod
+from json import dumps
+
+from schemas.experiments import (
+    ExperimentCreate,
+    ExperimentResponse,
+    ExperimentResultDownloadResponse,
+    ExperimentResultResponse,
+)
+from schemas.extras import ListingResponse
+
+from client import ApiClient
+
+
+class Experiments:
+    EXPERIMENTS_ROUTE = "experiments"
+
+    def __init__(self, c: ApiClient):
+        self.__client = c
+
+    def create_experiment(self, experiment: ExperimentCreate) -> ExperimentResponse:
+        """Creates a new experiment."""
+        response = self.__client.get_response(
+            self.EXPERIMENTS_ROUTE, HTTPMethod.POST, dumps(experiment)
+        )
+
+        if not response:
+            return []
+
+        data = response.json()
+        return ExperimentResponse(**data)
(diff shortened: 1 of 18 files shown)


javiermtorres:
@veekaybee so in [this case](https://github.com/mozilla-ai/lumigator/pull/230#discussion_r1787672804), should we just change the name of the dir? Make it into a module too by adding an `__init__.py`?

lumigator/python/mzai/sdk/experiments.py
```diff
@@ -0,0 +1,77 @@
...
+
+        data = response.json()
+        return ExperimentResponse(**data)
+
+    def get_experiment(self, experiment_id: UUID) -> ExperimentResponse:
+        """Returns information on the experiment for the specified ID."""
+        response = self.__client.get_response(f"{self.EXPERIMENTS_ROUTE}/{experiment_id}")
+
+        if not response:
+            return []
```
veekaybee, line 39: might be worth also adding something additional to indicate to folks whether the issue is at the API level, the client level, etc https://requests.readthedocs.io/en/latest/user/quickstart/#response-status-codes

lumigator/python/mzai/sdk/tests/conftest.py
```diff
@@ -51,3 +58,38 @@ def json_data_jobs() -> Traversable:
 @pytest.fixture(scope="session")
 def json_data_job() -> Traversable:
     return json_data_path() / "job.json"
+
+
+@pytest.fixture(scope="session")
+def json_data_experiments() -> Traversable:
```
veekaybee, line 64: given that we return a fully-qualified path, we might want to return a string or a Path object for simplicity, generally in Python it's idiomatic that file objects are either `Path` or `string` objects, depending on the use-case

chainlink:
@javiermtorres `schemas` will get it's own module back after the pants move 🚚

lumigator/python/mzai/sdk/experiments.py
peteski22: Thanks @veekaybee yep, this is something we probably want to consider more widely across the SDK as a way to convey problems to the consumer. I've raised a ticket for this internally to decide how we want to address it. 😄

veekaybee (approved):
Thanks for this! Approved after review and discussion

PR #261, opened 2024-10-09
Deploy locally on Linux without root access

### LINKED ISSUE #257: Run local deployment on Linux without root
Deploying Lumigator locally in a Linux environment assumes root access, as it attempts to install the necessary Python version in `/opt/python`. Since non-root users usually lack the permission to create directories under `/opt`, the script fails without elevated privileges.

 - OS: Linux (tested on Debian 12.5)
 - Lumigator version: 00227d5 (`git describe --always` output)

**Steps to reproduce the behavior:**
1. Log in to a Linux environment as a non-root user.
2. Clone the repository and check out the provided commit.
3. Install Pants.
4. Run `make bootstrap-dev-environment` from the repo's root directory.

**Expected behavior**
The script should run to completion, creating the local virtual environment (i.e., `mzaivenv`).

**Additional context**
Setting the `PYTHON_INSTALL_DIR` environment variable to a directory that the user has permissions (e.g., `$(HOME)/.local/python`) should resolve the issue. This change touches the [Makefile](https://github.com/mozilla-ai/lumigator/blob/77ff9952362ab546683d0c3779febd90a8252f18/Makefile#L19) and the [`bootstrap_python.sh`](https://github.com/mozilla-ai/lumigator/blob/77ff9952362ab546683d0c3779febd90a8252f18/devtools/shell/bootstrap_python.sh#L31) script

dpoulopoulos:
Deploying locally on Linux assumes root access, as it installs Python in `/opt/python`. Non-root users lack permission to create directories in `/opt`, causing the script to fail.

## What's changing

Change the Python install directory to `$HOME/.local/python`.

## How to test it

1. Log in to a Linux environment as a non-root user.
1. Clone the repository and check out the provided commit.
1. Install Pants.
1. Run make `bootstrap-dev-environment` from the repo's root directory.

Closes #257


dpoulopoulos:
Hi @ividal!

Thanks for getting back to me and for your feedback.

Sure thing, I can make that change and test it locally. I was under the impression that Docker always sets `$HOME`—at least that’s what they claim [here](https://docs.docker.com/engine/containers/run/#environment-variables). However, I understand that someone could potentially unset this in a script, which could cause issues.

Let me test this locally again, and I’ll send you an update!

dpoulopoulos:
@ividal I forced pushed the new changes. I also added a new commit on top fixing a small typo I saw.

Let me know what you think :)

PR #264, opened 2024-10-10
SDK: add `deployments`

javiermtorres:
## I already...

### DIFF
--- a/lumigator/python/mzai/sdk/deployments.py
+++ b/lumigator/python/mzai/sdk/deployments.py
@@ -1,0 +1,67 @@
+from pathlib import Path
+from uuid import UUID
+from http import HTTPMethod
+from json import dumps
+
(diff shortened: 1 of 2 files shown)


veekaybee:
The code looks fine 👍 , but since we're in the process removing `deployments` as an endpoint, I'd suggest we hold off on this one for now because we'll need to remove it.

javiermtorres:
Could you please check https://github.com/mozilla-ai/lumigator/pull/273 then? That one is about integration tests, so it should be mergeable right now. Thanks in advance :)

chainlink:
Can this be closed?

javiermtorres:
I was thinking about either adapting the code to the new jobs endpoint or making it from scratch. Probably starting a new PR is cleaner, so yes, I guess we can close this.

PR #267, opened 2024-10-10
LUMI-53   [Security] Remove secrets from RAY_WORKER_ENV_VARS

chainlink:
## What's changing
Set secrets directly on Ray node instead of via backend (secrets passed via env var of ray jobs api are leaked in the logs, etc) 

Other minor changes:
- AWS_HOST doesn't seem to be set anywhere anymore, so remove from docs,
- Set HF_TOKEN on directly on ray workers now
- Remove setting LOCAL_FSSPEC_S3 env vars on the backend side (only needed by the evaluate ray job)
- Remove secrets from mocks

## How to test it
Run `make local-up`
- [x] Try dataset upload
- [x] Try running eval job

## Related Jira Ticket

https://mzai.atlassian.net/browse/LUMI-53

## I already...

- [x] Verified this works by manually running the above tests

### DIFF
--- a/lumigator/python/mzai/backend/backend/settings.py
+++ b/lumigator/python/mzai/backend/backend/settings.py
@@ -32,18 +32,8 @@ class BackendSettings(BaseSettings):
     RAY_DASHBOARD_PORT: int = 8265
     RAY_SERVE_INFERENCE_PORT: int = 8000
     # the following vars will be copied, if present, from Ray head to workers
-    RAY_WORKER_ENV_VARS: list[str] = [
-        "AWS_ACCESS_KEY_ID",
-        "AWS_SECRET_ACCESS_KEY",
-        "AWS_ENDPOINT_URL",
-        "AWS_HOST",
-        "LOCAL_FSSPEC_S3_KEY",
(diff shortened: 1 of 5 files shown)


lumigator/python/mzai/backend/backend/settings.py
```diff
@@ -32,18 +32,8 @@ class BackendSettings(BaseSettings):
...
-        "AWS_HOST",
-        "LOCAL_FSSPEC_S3_KEY",
-        "LOCAL_FSSPEC_S3_SECRET",
-        "LOCAL_FSSPEC_S3_ENDPOINT_URL",
-        "HF_TOKEN",
-        "OPENAI_API_KEY",
-        "MISTRAL_API_KEY",
-    ]
+    # Secrets should be added directly to ray by setting env vars on the ray head/worker nodes
+    RAY_WORKER_ENV_VARS: list[str] = []
```
aittalam, line 36: This is great, so we don't lose the functionality in case we need it later for non-secret env vars. Thanks!

aittalam (approved):
I am not very familiar with the SDK yet but overall it LGTM, thank you!

PR #273, opened 2024-10-14
Add integration tests

javiermtorres:
## What's changing

Integration tests will check against a running Lumigator instead of mocking responses.

## How to test it

A separate test with address `lumigator/python/mzai/sdk/tests/int_test_datasets.py:sdk_tests` has been provided.

## Related Jira Ticket

LUMI-136

## I already...

### DIFF
--- a/lumigator/python/mzai/sdk/tests/BUILD.pants
+++ b/lumigator/python/mzai/sdk/tests/BUILD.pants
@@ -1,9 +1,0 @@
-python_test_utils(name="sdk_test_utils", sources=["conftest.py", "helpers.py"])
-
-python_tests(
-    name="sdk_tests",
-    sources=["**/test_*.py"],
-    dependencies=[":data-json", "3rdparty/python:3rdparty#pydantic"],
-)
-
-resources(name="data-json", sources=["data/*.json"])
--- a/lumigator/python/mzai/sdk/tests/data/dialogsum_exc.csv
+++ b/lumigator/python/mzai/sdk/tests/data/dialogsum_exc.csv
@@ -1,0 +1,55 @@
+examples,ground_truth
+"#Person1#: Hello, how are you doing today?
+#Person2#: I ' Ve been having trouble breathing lately.
+#Person1#: Have you had any type of cold lately?
+#Person2#: No, I haven ' t had a cold. I just have a heavy feeling in my chest when I try to breathe.
+#Person1#: Do you have any allergies that you know of?
+#Person2#: No, I don ' t have any allergies that I know of.
+#Person1#: Does this happen all the time or mostly when you are active?
+#Person2#: It happens a lot when I work out.
+#Person1#: I am going to send you to a pulmonary specialist who can run tests on you for asthma.
+#Person2#: Thank you for your help, doctor.",#Person2# has trouble breathing. The doctor asks #Person2# about it and will send #Person2# to a pulmonary specialist.
+"#Person1#: Hey Jimmy. Let's go workout later today.
+#Person2#: Sure. What time do you want to go?
+#Person1#: How about at 3:30?
+#Person2#: That sounds good. Today we work on Legs and forearm.
+#Person1#: Hey. I just played basketball earlier, so my legs are a little sore. Let's work out on arms and stomach today.
+#Person2#: I'm on a weekly schedule. You're messing everything up.
+#Person1#: C'mon. We're only switching two days. You can do legs on Friday.
+#Person2#: Aright. I'll meet you at the gym at 3:30 then.",#Person1# invites Jimmy to go workout and persuades him into working out on arms and stomach.
+"#Person1#: I need to stop eating such unhealthy foods.
+#Person2#: I know what you mean. I've started eating better myself.
+#Person1#: What foods do you eat now?
+#Person2#: I tend to stick to fruits, vegetables, and chicken.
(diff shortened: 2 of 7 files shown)


.github/workflows/integration_tests.yaml
```diff
@@ -0,0 +1,68 @@
...
+jobs:
+  integration:
+    name: Integration tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          # gets full repo history
+          fetch-depth: 0
```
chainlink, line 22: I don't think we'll need fetch-depth anymore (that was a pants thing)

javiermtorres: Thanks!!!!!

javiermtorres: I'll leave it there for the moment until we remove pants from the actions (how close are we already?). Otherwise the tests won't work.

.github/workflows/tests.yaml
```diff
@@ -1,4 +1,4 @@
-name: Tests
+name: Integration tests
```
veekaybee, line 1: suggestion: SDK integration tests?

lumigator/python/mzai/sdk/tests/int_test_datasets.py
```diff
@@ -0,0 +1,21 @@
...
+def test_get_datasets_remote_ok(lumi_client):
+    datasets = lumi_client.datasets.get_datasets()
+    assert datasets is not None
+
+
+def test_dataset_lifecycle_remote_ok(lumi_client, hf_data):
+    with Path.open(hf_data) as file:
+        dataset = lumi_client.datasets.create_dataset(dataset=file, format=DatasetFormat.EXPERIMENT)
+        datasets = lumi_client.datasets.get_datasets()
+        assert datasets.total is not 0
```
veekaybee, line 18: do we want to be specific around how many datasets are returned here?

.github/workflows/tests.yaml
javiermtorres: Ok!!!

lumigator/python/mzai/sdk/tests/int_test_datasets.py
javiermtorres: I don't see why not :)

chainlink:
Can you convert these to run via UV (template here https://github.com/mozilla-ai/lumigator/pull/287 )

javiermtorres:
@chainlink Yep, I'll adapt the tests to uv 👍

.github/workflows/integration_tests.yaml
javiermtorres: Removed.

lumigator/python/mzai/sdk/tests/BUILD.pants
```diff
@@ -3,7 +3,14 @@ python_test_utils(name="sdk_test_utils", sources=["conftest.py", "helpers.py"])
 python_tests(
     name="sdk_tests",
     sources=["**/test_*.py"],
+    # sources=["**/test_*.py", "**/int_test_*.py"],
     dependencies=[":data-json", "3rdparty/python:3rdparty#pydantic"],
 )
 
-resources(name="data-json", sources=["data/*.json"])
+python_tests(
```
ividal, line 10: Since I'm looking at this branch for my own local-up issues... Do we need this pants file at all anymore?

javiermtorres: No, removed both BUILD files.

agpituk:
Two things from my side, we currently have 
https://github.com/mozilla-ai/lumigator/blob/f4c871d1890f9560585fff2d3f1872983f57e475/lumigator/python/mzai/sdk/tests/data/thunderbird_gt_bart.csv
and https://github.com/mozilla-ai/lumigator/blob/f4c871d1890f9560585fff2d3f1872983f57e475/lumigator/python/mzai/sdk/thunderbird_gt_bart.csv

I believe is exactly the same file but in different locations. Can't we just use one of them? 
Second, are we happy to put this file here in the public repo? I'm not sure if this conversation has happened already (sorry If I missed it)

aittalam (changes requested):
Thanks @javiermtorres!
Just a few notes about the tb_gt_bart.csv dataset:
- it is in two different places (/sdk and /sdk/tests/data)
- it would be better if we used a public dataset for tests (ping me and I can provide you a csv version of e.g. dialogsum for summarization)
- I think we can have even just a few rows for a smaller footprint

agpituk (approved):
happy with the latest changes :)

lumigator/python/mzai/sdk/tests/data/dialogsum_exc.csv
```diff
@@ -0,0 +1,55 @@
+examples,ground_truth
```
ividal, line 1: Since both SDK and BE tests are now using this, how about placing it in an upper level, like `python/mzai/sample_data`? (Not sold on the name or location, be my guest 😁 )

ividal: Sure! I can move it on #290, for example. The sooner we get some of these tests in, the better :)

Thank you @javiermtorres ! 🙌

PR #274, opened 2024-10-14
Added GT field recording whether a dataset contains GT

aittalam:
## What's changing
A new field is added to the datasets table recording whether GT is included within a dataset.
Exposing this information allows the UI to act accordingly (e.g. warning the user before running an eval) without downloading the whole dataset.

## How to test it
- unit tests: pants test lumigator/python/mzai/backend/backend/tests:backend_tests@parametrize=darwin
- integration tests w curl

## Related Jira Ticket
https://mzai.atlassian.net/browse/LUMI-88

I added @peteski22 and @javiermtorres to get specific SDK-related feedback 🙏 

## I already...

- [ x] added some tests for any new functionality: UNIT TEST

### DIFF
--- a/lumigator/python/mzai/backend/backend/records/datasets.py
+++ b/lumigator/python/mzai/backend/backend/records/datasets.py
@@ -1,8 +1,8 @@
+from schemas.datasets import DatasetFormat
 from sqlalchemy.orm import Mapped
 
 from backend.records.base import BaseRecord
 from backend.records.mixins import CreatedAtMixin
-from schemas.datasets import DatasetFormat
 
 
 class DatasetRecord(BaseRecord, CreatedAtMixin):
@@ -11,3 +11,4 @@ class DatasetRecord(BaseRecord, CreatedAtMixin):
     filename: Mapped[str]
     format: Mapped[DatasetFormat]
     size: Mapped[int]
+    gt: Mapped[bool]
--- a/lumigator/python/mzai/backend/backend/records/datasets.py
+++ b/lumigator/python/mzai/backend/backend/records/datasets.py
@@ -11,4 +11,4 @@ class DatasetRecord(BaseRecord, CreatedAtMixin):
     filename: Mapped[str]
     format: Mapped[DatasetFormat]
     size: Mapped[int]
-    gt: Mapped[bool]
+    ground_truth: Mapped[bool]
--- a/lumigator/python/mzai/backend/backend/services/datasets.py
+++ b/lumigator/python/mzai/backend/backend/services/datasets.py
@@ -156,7 +156,7 @@ class DatasetService:
 
             # Create DB record
             record = self.dataset_repo.create(
-                filename=dataset.filename, format=format, size=actual_size, gt=has_gt
+                filename=dataset.filename, format=format, size=actual_size, ground_truth=has_gt
             )
 
             # convert the dataset to HF format and save it to S3
--- a/lumigator/python/mzai/backend/backend/services/datasets.py
+++ b/lumigator/python/mzai/backend/backend/services/datasets.py
@@ -10,15 +10,16 @@ from loguru import logger
 from mypy_boto3_s3.client import S3Client
 from pydantic import ByteSize
 from s3fs import S3FileSystem
+from schemas.datasets import DatasetDownloadResponse, DatasetFormat, DatasetResponse
+from schemas.extras import ListingResponse
 
 from backend.records.datasets import DatasetRecord
 from backend.repositories.datasets import DatasetRepository
 from backend.settings import settings
-from schemas.datasets import DatasetDownloadResponse, DatasetFormat, DatasetResponse
-from schemas.extras import ListingResponse
 
-ALLOWED_EXPERIMENT_FIELDS: set[str] = {"examples", "ground_truth"}
+GT_FIELD: str = "ground_truth"
 REQUIRED_EXPERIMENT_FIELDS: set[str] = {"examples"}
+ALLOWED_EXPERIMENT_FIELDS: set[str] = {"examples", GT_FIELD}
 
 
 def validate_file_size(input: BinaryIO, output: BinaryIO, max_size: ByteSize) -> int:
@@ -83,6 +84,15 @@ def validate_experiment_dataset(filename: str):
             )
 
 
+def dataset_has_gt(filename: str) -> bool:
+    with Path(filename).open() as f:
+        reader = csv.DictReader(f)
+        fields = set(reader.fieldnames or [])
+        has_gt = GT_FIELD in fields
+
+    return has_gt
+
+
 class DatasetService:
     def __init__(
         self, dataset_repo: DatasetRepository, s3_client: S3Client, s3_filesystem: S3FileSystem
@@ -141,11 +151,12 @@ class DatasetService:
             # Validate format
             validate_dataset_format(temp.name, format)
 
+            # Verify whether the dataset contains ground truth
+            has_gt = dataset_has_gt(temp.name)
+
             # Create DB record
             record = self.dataset_repo.create(
-                filename=dataset.filename,
-                format=format,
-                size=actual_size,
+                filename=dataset.filename, format=format, size=actual_size, gt=has_gt
             )
 
             # convert the dataset to HF format and save it to S3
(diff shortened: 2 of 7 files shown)


lumigator/python/mzai/sdk/tests/data/dataset.json
```diff
@@ -3,5 +3,6 @@
     "filename": "test.hf",
     "format": "experiment",
     "size": 16,
+    "gt": true,
```
veekaybee, line 6: suggestion: ground_truth instead of gt so it's clearer

lumigator/python/mzai/sdk/tests/data/datasets.json
```diff
@@ -6,20 +6,23 @@
             "filename": "ds1.hf",
             "format": "experiment",
             "size": 16,
+            "gt": true,
```
veekaybee, line 9: same here suggestion: ground_truth instead of gt so it's clearer

lumigator/python/mzai/backend/backend/records/datasets.py
```diff
@@ -11,3 +11,4 @@ class DatasetRecord(BaseRecord, CreatedAtMixin):
     filename: Mapped[str]
     format: Mapped[DatasetFormat]
     size: Mapped[int]
+    gt: Mapped[bool]
```
veekaybee, line 14: suggestion: let's call it ground_truth throughout for clarity

lumigator/python/mzai/backend/backend/services/datasets.py
```diff
@@ -83,6 +84,15 @@ def validate_experiment_dataset(filename: str):
             )
 
 
+def dataset_has_gt(filename: str) -> bool:
```
veekaybee, line 87: we should pass in a complete path here and then pass that into the with open

```
with open(filename, r) as f
```

are we also assuming that the file is a csv here? we might want to specify or do a type check before we open/assign

aittalam: I am using the same identical interface of `validate_experiment_dataset` here for consistency.
We assume the file is a CSV as we have already checked it in `validate_dataset_format`.

lumigator/python/mzai/backend/schemas/datasets.py
```diff
@@ -19,4 +19,5 @@ class DatasetResponse(BaseModel, from_attributes=True):
     filename: str
     format: DatasetFormat
     size: int
+    gt: bool
```
javiermtorres, line 22: ground_truth too please ^_^

lumigator/python/mzai/sdk/tests/data/datasets.json
javiermtorres, line 9: Thanks a lot for providing the SDK test data too ^_^

lumigator/python/mzai/backend/backend/tests/api/routes/test_datasets.py
```diff
@@ -102,8 +110,8 @@ def test_presigned_download(app_client: TestClient, valid_experiment_dataset):
 
 def test_experiment_format_validation(
```
ividal, line 111: What do you think of using mark.parametrize here instead of the for loop over datasets? 
You know the input you want to check (and it's just one at a time), you know the output you expect, so it could be:

```
@pytest.mark.parametrize(
    "dataset, expected_output", 
    [(missing, status.HTTP_403_FORBIDDEN),
    (extra, status.HTTP_403_FORBIDDEN)]
def test_experiment_format_validation(app_client, dataset, expected):
....
```

And it will become easy to add cases as we move into dataset checks later on.

lumigator/python/mzai/backend/backend/tests/api/routes/test_datasets.py
```diff
@@ -113,3 +121,27 @@ def test_experiment_format_validation(
             files={"dataset": ("dataset.csv", d)},
         )
         assert response.status_code == status.HTTP_403_FORBIDDEN
+
+
+def test_experiment_ground_truth(
```
ividal, line 126: Thanks for including this one! 😄

lumigator/python/mzai/backend/backend/services/datasets.py
```diff
@@ -141,11 +151,12 @@ def upload_dataset(self, dataset: UploadFile, format: DatasetFormat) -> DatasetR
...
 
+            # Verify whether the dataset contains ground truth
+            has_gt = dataset_has_gt(temp.name)
+
             # Create DB record
             record = self.dataset_repo.create(
-                filename=dataset.filename,
-                format=format,
-                size=actual_size,
+                filename=dataset.filename, format=format, size=actual_size, gt=has_gt
```
ividal, line 159: It seems obvious what it is when you're deep into it, but here's another vote to using ground_truth wherever there is gt... 😅

ividal (changes requested):
Thanks a lot @aittalam ! Some minor requests to make it easier to add later :)

lumigator/python/mzai/backend/backend/services/datasets.py
```diff
@@ -83,6 +84,15 @@ def validate_experiment_dataset(filename: str):
             )
 
 
+def dataset_has_gt(filename: str) -> bool:
+    with Path(filename).open() as f:
+        reader = csv.DictReader(f)
+        fields = set(reader.fieldnames or [])
+        has_gt = GT_FIELD in fields
```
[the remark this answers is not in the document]
aittalam: I'd argue that here the semantics is a bit different, i.e. we are validating the existence of a field rather than rows being consistent with a schema (see examples below).

However, I think the idea is neat and I'd like to include it into another PR. I'd extend this to the whole dataset (instead of just the first row) and use to make sure all of our data is consistent with the schema we provide. WDYT?

---

This is the code I tested with:

```
class CSVFields(BaseModel):
    examples: str
    ground_truth: str | None = None


def validate_ground_truth(filename) -> bool:
    has_ground_truth = False

    with Path(filename).open() as f:
        reader = csv.DictReader(f)
        first_row = next(reader, None)
        logger.info(f"First row: {first_row}")
        if first_row:
            try:
                validated_row = CSVFields.model_validate(first_row)
                logger.info(f"Validated row: {validated_row}")
                has_ground_truth = validated_row.ground_truth is not None
            except ValidationError as e:
                print(f"Validation error: {e}")

    logger.info(f"has_ground_truth: {has_ground_truth}")
    return has_ground_truth
```

And this what I get with different inputs:

INPUT (no gt):
```
examples
"This is an example"
"This is another example"
```
OUTPUT:
```
First row: {'examples': 'This is an example'}
Validated row: examples='This is an example' ground_truth=None
has_ground_truth: False
```

---


INPUT (gt field, one row without gt):
```
examples,ground_truth
"This is an example", "This is a summary"
"This is another example"
```
OUTPUT:
```
First row: {'examples': 'This is an example', 'ground_truth': ' "This is a summary"'}
Validated row: examples='This is an example' ground_truth=' "This is a summary"'
has_ground_truth: True
```

---

INPUT (same file as before, flipped rows)
```
examples,ground_truth
"This is another example"
"This is an example", "This is a summary"
```

OUTPUT:
```
First row: {'examples': 'This is another example', 'ground_truth': None}
Validated row: examples='This is another example' ground_truth=None
has_ground_truth: False
```

---

INPUT (same file as before, one extra comma)
```
examples,ground_truth
"This is another example",
"This is an example", "This is a summary"
```

OUTPUT:
```
First row: {'examples': 'This is another example', 'ground_truth': ''}
Validated row: examples='This is another example' ground_truth=''
has_ground_truth: True```

---

lumigator/python/mzai/backend/backend/tests/api/routes/test_datasets.py
aittalam: Oh this is nice! Added it, thanks for the suggestion 🙏

ividal (approved):

lumigator/python/mzai/backend/backend/services/datasets.py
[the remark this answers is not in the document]
veekaybee: Sounds good 👍

veekaybee (approved):

PR #275, opened 2024-10-14
Update local installation for non developers 

agpituk:
## What's changing

- Simplified the Lumigator setup by using a pre-built external container, removing the need for users to set up a local Python environment.
- Added new Makefile entry points to streamline running and managing the app.
- Maintained support for the original local setup for backward compatibility.

## How to test it
Run both versions of lumigator, the old one with make-local-up and the new one with start-lumigator and test the app
## I already...

- [N/A ] added some tests for any new functionality;

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,10 +81,30 @@ It consists of:
 + A Postgres database to track platform-level tasks and dataset metadata
 
 # Get Started
+You can run the project locally using `docker-compose` on Mac or Linux, or into a distributed environment using Kubernetes [`Helm charts`](lumigator/infra/mzai/helm/lumigator/README.md)
 
-You can build the local project using `pants` and `docker-compose` on Mac or Linux,  or into a distributed environment using Kubernetes [`Helm charts`](lumigator/infra/mzai/helm/lumigator/README.md)
+## Prerequisites:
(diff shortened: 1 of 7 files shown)


agpituk:
Right now we're using an image @macaab26 built for me, but I'm not super comfortable with that, we'll improve the process in follow-ups where we'll use an image that's built from Main using github actions

agpituk:
Moved back to draft after some initial feedback from @aittalam , will reopen it as soon as it gets fixed

README.md
```diff
@@ -81,10 +81,30 @@ It consists of:
...
+1. `git clone git@github.com:mozilla-ai/lumigator.git`
+2. `make start-lumigator`
+3. The application should be available at http://localhost:8000. 
+
+# Running Lumigator with an external Ray cluster
+To run Lumigator with an external Ray cluster, ensure the ray host and port are configured in [`.env.example.`](.env.example) and then use the following command:
+
+`make start-lumigator-external-ray`
+
+## Local Development Requirements
```
veekaybee, line 106: it would make sense to move this into the contributing section so that folks working with the app know how to build it, and leave a link in the main README

PR #279, opened 2024-10-15
[LUMI-101]Adding build & push new action for uv

macaab26:
## I already...


.github/workflows/build_images_uv.yaml
```diff
@@ -0,0 +1,71 @@
...
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build the Docker image
+        run: |
+          docker build ./lumigator/python/mzai/backend/ --file ./lumigator/python/mzai/backend/Dockerfile.uv --tag mzdotai/lumigator:backend_dev_$REPOVERSION
+
+      - name: Create an image with latest tag
+        run: docker tag mzdotai/lumigator:backend_dev_$REPOVERSION mzdotai/lumigator:latest
```
javiermtorres, line 65: Shouldn't this be just `mzdotai/lumigator:$REPOVERSION` if we got for semver? Maybe `mzdotai/lumigator_backend:$REPOVERSION`?

.github/workflows/build_images_uv.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+# https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services
+permissions:
+  id-token: write
+  contents: read # This is required for actions/checkout
+
+jobs:
+  build:
+    name: Build Docker image
+    runs-on: ubuntu_2204_150gb # this is a mzai-custom image on github actions runners.
```
chainlink, line 28: We probably don't need to use a custom runner anymore

.github/workflows/build_images_uv.yaml
```diff
@@ -0,0 +1,71 @@
...
+    env:
+      # head_ref: source branch of the pull request.
+      # only set on PRs; e.g., for a PR from branch `myname/amazing_model`, will be `myname/amazing_model`
+      # ref_name: short ref name of branch triggering runs. works for pushes to a development branch.
+      # assignment works starting with the head_ref
+      BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
```
chainlink, line 34: It doesn't look like this is used anywhere and can be removed.

.github/workflows/build_images_uv.yaml
```diff
@@ -0,0 +1,71 @@
...
+    # set up a local registry
+    services:
```
chainlink, line 36: I don't think we'll need the local registry anymore

.github/workflows/build_images_uv.yaml
```diff
@@ -0,0 +1,71 @@
...
+    branches:
+      - main
+  # required to enable manual triggers on the GH web ui
+  workflow_dispatch:
+
+env:
+  AWS_REGION: us-east-2
+
+# https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services
+permissions:
```
chainlink, line 21: We don't need this section (was for pushing to ECR)

.github/workflows/build_images_uv.yaml
```diff
@@ -0,0 +1,71 @@
+name: Docker Image CI
+
+on:
+  pull_request:
+    paths:
+      - "lumigator/**"
+      - ".github/**"
+    branches: ["main"]
```
chainlink, line 8: I don't think we need the `branch` section under `pull_request`

chainlink (approved):
Some small changes, but otherwise looks good - pre-approving once those are in

.github/workflows/build_images_uv.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build the Docker image
+        run: |
+          docker build ./lumigator/python/mzai/backend/ --file ./lumigator/python/mzai/backend/Dockerfile.uv --tag mzdotai/lumigator:backend_dev_$REPOVERSION
```
agpituk, line 62: @chainlink is this the right Dockerfile? I thought the new uv dockerfile was the one in the root folder (the one used by the Docker-compose)

.github/workflows/build_images_uv.yaml
chainlink: I'm good with the short tag for now

macaab26: Hey @javiermtorres, yep agree, I don't like that style of tagging. Will change it. But just to make sure that we are on the same page, $REPOVERSION is a hash. The versions tagged with semver will be those that we will release “publicly”. Are we okay with that? @chainlink @agpituk @peteski22

.github/workflows/build_images_uv.yaml
macaab26: 100% good catch!

.github/workflows/build_images_uv.yaml
macaab26: Yeah, you're right

.github/workflows/build_images_uv.yaml
macaab26: I leave it because I only want to create new images from main. If someone creates a PR with two different branches (not main) involved, I don't want to generate a new image. WDYT?

.github/workflows/build_images_uv.yaml
macaab26: Yep, let me try

.github/workflows/build_images_uv.yaml
javiermtorres: I'd support that. It's not common but sometimes we will want PRs among non-main branches for some complex features being developed by more than one person.

.github/workflows/build_images_uv.yaml
javiermtorres: Ah ok.... So this would be for dev images.... Please then say

> Create an image with latest *commit hash*

Otherwise the reader may think it's only for tags. And maybe instead of running the docker command directly, you can use the docker github actions, which make it way easier to create and tag images with git info: https://docs.docker.com/build/ci/github-actions/manage-tags-labels/

.github/workflows/build_images_uv.yaml
macaab26: Cool, will change it

.github/workflows/build_images_uv.yaml
chainlink: This PR just focuses on dev versions which use the hash. We can tackle the tagged release process in another PR (when we have a release process)

.github/workflows/build_images_uv.yaml
macaab26: It worked :)

PR #280, opened 2024-10-15
Remove summarizer, ground truth deployment, and Ray serve

veekaybee:
## What's changing

We're working towards migrating from Ray Serve to offline inference via vLLM for our ground truth generation. This involves three moving pieces: 

1. Getting rid of summarizer deployment and all related ground truth service assets since we won't be serving the model as a stand-alone web application but running in-process batch async evaluation.  **<-- this PR**
2. Moving `experiments` logically to jobs since we now have granular inference and evaluation jobs instead of full-length evaluation experiments. 
3. Creating a new module, `inference`, that allows for only the generation of ground truth in the same way as an evaluation job. 

**I won't be merging this PR until the other ground truth functionality is ready and working, including tests, but wanted to get it approved in small chunks** 

## How to test it

You should no longer see a /groundtruth endpoint and not be able to hit it via cURL
Screenshot 2024-10-15 at 11 33 38 AM

## I already...

- [X] added some tests for any new functionality - removed tests for deployment components
- [X] updated the documentation.

cc @javiermtorres @peteski22 FYI

PR #285, opened 2024-10-16
[LUMI-37] Refactor experiments service into jobs and create inference and evaluation jobs

veekaybee:
## What's changing

As part of our broader work to refactor to enable multiple types of evaluation and inference jobs, we are creating new endpoints. In machine learning, there are two ways to perform model inference, _online ad-hoc_ and _offline in batch_. Right now within Lumigator we only need offline to perform evaluation and ground truth generation jobs. These are both batch workflows that don't currently require a model hosted beyond the duration of the batched inference. 

However previously, our ground truth job was using Ray Serve in an online manner. Ray Serve has been difficult to work with and orthogonal to our needs, and so we removed that component in this PR. https://github.com/mozilla-ai/lumigator/pull/280

Screenshot 2024-10-21 at 2 46 06 PM

In order to enable LLM inference without evaluation, we need to create a more granular API entrypoint than is currently offered by `experiments`, so we are creating an entrypoint, `jobs`, that allows for `inference` and `evaluation` jobs at a granular level. 

This PR: 

+ renames the `experiments` endpoint to jobs, offers a job type, `inference` and `evaluation` 
+ Creates a `jobs` service that allows for inference and evaluation jobs based on the route path 
+ UUIDs are now assigned  at the per-job level. 

## How to test it

+ Run local unit and integration tests by cd into backend and running `SQLALCHEMY_DATABASE_URL=sqlite:///local.db uv run pytest`
+ `make local-up` and run a job

+ We will need to refactor the SDK as a result of the logic change in experiments

## I already...

- [X] added some tests for any new functionality;
- [X] updated the documentation.

fyi @peteski22 and @javiermtorres for future SDK impact (i.e. we'll need to rename `experiments`) and fork logic for testing inference and evaluation routes under `jobs`. Happy to help with this once this PR is merged.


veekaybee:
Note: I'm still working on fixing the tests but wanted to get some feedback ahead of time as the code is complete at this point.

lumigator/python/mzai/backend/backend/api/deps.py
```diff
@@ -50,16 +49,16 @@ def get_dataset_service(
...
+) -> JobService:
+    job_repo = JobRepository(session)
+    result_repo = JobResultRepository(session)
     ray_client = JobSubmissionClient(settings.RAY_DASHBOARD_URL)
-    return ExperimentService(experiment_repo, result_repo, ray_client, dataset_service)
+    return JobService(job_repo, result_repo, ray_client, dataset_service)
 
 
-ExperimentServiceDep = Annotated[ExperimentService, Depends(get_experiment_service)]
+JobServiceDep = Annotated[JobService, Depends(get_job_service)]
```
veekaybee, line 61: Note we're changing the service dependency here.

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -0,0 +1,61 @@
...
+    JobResultResponse,
+)
+
+from backend.api.deps import JobServiceDep
+
+router = APIRouter()
+
+
+@router.post("/inference", status_code=status.HTTP_201_CREATED)
+def create_inference_job(
```
veekaybee, line 18: Note that we now have two endpoints, `jobs/inference` and `jobs/evaluate`

lumigator/python/mzai/backend/backend/config_templates.py
```diff
@@ -1,5 +1,6 @@
-seq2seq_template = """{{
-    "name": "{experiment_name}/{experiment_id}",
+# Evaluation templates
```
veekaybee, line 1: Note that we now have two sets of config templates, one for evaluation and one for inference

lumigator/python/mzai/backend/backend/config_templates.py
```diff
@@ -61,15 +62,60 @@
...
+            "engine": "{model_path}",
+            "system_prompt": "{system_prompt}",
+            "max_retries": 3
+        }}
+    }},
+    "dataset": {{ "path": "{dataset_path}" }},
+}}"""
+
+config_infer_template = {
+    "hf://facebook/bart-large-cnn": bart_infer_template,
```
veekaybee, line 100: I made the inference set of models less than the ones available for evaluation per our previous conversations that ground truth should offer a small set of models, let me know what you think @aittalam

lumigator/python/mzai/backend/backend/records/jobs.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class JobRecord(BaseRecord, NameDescriptionMixin, JobStatusMixin, DateTimeMixin):
+    __tablename__ = "jobs"
+
+
+class JobResultRecord(BaseRecord):
+    __tablename__ = "job-results"
+
+    job_id: Mapped[uuid.UUID] = mapped_column(ForeignKey("jobs.id"), unique=True)
```
veekaybee, line 18: Note that here we now set the job_id as the UUID rather than at the experiment level

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -0,0 +1,315 @@
...
+        record = self._get_job_record(job_id)
+
+        return str(
+            Path(settings.S3_JOB_RESULTS_PREFIX)
+            / settings.S3_JOB_RESULTS_FILENAME.format(
+                job_name=record.name, job_id=record.id
+            )
+        )
+
+    def create_inference_job(self, request: JobCreate, type: str) -> JobResponse:
```
veekaybee, line 74: note that we create a type of job based on the path prefix in the API request.

lumigator/python/mzai/backend/backend/tests/README.md
```diff
@@ -6,7 +6,7 @@ The types of tests can be broken down into two distinct categories:
 - `integration` tests that depend on some external service to run
 
 The external services that the application depends on are
-the Postgres database, Ray cluster, and S3 storage.
+the SQLite database, Ray cluster, and S3 storage.
```
veekaybee, line 9: Let me know what you think about this change per our convo @chainlink

lumigator/python/mzai/backend/backend/tests/api/routes/test_api_workflows.py
```diff
@@ -5,21 +5,16 @@
 from fastapi.testclient import TestClient
 from schemas.datasets import DatasetFormat, DatasetResponse
 
+from backend.main import app
+
 
 # First test waits and polls up to 10 times for the real backend to be up.
+@app.on_event("startup")
```
veekaybee, line 12: Injecting a real instance of the app on startup, per discussion here https://github.com/mozilla-ai/lumigator/pull/290#discussion_r1808768658

lumigator/python/mzai/evaluator/evaluator/configs/huggingface.py
```diff
@@ -112,7 +112,7 @@ def as_huggingface(self) -> PeftConfig:
 
 
 class QuantizationConfig(EvaluatorConfig):
-    """Basic quantization settings to pass to training and evaluation jobs.
+    """Basic quantization settings to pass to evaluation jobs.
```
veekaybee, line 115: removing notes on finetuning here since we don't do that anymore

lumigator/python/mzai/backend/backend/tests/api/routes/test_api_workflows.py
```diff
@@ -46,6 +41,6 @@ def test_upload_data_launch_experiement(local_client: TestClient, dialog_dataset
         "config_template": "string",
     }
 
-    create_experiment_response = local_client.post("/experiments", headers=headers, json=payload
+    create_experiment_response = local_client.post("/jobs/evaluate", headers=headers, json=payload
```
veekaybee, line 44: Could add another test here for jobs/inference as well.

veekaybee:
Tested and running locally but I may need to look into adapting arch
Screenshot 2024-10-21 at 9 08 42 PM

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -0,0 +1,315 @@
+from pathlib import Path
```
veekaybee, line 1: Note that on a next pass, we can refactor a lot of this repeated code at the type of job logic split into a common shared class between all types of jobs. Since this is a big change, I'm only doing a straight duplication to make sure it all works first.

lumigator/python/mzai/backend/backend/api/deps.py
```diff
@@ -10,10 +10,9 @@
 
 from backend.db import session_manager
 from backend.repositories.datasets import DatasetRepository
-from backend.repositories.experiments import ExperimentRepository, ExperimentResultRepository
 from backend.services.completions import MistralCompletionService, OpenAICompletionService
 from backend.services.datasets import DatasetService
-from backend.services.experiments import ExperimentService
+from backend.services.jobs import JobRepository, JobResultRepository, JobService
```
aittalam, line 15: Are `JobRepository`, `JobResultRepository` in `backend.services.jobs` or `backend.repositories.jobs`?

```suggestion
from backend.services.jobs import JobService
from backend.repositories.jobs import JobRepository, JobResultRepository
```

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -0,0 +1,61 @@
+from uuid import UUID
+
+from fastapi import APIRouter, status
+from schemas.extras import ListingResponse
+from schemas.jobs import (
+    JobCreate,
+    JobResponse,
+    JobResultDownloadResponse,
+    JobResultResponse,
```
aittalam, line 9: Just a note for later on that we are not currently using Job results as everything's stored in the output files and we might want to remove this when we move to proper experiment tracking / artifact storage.

lumigator/python/mzai/backend/backend/config_templates.py
aittalam: I agree on GT offering a smaller set of models but I would not constrain it here, the main reason being batch inference as a first step for evaluation (so we need to support all the models we currently support in eval).
We can think about this list as "all the models we want to make available for batch inference", and then constrain which ones we suggest for GT via the API (i.e. with a `(task, model_list)` mapping, similarly to what we want to do for eval itself). WDYT?

lumigator/python/mzai/backend/backend/services/jobs.py
aittalam: That's perfectly fine! I myself am experimenting with inference batch jobs by simply modifying the current eval code, and I see that a good part of it is common.
I'd say let's do this round, see what differs between different job calls, then finally factor this out and make it parametric wrt the job type.

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -0,0 +1,315 @@
...
+        # set num_gpus per worker (zero if we are just hitting a service)
+        if not request.model.startswith("hf://"):
+            worker_gpus = settings.RAY_WORKER_GPUS_FRACTION
+        else:
+            worker_gpus = settings.RAY_WORKER_GPUS
+
+        runtime_env = {
+            "pip": settings.PIP_REQS,
+            "working_dir": settings.EVALUATOR_WORK_DIR,
+            "env_vars": runtime_env_vars,
```
aittalam, line 149: A heads up that we'll have different `PIP_REQS`, `EVALUATOR_WORK_DIR`, and (well, of course) `infer_command` here.
How would you prefer to proceed? Happy to send a PR to this branch so we can test everything together, or wait for this to be merged and send a PR to main later on

ividal (changes requested):

lumigator/python/mzai/backend/backend/api/routes/experiments.py
```diff
@@ -1,54 +0,0 @@
...
-    ExperimentResultResponse,
-)
-from schemas.extras import ListingResponse
-
-from backend.api.deps import ExperimentServiceDep
-
-router = APIRouter()
-
-
-@router.post("/", status_code=status.HTTP_201_CREATED)
```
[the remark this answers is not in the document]
veekaybee: Per our discussion, adding a dummy `/experiments` route for now while we decide on next steps

lumigator/python/mzai/backend/backend/api/routes/jobs.py
veekaybee: should I take this out in that case, wdyt?

lumigator/python/mzai/backend/backend/services/jobs.py
veekaybee: Let's merge this for now and change it in the next one so we can fully test it, does that work for you?

lumigator/python/mzai/backend/backend/settings.py
```diff
@@ -12,20 +12,13 @@ class BackendSettings(BaseSettings):
     DEPLOYMENT_TYPE: DeploymentType = DeploymentType.LOCAL
     MAX_DATASET_SIZE: ByteSize = "50MB"
 
-    # Postgres
-    POSTGRES_HOST: str = "localhost"
```
veekaybee, line 16: @chainlink are we ok to remove this?

lumigator/python/mzai/backend/backend/config_templates.py
veekaybee: will eval and batch inference take the same parameters? The reason I split them out is that for ground truth inference (or others) we don't need the eval metrics.

lumigator/python/mzai/backend/backend/config_templates.py
veekaybee: Yup, that sounds good, left this list in-tact for now

lumigator/python/mzai/backend/backend/config_templates.py
ividal: > eval and batch inference take the same parameters

I wouldn't expect it. Let's see what you all think - I would imagine batch inference to just need the parameters the model needs to generate content (and this could be just predictions to be tested or it could be "ground-truth" from a reference model). The evaluation would take it from there, separately (metric, params the metric needs, references to the data it has to compare/measure).


> reason I split them out is that for ground truth inference (or others) we don't need the eval metrics

Completely agree.

lumigator/python/mzai/backend/backend/settings.py
veekaybee: Keeping it in for now in lieu of a larger postgres refactor

lumigator/python/mzai/backend/backend/config_templates.py
aittalam: I agree on splitting jobs, for this reason I did not see evaluation needing a list of model configurations (as an inference job will be used to run inference for that down the line).
But I think it's ok to keep it like this (separate configs) for now, and then when eval only does evals we can remove redundant eval model_configs

lumigator/python/mzai/backend/backend/api/routes/jobs.py
aittalam: I think this is out of this PR's scope, but I'll add an issue for that

lumigator/python/mzai/backend/backend/services/jobs.py
aittalam: Absolutely, go with it and I can change it (well I have done it already :-)) in my PR!

PR #290, opened 2024-10-17
Initial backend integration tests involving experiments 

ividal:
## What's changing
Currently we are testing the backend manually with curl commands when the whole app is up. This PR introduces a way to chain backend tests without dummies. The goal is to make it easier to add realistic integration tests close to the backend. 

We can use the same set-up to add any other flows we're currently testing manually.

## How to test it

`SQLALCHEMY_DATABASE_URL=sqlite:///local.db uv run pytest -v`
The addition is the test itself.

N/A

## I already...


lumigator/python/mzai/backend/backend/tests/BUILD.pants
```diff
@@ -6,4 +6,17 @@ python_sources(
 python_tests(
```
veekaybee, line 6: same here, don't need pants anymore

lumigator/python/mzai/backend/backend/tests/api/routes/test_datasets.py
```diff
@@ -113,3 +113,35 @@ def test_experiment_format_validation(
             files={"dataset": ("dataset.csv", d)},
         )
         assert response.status_code == status.HTTP_403_FORBIDDEN
+
+
+def test_upload_csv_delete(app_client: TestClient):
+    response = app_client.get("/health")
+    assert response.status_code == status.HTTP_200_OK
+    cwd = Path(__file__).parent
```
veekaybee, line 121: Let's pull in all datasets as fixtures from a `conftest` file following the existing pattern here https://github.com/mozilla-ai/lumigator/blob/86cc96c2c8da01e8b1b798adc4f451f4a02149a6/lumigator/python/mzai/evaluator/evaluator/tests/conftest.py#L27

lumigator/python/mzai/backend/backend/tests/api/routes/test_datasets.py
```diff
@@ -113,3 +113,35 @@ def test_experiment_format_validation(
             files={"dataset": ("dataset.csv", d)},
         )
         assert response.status_code == status.HTTP_403_FORBIDDEN
+
+
+def test_upload_csv_delete(app_client: TestClient):
+    response = app_client.get("/health")
+    assert response.status_code == status.HTTP_200_OK
```
veekaybee, line 120: I don't think we need to test the health endpoint here

lumigator/python/mzai/backend/backend/tests/api/routes/test_e2e.py
```diff
@@ -0,0 +1,48 @@
+from pathlib import Path
+
+import requests
+from fastapi import status
+from fastapi.testclient import TestClient
+from schemas.datasets import DatasetFormat, DatasetResponse
+
+CWD = Path(__file__).parent
```
veekaybee, line 8: Same comment here, let's use pytest fixtures for this from `conftest`

lumigator/python/mzai/backend/backend/tests/api/routes/test_e2e.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def test_e2e_launched_backend():
```
veekaybee, line 11: let's set up these tests using FastAPI's `TestClient` which does a lot of the heavy lifting for you
https://fastapi.tiangolo.com/tutorial/testing/#using-testclient

.github/workflows/integration_tests.yaml
```diff
@@ -0,0 +1,61 @@
...
+          fetch-depth: 0
+
+      - name: Initialize Pants
+        uses: pantsbuild/actions/init-pants@main
+        with:
+          # cache0 makes it easy to bust the cache if needed
+          gha-cache-key: pants-py3.11.9-v0
+          # The Python backend uses named_caches for Pip/PEX state,
+          # so it is appropriate to invalidate on lockfile changes.
+          named-caches-hash: ${{ hashFiles('3rdparty/python/linux_cpu.lock', '.python') }}
```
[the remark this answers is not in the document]
ividal: Yup! I realised the consequences of #289 after opening the PR. I rebased now - a lot of this is going away and I'll remove the "draft" label.

lumigator/python/mzai/backend/backend/tests/api/routes/test_datasets.py
ividal: You'd be surprised (or not!) but this one is helpful as a sanity check: it helped see a 404 I was getting was due to me messing with how I was handling URLs, so it might help someone else.

lumigator/python/mzai/backend/backend/tests/api/routes/test_e2e.py
ividal: That's what we're doing for test_dataset et al. and the dummy tests. Here I was going for the simplicity of not creating a new TestClient - but what I can do is see if we can reuse the other one.

ividal:
⚠️ Rebased - pants stuff gone and PR became a reasonable size now.
Draft label still on while I look at the tests that require local-up.

.github/workflows/integration_tests.yaml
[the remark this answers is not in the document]
ividal: Pants stuff was indeed gone upon rebasing :)

lumigator/python/mzai/backend/backend/tests/api/routes/test_e2e.py
ividal: And upon second thought, also to keep it consistent with the rest of the codebase, we're now using a TestClient that talks to the real backend on localhost.

.github/workflows/tests_uv.yaml
```diff
@@ -18,6 +18,12 @@ jobs:
         run: uv python install
         working-directory: lumigator/python/mzai/backend
 
+      - name: Start backend
+        run: make start-lumigator
```
chainlink, line 22: We should be forcing a build here so we're testing against the current version of the code. I think there's a way to specify both `build` and `image` in the docker-compose file, but that will require a bit more testing. Overall agree with using the non-mounted image though

ividal: Hum, should we add a build as a previous step?

So far, it's been testing the current commit without exception, even if we're not mounting the volume.

ividal:
⚠️ Rebased on current main following @chainlink 's #295.

lumigator/python/mzai/backend/backend/tests/data/dialogsum_converted.csv
aittalam, line 1: Just a heads up that Javier is also adding a dialogsum dataset in [PR273](https://github.com/mozilla-ai/lumigator/pull/273). I think it's ok to have something close to the tests, perhaps just keeping the first few lines might reduce its fingerprint (TL/DR you can copy his few-lines CSV here if you want)

ividal: Yup, I saw, I think we should put that in a common space visible to sdk and be @javiermtorres Where do you think, `python/mzai/sample_data`?

.github/workflows/tests_uv.yaml
javiermtorres: AFAICT it should work ok as it is right now?

lumigator/python/mzai/backend/backend/tests/api/routes/test_api_workflows.py
```diff
@@ -0,0 +1,51 @@
...
+    assert response.status_code == 200
+
+    create_response = local_client.post("/datasets",
+            data={},
+            files={"dataset": dialog_dataset, "format": (None, DatasetFormat.EXPERIMENT.value)},
+        )
+
+    assert create_response.status_code == 201
+
+    created_dataset = DatasetResponse.model_validate(create_response.json())
```
javiermtorres, line 33: A shorter way could be:

```suggestion
    created_dataset = DatasetResponse(**(response.json()))
```

But I'm not sure of the differences between each method.

lumigator/python/mzai/backend/backend/tests/api/routes/test_api_workflows.py
```diff
@@ -0,0 +1,51 @@
...
+    headers = {
+        "accept": "application/json",
+        "Content-Type": "application/json",
+    }
+    payload = {
```
javiermtorres, line 38: Maybe take from a json file?

javiermtorres (approved):
Overall LGTM. IMHO we should be able to share the data CSV file.

lumigator/python/mzai/backend/backend/tests/data/dialogsum_converted.csv
aittalam: I think that's ok! It does not add a lot of extra data and also gives us the freedom to test on different data later on if we feel the need to do so

.github/workflows/integration_tests.yaml
```diff
@@ -0,0 +1,61 @@
...
+          # may give a useful partial result that will save time over completely clean state,
+          # but will cause the cache entry to grow without bound over time.
+          # See https://pants.readme.io/docs/using-pants-in-ci for tips on how to periodically clean it up.
+          # Alternatively you change gha-cache-key to ignore old caches.
+
+      - name: Set up Python
+        run: |
+          make bootstrap-python
+
+      - name: Lint
```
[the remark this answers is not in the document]
ividal: New link action covers backend (incl. tests) too: https://github.com/mozilla-ai/lumigator/pull/294/files

lumigator/python/mzai/backend/backend/tests/data/dialogsum_converted.csv
ividal: Agreed! In that case a tiny PR to move the data around will do :)

lumigator/python/mzai/backend/backend/tests/api/routes/test_api_workflows.py
ividal: For one... readability, which for this repo is a big thing. If you don't feel the second method adds robustness, I would keep the first one  (❓ )

lumigator/python/mzai/backend/backend/tests/api/routes/test_api_workflows.py
ividal: Seeing what the changes look like and having a not-real json with only partial values lying around, the balance efficiency to readability doesn't pan out.

.github/workflows/tests_uv.yaml
ividal: It is working as expected, yes. Let's keep an eye on it in case later we need to force a build later on.

aittalam (approved):

ividal:
Thank you all for your feedback 🙏

lumigator/python/mzai/backend/backend/tests/api/routes/test_api_workflows.py
```diff
@@ -0,0 +1,51 @@
...
+import pytest
+import requests
+from fastapi.testclient import TestClient
+from schemas.datasets import DatasetFormat, DatasetResponse
+
+
+# First test waits and polls up to 10 times for the real backend to be up.
+def test_health_ok(local_client: TestClient):
+    attempt = 0
+    while attempt < 10:
```
veekaybee, line 12:
```suggestion
    while attempt < range(10)
```
then we don't need to increment a counter

lumigator/python/mzai/backend/backend/tests/api/routes/test_api_workflows.py
```diff
@@ -0,0 +1,51 @@
+from time import sleep
+
+import pytest
+import requests
+from fastapi.testclient import TestClient
+from schemas.datasets import DatasetFormat, DatasetResponse
+
+
+# First test waits and polls up to 10 times for the real backend to be up.
+def test_health_ok(local_client: TestClient):
```
veekaybee, line 10: we can use FastAPI decorated idioms here as well, `on_startup`

https://fastapi.tiangolo.com/advanced/testing-events/

so we should be able to do something like,

```python
@app.on_event("startup")
def test_health_ok(local_client: TestClient):
        response = local_client.get("/health/")
```

veekaybee:
Sorry, missed that this was merged. But we should still be able to simplify these tests going forward

.github/workflows/tests_uv.yaml
chainlink: > It is working as expected, yes. Let's keep an eye on it in case later we need to force a build later on.
This is actually testing against a completely arbitrary version of lumigator due to https://mzlai.slack.com/archives/C07LJP9Q5KJ/p1729259395864809

lumigator/python/mzai/backend/backend/tests/api/routes/test_api_workflows.py
ividal: Right you are 😆 
I'll touch that next PR I open if someone doesn't fix that sooner.

lumigator/python/mzai/backend/backend/tests/api/routes/test_api_workflows.py
ividal: Oh, nice one, thanks!

Since the comment comes in after merging, I'm adding this to a small PR I'll open for a bugfix today/tomorrow.

veekaybee: I included this in https://github.com/mozilla-ai/lumigator/pull/285 so no need 👍

PR #291, opened 2024-10-17
Added support for Local environment for non contributors

agpituk:
## What's changing

- Simplified the Lumigator setup by using a pre-built external container, removing the need for users to set up a local Python environment.
- Added new Makefile entry points to streamline running and managing the app.
- Maintained support for the original local setup for backward compatibility.

## How to test it
Run both versions of lumigator, the old one with make-local-up and the new one with start-lumigator and test the app
## I already...

- [N/A] added some tests for any new functionality;


.devcontainer/docker-compose.override.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - PIP_REQS=/mzai/lumigator/python/mzai/evaluator/requirements.txt
+      - EVALUATOR_WORK_DIR=/mzai/lumigator/python/mzai/evaluator
+      - RAY_DASHBOARD_PORT=8265
+      - RAY_HEAD_NODE_HOST=ray
+      - MISTRAL_API_KEY=${MISTRAL_API_KEY}
+      - OPENAI_API_KEY=${OPENAI_API_KEY}
+      - RAY_WORKER_GPUS=0
+      - RAY_WORKER_GPUS_FRACTION=0
+    volumes:
+      - ./:/mzai
```
macaab26, line 36: If I'm not wrong, we are building with the Dockerfile located in the root of the repo, that one includes the whole mzai directory, so this volume is not needed. Am I right, or I'm missing something?

agpituk: Hi Mario! This volume is to mount the code in order to do local development. The docker-compose override is only used in the "local-up" (not start-lumigator), so this will be only used for local development.

docker-compose.yaml
```diff
@@ -64,17 +72,16 @@ services:
     #       deployments.
     extra_hosts:
       - "localhost:host-gateway"
+    profiles: 
+      - !external
```
aittalam, line 76: From the documentation it looks like we'll be relying on an external S3-compatible object store too. Should we also disable localstack with `!external`?

.devcontainer/docker-compose.override.yaml
macaab26: Yep, I already knew that, but I missed the "--reload" in the uv command, now everything makes sense 👍 Thank you!

macaab26 (approved):

.devcontainer/docker-compose.override.yaml
```diff
@@ -0,0 +1,43 @@
...
+  backend:
+    build:
+      context: .
+      dockerfile: "Dockerfile"
+    platform: linux/amd64
+    command: ["uv", "run", "uvicorn", "backend.main:app", "--host", "0.0.0.0", "--port", "8000",  "--reload"]
+    depends_on:
+      - localstack
+    ports:
+      - 80:8000
```
aittalam, line 14: What is the rationale behind mapping to port 80 for local development and keeping port 8000 otherwise?
If we want to keep them different (my 2 cents is we could use the same port btw), it might be helpful to make it explicit in our documentation

README.md
```diff
@@ -88,6 +88,27 @@ You can build the local project `docker-compose` on Mac or Linux,  or into a dis
     + On Linux, please also follow the [post-installation steps](https://docs.docker.com/engine/install/linux-postinstall/).
 + System Python (that is: no version manager, such as pyenv, should be active).
 
+# Running Lumigator
+
+## Running Lumigator Locally
+
+1. `git clone git@github.com:mozilla-ai/lumigator.git`
+2. `make start-lumigator`
+3. The application should be available at http://localhost:8000.
```
aittalam, line 97: Perhaps "The REST API should be available..."? To me application looks a bit more advanced than our API UI, but it can be subjective. WDYT?

.devcontainer/docker-compose.override.yaml
agpituk: It's actually a great point Davide. I think it's a good moment to change this! I didn't want to change 80 for our dev environment because I know some scripts we use for testing (like the ones you shared in slack) point directly to localhost, but it would be a great moment to choose a port that works for us and use it everywhere! Should I use 8000 everywhere? or something more random defined by us like 5151?

docker-compose.yaml
agpituk: It definitely makes sense to add !external to this. The only caveat is that I would need to remove the depends_on in the backend section and I'm not entirely happy with it. I've tested it and it works, but it might be because I have things cached locally, if I do a fresh start it might fail without the depends_on. If the users want to launch lumigator with external ray and I don't add the !profile as you suggest, the only issue right now is that the user will have an extra container locally. I think a possible solution it's to divide the compose into different files and load them on demand with the Makefile. Would you be happy to be addressed as a refactor in a different ticket?

aittalam: Thanks for explaining! I think it's ok, let us just track this via an issue and merge the current code

.devcontainer/docker-compose.override.yaml
aittalam: Thanks! I think 8000 is fine, as long as it is well documented so ppl know where to point their code and where to change this value in case they have something else running on the same port

aittalam (approved):
A couple of notes but overall LGTM, thank you!

.devcontainer/docker-compose.override.yaml
agpituk: Updated Readme and consolidated port 8000 for local & Dev

Issue #225, opened 2024-09-23
Re-enable disabled ruff rules


aittalam:
**Is your feature request related to a problem? Please describe.**
In PR209 we disabled a few ruff rules to more easily integrate lm-buddy.
It would be useful to go through the [diff](https://github.com/mozilla-ai/lumigator/pull/209/files#diff-26f436cb29eecc7bb4f9066747b43df8da81ac808955ce95c109f5d7aa778989), decide which ones we want to keep, and update the code to be compliant with them.

**Describe the solution you'd like**
If we agree that some rules are good to keep, I'd like to make sure that all the code (not just lumigator's, but also lm-buddy's) is compliant.

**Describe alternatives you've considered**
The alternative is not to re-enable any of the rules. Which is fine, as long as we agree we do not need them - I guess the main point is converging on some rules we are happy to keep across all our projects.

PR #293, opened 2024-10-17
Ruff fixes

macaab26:
## What's changing

This PR is an attempt to "clean" the repo from the errors/suggestion that ruff creates prior to configure it in the action (that PR will come later!)

Ruff fixes!

## I already...

### DIFF
--- a/notebooks/lumigator_demo.py
+++ b/notebooks/lumigator_demo.py
@@ -1,13 +1,13 @@
 """Common definitions and methods for the lumigator demo notebook."""
 
 import io
+import json
 import os
 from pathlib import Path
 from typing import Any, Dict  # noqa: UP035
 from uuid import UUID
-import json
-import numpy as np
 
+import numpy as np
 import pandas as pd
 import requests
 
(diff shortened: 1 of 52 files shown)


notebooks/lumigator_demo.py
```diff
@@ -241,7 +241,7 @@ def parse_model_results(results):
 
             row[column] = value
 
-        row["RAM_GB"] = mi[mi.model_name == model_name]['RAM_GB'].values[0]
+        row["RAM_GB"] = mi[mi.model_name == model_name]['RAM_GB'].to_numpy()[0]
```
veekaybee, line 244: we'll want to check this one just in case

veekaybee:
LGTM pending the numpy check, I also checked the rules that we disabled/re-enabled for the lm-buddy import https://github.com/mozilla-ai/lumigator/issues/225

and those look ok as well, cc @aittalam

veekaybee (approved):
Pre-approving once we double-check a few items

notebooks/lumigator_demo.py
aittalam: Thanks @veekaybee ! Yes, it makes sense and I tested it, I confirm it works as expected 👍

aittalam (approved):
Many thanks @macaab26 , LGTM!

PR #298, opened 2024-10-21
[ LUMI-184] Build lumigator before tests

chainlink:
## What's changing
This runs both integration test suites against the _current_ code. Not the `latest` tag (which currently represents the `latest` code built, and as such could literally be any branch)

## How to test it
Run pytest in backend and sdk projects

### DIFF
--- a/.github/workflows/sdk_integration_tests.yaml
+++ b/.github/workflows/sdk_integration_tests.yaml
@@ -9,7 +9,6 @@ on:
       - "**"
   workflow_dispatch:
 
-
 jobs:
   integration:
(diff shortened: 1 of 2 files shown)


javiermtorres (approved):
For extra security, _maybe_ we could think about using the SHA in the tag instead of latest. But I guess that would be part of a separate PR. LGTM right now.

ividal:
Thanks a **lot** @chainlink 🙏 I very much like that now this is doing what make local-up was before.

ividal (approved):

PR #304, opened 2024-10-22
Separate dev dependencies and add pre-commit support

macaab26:
## What's changing

I realized that we were introducing dev dependencies to the image, like pytest. At the same time, I also discover that we weren't installing anywhere the pre-commit dependency, and we didn't configure it. 

So, I fix it. I modified the compose and the Dockerfile and the compose, and I created a new uv/python project in the root of the repo only including the pre-commit. When we execute `make local-up`, it also configures the pre-commit.

As long as the pre-commit is configured at a repo level inside .git/webhooks/, after execute the `make local-up` you can go to whatever folder and follow the development instructions `uv sync`, etc. 

## How to test it

Clean the .venv folders if you already have one (or more), just to make sure, and execute `make local-up`, go to the backend folder, run `uv sync,` and start developing! 

## I already...


agpituk (approved):

.devcontainer/docker-compose.override.yaml
```diff
@@ -3,6 +3,9 @@ name: lumigator
 services:
 
   backend:
-    command: ["uv", "run", "uvicorn", "backend.main:app", "--host", "0.0.0.0", "--port", "8000",  "--reload"]
+    build:
```
veekaybee, line 6: just double-checking, this is for the shipped image, not local dev  right? want to make sure we can still hot-reload the app locally

macaab26: Hey @veekaybee, yep, the override is for local development. I have to reestructure the Dockerfile, and with the new changes there's no need to have it here. Now it's located [here](https://github.com/mozilla-ai/lumigator/blob/f61e79ee1b603527f55d241b293af2866c740fa6/Dockerfile#L21)

chainlink:
@macaab26 can you hold before merging this, I think it needs further discussion.

.github/workflows/build_images_uv.yaml
```diff
@@ -51,6 +51,7 @@ jobs:
         with:
           platforms: linux/amd64,linux/arm64
           push: true
+          target: "main_image"
           tags: |
             mzdotai/lumigator:backend_dev_${{ env.REPOVERSION }}
```
javiermtorres, line 56: Maybe factoring `mzdotai/lumigator` as a variable? So that things are kept consistent if the registry changes.

.github/workflows/build_images_uv.yaml
```diff
@@ -51,6 +51,7 @@ jobs:
         with:
           platforms: linux/amd64,linux/arm64
           push: true
+          target: "main_image"
           tags: |
```
javiermtorres, line 55: Instead of setting the tags manually, it could be worth checking out https://github.com/docker/metadata-action
It should extract and format most tags like semver more easily.

javiermtorres (approved):

peteski22 (approved):

.github/workflows/build_images_uv.yaml
macaab26: Hey @javiermtorres yep, it's a good idea, but I will introduce it in another PR with the semver that you mentioned in your other comment 👍

.github/workflows/build_images_uv.yaml
macaab26: Agree @javiermtorres let's decide how we want to manage the semver, and let's include it in another PR

.github/workflows/build_images_uv.yaml
javiermtorres: Ok, np!!!!

.github/workflows/build_images_uv.yaml
javiermtorres: Ok 👍

PR #307, opened 2024-10-23
Backend: Introduce `alembic` to assist with database migrations

peteski22:
## What's changing

This PR introduces `alembic` as a tool which should be used to help with database migrations (e.g. schema changes).

The Github PR template also has an additional checklist item to make sure engineers have considered whether a migration version is required or not.

Some of the `ruff` checks (`uvx ruff check --fix --config ruff.toml`) weren't happy about the generated output for the initial version/migration, so I've added `# pragma: allowlist secret` to them, and the template that generates them to stop it happening each time.

**NOTE**: 

There's a [slightly shonky](https://github.com/mozilla-ai/lumigator/pull/307/files#diff-5c0e0458001b672561120aa24b1ada87a1a568f0ee454a6ddf040ec6911c276bR4-R16) thing going on at the top of `.alembic/env.py`, where we have to import all the types that inherit from `BaseRecord` in order to ensure their metadata is represented (and they're included in the changes that `alembic` sees).

## How to test it

With a non-existent local DB:

```bash
export SQLALCHEMY_DATABASE_URL=sqlite:///local.db # export env var to point to local DB

cd lumigator/python/mzai/backend
uv sync && uv sync --dev && source .venv/bin/activate

alembic upgrade head # should run upgrade and create DB tables

alembic current # should report e75fa022c781 (head)
```

As future changes are made to the data models, they should be captured (see `.alembic/README.md`) by our engineers (ideally using `--autogenerate` when creating an `alembic revision`). This information is captured in the READMEs as well as a comment in the `env.py` file.

## I already...

- [x] Manually tested the change.
- [x] Updated the documentation.


lumigator/python/mzai/backend/pyproject.toml
```diff
@@ -28,7 +28,9 @@ dependencies = [
 
 [tool.uv]
 dev-dependencies = [
+    "alembic>=1.13.3",
     "pytest>=8.3.3",
+    "ruff>=0.7.0",
```
peteski22, line 33: Not 100% sure I need this one, open to suggestion/advice (and can remove and sort out the lock file too).

lumigator/python/mzai/backend/.alembic/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+```bash
+SQLALCHEMY_DATABASE_URL=sqlite:///local.db alembic history
+```
+
+Examples shown in this document will presume `SQLALCHEMY_DATABASE_URL` is exported.
+
+If `SQLALCHEMY_DATABASE_URL` is not present then a default of `sqlite:///local.db` will be used (see: `alembic.ini`)
+
+## Viewing migration history
```
veekaybee, line 39: Non-blocking: might want to mention that alembic will be available in your local environment via virtualenv or you should be able to run with `uv run` directly?

lumigator/python/mzai/backend/.alembic/env.py
```diff
@@ -0,0 +1,89 @@
...
+If a new data model package is added, then it MUST be manually imported here!
+"""
+
+
+# this is the Alembic Config object, which provides
+# access to the values within the .ini file in use.
+config = context.config
+
+# Interpret the config file for Python logging.
+# This line sets up loggers basically.
```
veekaybee, line 24: is alembic logging separate from the app-level logging we do via logguru? would we want it to be the same?

peteski22: The Alembic 'stuff' is totally separate from the app running, so I think it's probably fine? 🤔 A lot of the contents of that file is scaffolding from the `alembic init` command. In the `env.py` the main 'changes' are the imports at the top (with the note about them having to be there), and:

```
target_metadata = BaseRecord.metadata

# Override the SQLAlchemy URL with the one we have stored in our environment.
config.set_main_option(
    'sqlalchemy.url',
    os.environ.get("SQLALCHEMY_DATABASE_URL", config.get_main_option("sqlalchemy.url")))
```

it's a manual step you'd run if you made schema changes to the data models, to generate the migration version, so any logging would be something you'd probably want to see in your terminal window.

lumigator/python/mzai/backend/.alembic/README.md
peteski22: @veekaybee does this cover the ways to run things sufficiently?

https://github.com/mozilla-ai/lumigator/pull/307/commits/acf0bd703cd8b5d3b247b0a22cbac8ef202aed56

veekaybee: works for me!

lumigator/python/mzai/backend/.alembic/env.py
veekaybee: Awesome, thanks for the explanation!

veekaybee (approved):
High-level code looks good but will defer to folks who have worked with Alembic before for additional comments, thanks for adding!

chainlink:
Sorry if I missed this in the PR, but is there a way to initialize the db via docker compose?

peteski22:
> Sorry if I missed this in the PR, but is there a way to initialize the db via docker compose? (So the db is automatically set up when you run make local-up, etc)

I think this might be what you mean: https://github.com/mozilla-ai/lumigator/pull/307/commits/a7739dc417e7877ed9d61aa4043d89897bf71675#diff-8906ffee58947d988f8df66765cbd440cc8445f04f4c1930b7cc6708c540ba83 🤞🏼

javiermtorres:
Is there any short doc listing the usual procedures and how to perform them? E.g. initializing the db with the first "migration" (already described in the comment above), how to perform a live db migration, etc? I guess there won't be so many such procedures (four, five)?

javiermtorres (approved):
LGTM - Maybe some howto doc can be added (now or in a separate PR).

javiermtorres:
@peteski22 sorry for missing https://github.com/mozilla-ai/lumigator/pull/307/files#diff-1295ef5849cbe80ac81064e53e59517be8f8198b2a5a5069e59c6363aaa83dbcR65
Yes, I think the information I was missing is there. Sorry again!!!

PR #308, opened 2024-10-23
Linting: `uvx ruff check --fix --config ruff.toml`

peteski22:
## What's changing

`uvx ruff check --fix --config ruff.toml` found these required fixes/changes.


veekaybee (approved):
Thanks for this!

PR #309, opened 2024-10-23
Batch inference round 1/n

aittalam:
## What's changing

- evaluator has been moved from `mzai/evaluator` to `mzai/jobs/evaluator`
- a new inference job has been added into `mzai/jobs/inference`
- minor fixes: broken evaluator test, minor python version for backend

## How to test it

- backend, sdk: `uv run pytest` (strangely SDK still appears to be running ok with the mock `/experiments` endpoint?)
- evaluator: `uv venv --python 3.11.9 && uv pip install pytest && uv pip install -r requirements.txt` to set the environment, then `uv run pytest`
- inference: no tests yet

This is the first of a few  PRs to split the larger task into smaller ones:
- move evaluator, wire an inference job into lumigator [this PR]
- update the inference job configuration (as right now it takes everything that is provided to evaluation and just ignores unuseful params e.g. metrics) and the respective API
- improve the inference job code (using vLLM-as-a-library and relying on ray.data for further parallelization)
- fix outputs so an updated CSV file is returned instead of a JSON (this requires a discussion to choose whether/where we want to store further metadata)
- unit tests

## I already...


lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -14,14 +14,15 @@
 router = APIRouter()
 
 
-@router.post("/inference", status_code=status.HTTP_201_CREATED)
+@router.post("/inference/", status_code=status.HTTP_201_CREATED)
```
veekaybee, line 17: thanks!

lumigator/python/mzai/backend/backend/settings.py
```diff
@@ -42,8 +42,23 @@ class BackendSettings(BaseSettings):
...
     DEFAULT_SUMMARIZER_PROMPT: str = "You are a helpful assistant, expert in text summarization. For every prompt you receive, provide a summary of its contents in at most two sentences."  # noqa: E501
 
-    # Eval
+    # Eval job details
     EVALUATOR_WORK_DIR: str | None = None
+    EVALUATOR_PIP_REQS: str | None = None
+
+    @computed_field
+    @property
+    def EVALUATOR_COMMAND(self) -> str:  # noqa: N802
```
veekaybee, line 51: non-blocking: add a note about why LD_PRELOAD_PREFIX here

lumigator/python/mzai/jobs/inference/inference.py
```diff
@@ -0,0 +1,161 @@
...
+from box import Box
+from datasets import load_from_disk
+from loguru import logger
+from model_clients import (
+    BaseModelClient,
+    MistralModelClient,
+    OpenAIModelClient,
+)
+
+# SummarizationPipelineModelClient,
```
veekaybee, line 21: can we remove?

lumigator/python/mzai/jobs/inference/inference.py
```diff
@@ -0,0 +1,161 @@
...
+    for sample_txt in dataset_iterable:
+        predictions.append(model_client.predict(sample_txt))
+
+    return predictions
+
+
+def save_outputs(config: Box, inference_results: dict) -> Path:
+    storage_path = config.evaluation.storage_path
+
+    def save_to_disk(local_path: Path):
```
veekaybee, line 57: do we want a nested method call here?

lumigator/python/mzai/jobs/inference/inference.py
```diff
@@ -0,0 +1,161 @@
...
+
+        return local_path
+
+    except Exception as e:
+        logger.error(e)
+
+
+def run_inference(config: Box) -> Path:
+    # Do something with your config here, for example:
+    print("Received configuration:")
```
veekaybee, line 94: switch to logging here?

lumigator/python/mzai/jobs/inference/inference.py
```diff
@@ -0,0 +1,161 @@
...
+        output_model_name = config.model.inference.engine
+        if "mistral" in base_url:
+            # run the mistral client
+            logger.info(f"Using Mistral client. Endpoint: {base_url}")
+            model_client = MistralModelClient(base_url, config.model)
+        else:
+            # run the openai client
+            logger.info(f"Using OAI client. Endpoint: {base_url}")
+            model_client = OpenAIModelClient(base_url, config.model)
+    # else:
```
veekaybee, line 125: do we still need this branch?

lumigator/python/mzai/backend/backend/tests/api/routes/test_api_workflows.py
```diff
@@ -7,19 +7,22 @@
...
 def test_upload_data_launch_job(local_client: TestClient, dialog_dataset):
     response = local_client.get("/health")
     assert response.status_code == 200
 
-    create_response = local_client.post("/datasets",
-            data={},
-            files={"dataset": dialog_dataset, "format": (None, DatasetFormat.EXPERIMENT.value)},
-        )
+    create_response = local_client.post(
+        "/datasets",
```
veekaybee, line 22: should we change to `/datasets/' re slack convo?

lumigator/python/mzai/jobs/inference/inference.py
```diff
@@ -0,0 +1,161 @@
...
+    MistralModelClient,
+    OpenAIModelClient,
+)
+
+# SummarizationPipelineModelClient,
+# HuggingFaceModelClient,
+from tqdm import tqdm
+
+
+def timer(func):
```
veekaybee, line 26: are we still thinking we want this given job logging in Ray?

lumigator/python/mzai/jobs/inference/inference.py
```diff
@@ -0,0 +1,161 @@
...
+            save_to_s3(local_path, storage_path)
+            return storage_path
+
+        return local_path
+
+    except Exception as e:
+        logger.error(e)
+
+
+def run_inference(config: Box) -> Path:
```
veekaybee, line 92: Would using dictionaries or Pydantic model field access be sufficient here? I'm wary of adding levels of complexity if we don't need them.

lumigator/python/mzai/jobs/inference/inference.py
```diff
@@ -0,0 +1,161 @@
...
+    #     # for inference
+    #     if config.evaluation.use_pipeline:
+    #         logger.info(f"Using summarization pipeline. Model: {model_name}")
+    #         model_client = SummarizationPipelineModelClient(model_name, config)
+    #     else:
+    #         logger.info(f"Using direct HF model invocation. Model: {model_name}")
+    #         model_client = HuggingFaceModelClient(model_name, config)
+
+    # run inference
+    output = {}
```
veekaybee, line 139: suggestion: put `output` at top of method so it's clearer to read "inputs" and return

lumigator/python/mzai/jobs/inference/inference.py
```diff
@@ -0,0 +1,161 @@
...
+    output["predictions"], output["inference_time"] = predict(dataset_iterable, model_client)
+    output["examples"] = dataset["examples"]
+    output["ground_truth"] = dataset["ground_truth"]
+    output["model"] = output_model_name
+
+    output_path = save_outputs(config, output)
+    return output_path
+
+
+if __name__ == "__main__":
```
veekaybee, line 149: given that we use click elsewhere in lumigator, it makes sense to standardize on that as CLI interface https://github.com/mozilla-ai/lumigator/blob/543a3a10d629296d1b83ad76a9958175235d1638/lumigator/python/mzai/evaluator/evaluator/__main__.py#L9

lumigator/python/mzai/jobs/inference/model_clients.py
```diff
@@ -0,0 +1,198 @@
...
+        """Used to initialize the model / inference service."""
+        pass
+
+    @abstractmethod
+    def predict(self, prompt: str) -> str:
+        """Given a prompt, return a prediction."""
+        pass
+
+
+# class SummarizationPipelineModelClient(BaseModelClient):
```
veekaybee, line 40: Can we remove?

lumigator/python/mzai/jobs/inference/inference.py
aittalam: Yes! I was planning to also have transformers code originally but if we go the vLLM way we won't need this

lumigator/python/mzai/jobs/inference/inference.py
aittalam: Removed - that was just debugging code I would have taken away at the end anyway

lumigator/python/mzai/jobs/inference/inference.py
aittalam: we'll need to branch if we are not calling a service, but I'll update the code to use the proper model client (vLLM likely)

lumigator/python/mzai/backend/backend/tests/api/routes/test_api_workflows.py
aittalam: +1 to that. I think the way `local_client` handles it changes nothing (as it gets a 307 redirect that it follows automatically) but given a simple curl would break without the ending slash it makes sense to be consistent here and avoid redirects

lumigator/python/mzai/jobs/inference/inference.py
aittalam: Not sure about that: it is convenient now that the code is still in development but I think it was more useful in the eval script (that returned a json including these metrics) rather than here (where we'll likely just return a CSV file with one additional column)

lumigator/python/mzai/jobs/inference/inference.py
```diff
@@ -0,0 +1,161 @@
...
+        if storage_path.endswith("/"):
+            storage_path = "s3://" + str(
+                Path(storage_path[5:]) / config.name / "inference_results.json"
+            )
+        logger.info(f"Storing into {storage_path}...")
+        s3.put_file(local_path, storage_path)
+
+    # generate local temp file ANYWAY
+    # (we don't want to lose all eval data if there is an issue wth s3)
+    local_path = Path(
```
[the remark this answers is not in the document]
aittalam: I think this is mostly an artifact from the old lm-buddy use-case: we used temp files in the past but in this specific case I wanted to have an actual file which is deleted after the upload, so that if anything breaks on the S3 side the file is at least available locally.
However, this is run by a worker now and even if the file is stored locally it will be deleted when the job ends... So on the one hand we don't have to worry about that, on the other one though we'd like to make sure the file remains available for the above reasons?

lumigator/python/mzai/jobs/inference/inference.py
aittalam: I agree on avoiding extra complexity, here I chose Box to reduce complexity at the expense of one extra dependency. The experience is similar to our previous pydantic configs: attribute-based access on nested values with a definable default, so we can try directly accessing a value and the worst that can happen is we get a None.

Ideally we'd have pydantic configs everywhere, but sharing them between jobs and lumigator definitely adds complexity and we are discussing the possibility of having them in lumigator only.

A simpler alternative would be to write some code to convert a nested dictionary using e.g. `types.SimpleNamespace`. Similar result with no dependencies, but without the default values for missing keys.

My 2 cents would be to continue using Box until we find a good solution for pydantic. WDYT?

lumigator/python/mzai/jobs/inference/inference.py
aittalam: As these scripts are not going to be called via CLI as lm-buddy did, I'd argue we should provide a lighter interface reducing the overall complexity and dependencies. But I am happy to take this discussion on slack or in a meeting and converge on what'd be the best solution going forward!

lumigator/python/mzai/jobs/inference/inference.py
veekaybee: Per online discussion, let's take out box once we've aligned on a set of fields we'd like to access

lumigator/python/mzai/jobs/inference/inference.py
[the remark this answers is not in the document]
veekaybee: Yup, the file will still be available, mkdtemp just offers a cleaner way to get to it

lumigator/python/mzai/jobs/inference/inference.py
veekaybee: in that case, let's remove?

aittalam: Done!

lumigator/python/mzai/jobs/inference/inference.py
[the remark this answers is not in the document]
aittalam: Cool, thanks! I'll add this into the next PR then!

lumigator/python/mzai/jobs/inference/model_clients.py
```diff
@@ -0,0 +1,143 @@
...
+class MistralModelClient(APIModelClient):
+    """Model client for models served via Mistral API.
+    - The base_url is fixed
+    - Choose an engine name (see https://docs.mistral.ai/getting-started/models/)
+    - Customize the system prompt if needed
+    """
+
+    def __init__(self, base_url: str, config: Box):
+        super().__init__(config)
+        self._client = MistralClient(api_key=os.environ["MISTRAL_API_KEY"])
```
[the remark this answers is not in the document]
aittalam: You mean settings.py? That is where it is coming from (local env -> docker compose -> settings -> ray.submit env -> local env for the worker)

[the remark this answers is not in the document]
veekaybee: totally forgot this was on Ray 🤦‍♀️

lumigator/python/mzai/jobs/inference/inference.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def save_outputs(config: Box, inference_results: dict) -> Path:
+    storage_path = config.evaluation.storage_path
+
+    # generate local temp file ANYWAY
+    # (we don't want to lose all eval data if there is an issue wth s3)
+    local_path = Path(
+        Path.home() / ".lumigator" / "results" / config.name / "inference_results.json"
+    )
```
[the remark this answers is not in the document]
aittalam: In this case, it is automatically deleted when the job ends. But I agree that making this explicit would be better. I'll follow Vicki's temp file suggestion!

lumigator/python/mzai/jobs/inference/inference.py
```diff
@@ -0,0 +1,124 @@
...
+            save_to_s3(local_path, storage_path)
+            return storage_path
+
+        return local_path
+
+    except Exception as e:
+        logger.error(e)
+
+
+def run_inference(config: Box) -> Path:
```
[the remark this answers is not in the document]
aittalam: I agree! That's very close to what we had in evaluator and then dropped it in lumigator, but I think I'd rather have the opposite (make sure the data is sound before it is sent to workers) if I had to choose, and ideally both (using some shared pydantic configs library)

lumigator/python/mzai/jobs/inference/inference.py
```diff
@@ -0,0 +1,124 @@
...
+    # initialize output dictionary
+    output = {}
+
+    # Load dataset given its URI
+    dataset = load_from_disk(config.dataset.path)
+
+    # Limit dataset length if max_samples is specified
+    max_samples = config.evaluation.max_samples
+    if max_samples and max_samples > 0:
```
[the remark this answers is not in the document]
aittalam: I agree on `if max_samples is not None`, happy to change this now in the code!

PR #317, opened 2024-10-25
API: `health/jobs` related endpoints - correctly handle 404s etc

peteski22:
## What's changing

`200` OK and an empty JSON object were returned when querying the current health job metadata endpoint under  `health/jobs/{job_id}`. 

This PR improves the way the response from upstream is handled, now returning a `404` NOT FOUND when it's not found.

**TODO**: Add unit tests with mocks

## How to test it

Where ID `3fa85f64-5717-4562-b3fc-2c963f66afa6` does not exist:

```
curl -v http://localhost:8000/api/v1/health/jobs/3fa85f64-5717-4562-b3fc-2c963f66afa6 \
```

Returns:

```
HTTP/1.1 404 Not Found
```

and JSON response body:

```json
{
  "detail": "Job metadata for ID: 3fa85f64-5717-4562-b3fc-2c963f66afa6 not found"
}
```

## I already...

### DIFF
--- a/lumigator/python/mzai/backend/backend/api/routes/health.py
+++ b/lumigator/python/mzai/backend/backend/api/routes/health.py
@@ -1,8 +1,10 @@
 import json
+from http import HTTPStatus
 from uuid import UUID
 
+import loguru
 import requests
-from fastapi import APIRouter
+from fastapi import APIRouter, HTTPException
 from schemas.extras import HealthResponse
 from schemas.jobs import JobSubmissionResponse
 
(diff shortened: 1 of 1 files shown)


javiermtorres:
To be pedantic, the errors are actually different (null object vs empty list). I think I prefer to have an empty list, so the end user knows at least they're reaching the right endpoint. But let's see what other people think.

peteski22:
> To be pedantic, the errors are actually different (null object vs empty list). I think I prefer to have an empty list, so the end user knows at least they're reaching the right endpoint. But let's see what other people think.

When you ask the API for the job metadata for a specific job via `job_id` you might get a 404 which it makes sense to error out on the same way.

 https://github.com/mozilla-ai/lumigator/pull/317/files#diff-bc21eb24ef7e8d2c97660608a2ca09d7f46e7f6ffd2f749c4c75e03ad8bb20e8R21-R22

When you get 'all' jobs if the upstream service replies with a 200 - OK. then you will end up with an empty list, as long as there's no problem parsing the JSON response (same as before)

https://github.com/mozilla-ai/lumigator/pull/317/files#diff-bc21eb24ef7e8d2c97660608a2ca09d7f46e7f6ffd2f749c4c75e03ad8bb20e8R49-R50

PR #318, opened 2024-10-25
API: `health/jobs` related endpoints - correctly handle 404s etc

peteski22:
## What's changing

`200` OK and a JSON object with default values were returned when querying the current health job metadata endpoint under  `health/jobs/{job_id}`. 

This PR improves the way the response from upstream is handled, now returning a `404` NOT FOUND when it's not found.

The PR also includes a change to `pre-commit` to exclude json files within `tests/data/` folders, so we can have sample data for test comparison. 

When services are injected consistently, the mocking in these tests can be changed and the dev dependency removed.

## How to test it

Run `make local-up`

Where ID `00000000-0000-0000-0000-000000000000` does not exist:

```
curl -v http://localhost:8000/api/v1/health/jobs/00000000-0000-0000-0000-000000000000 \
```

Returns:

```
HTTP/1.1 404 Not Found
```

and JSON response body:

```json
{
  "detail": "Job metadata for ID: 3fa85f64-5717-4562-b3fc-2c963f66afa6 not found"
}
```

## I already...


peteski22:
Recreated using branch with correct naming (from https://github.com/mozilla-ai/lumigator/pull/317)

lumigator/python/mzai/backend/backend/tests/conftest.py
```diff
@@ -6,6 +7,7 @@
 from fastapi import FastAPI
 from fastapi.testclient import TestClient
 from mypy_boto3_s3 import S3Client
+from requests_mock import Mocker
```
javiermtorres, line 10: Maybe `import requests_mock` directly to use Mocker in a qualified way, so we know what kind of Mock we're using exactly?

javiermtorres (approved):
LGTM. @peteski22 did a live demo of the current vs previous return result, so I consider this to be tested ok.

lumigator/python/mzai/sdk/tests/test_jobs.py
```diff
@@ -64,26 +64,26 @@ def test_get_jobs_none(mock_requests_response, mock_requests, lumi_client):
...
-def _test_get_job_id_does_not_exist(mock_requests_response, mock_requests, lumi_client):
+def test_get_job_id_does_not_exist(mock_requests_response, mock_requests, lumi_client):
     mock_requests_response.status_code = HTTPStatus.NOT_FOUND
     mock_requests_response.json = lambda: None
     error = HTTPError(response=mock_requests_response)
     mock_requests.side_effect = error
 
     # We expect the SDK to handle the 404 and return None.
-    job = lumi_client.health.get_job("6f6487ac-7170-4a11-af7a-0f6db1ec9a74")
+    job = lumi_client.health.get_job("5f0487ac-7170-4a11-af7a-0f6db1ec9a74")
```
javiermtorres, line 88:
```suggestion
    job = lumi_client.health.get_job("00000000-0000-0000-0000-000000000000")
```
This seems to be a special UUID instance as per https://datatracker.ietf.org/doc/html/rfc4122#section-4.1.7 and I would assume that most systems will not assign it as a valid UUID.

peteski22:
> Do we need to update directly or indirectly the openapi spec?

Currently we don't explicitly declare 404s in other endpoints, so I'd leave that out of this PR, but we should look at how we create the spec for sure! 👍🏼

PR #320, opened 2024-10-28
Backend: Tests use Alembic

peteski22:
## What's changing

The recent [change](https://github.com/mozilla-ai/lumigator/pull/307) to support `alembic` came with the removal of the DB engine which impacts some of our tests. This PR allows the tests now use alembic.

The PR also moves the `.alembic` folder containing migrations a level down under `backend/backend` and renamed to just `alembic` dropping the `.` prefix.

## I already...


macaab26 (approved):

lumigator/python/mzai/backend/backend/main.py
```diff
@@ -19,9 +19,12 @@
 
 
 def _init_db():
-    alembic_cfg = Config(str(Path().parent / "alembic.ini"))
+    logger.info("Initializing database via Alembic")
+    backend_root = Path(Path(Path(__file__).resolve()).parent).parent
```
chainlink, line 23: I think there's a better way to do this pathing cc @veekaybee

veekaybee: If we import alembic directly, it should read the config and resolve directly relative to the path based on the alembic config, but I might be some details.

```
from alembic.config import Config
def run_migrations():
    alembic_cfg = Config("alembic.ini")
    command.upgrade(alembic_cfg, "head")
```

Another way that's specific fastapi is to mount static files, in this case alembic.ini as a path that the app then recognizes. 

https://fastapi.tiangolo.com/tutorial/static-files/#use-staticfiles

veekaybee: Finally, if we're looking to do the same thing but break this up more cleanly, we can do 

```python
# /...../lumigator/lumigator/python/mzai/backend 
BASE_DIR = Path(__file__).resolve().parent 

#  .../...lumigator/lumigator/python/mzai/backend/alembic.ini
ALEMBIC_PATH = BASE_DIR  /  "alembic.ini" 

alembic_cfg = Config(str(ALEMBIC_PATH))
```

veekaybee: Not a blocker for me but I'd have to do more research on what idiomatic ways to trigger migrations at startup are.  (just did a quick search for some sample repos using Alembic + FastAPI and I'm not seeing the `upgrade` initiated at at the app entrypoint , but you're much further down this path than me.)

Another option here would be to pass the config path in from app settings where we already pass some filenames. https://github.com/mozilla-ai/lumigator/blob/main/lumigator/python/mzai/backend/backend/settings.py, what do you think?

chainlink: This might be something to look into to https://pytest-alembic.readthedocs.io/en/latest/setup.html Although perhaps after this PR

Issue #228, opened 2024-09-24
Add support for llamafile model names


veekaybee:
Currently llamafiles are supported as generic OAI-compatible servers. While this works, it forces us to specify a valid (i.e. hf-transformers loadable) model name - which is needed for vLLM but might not be the case here as one generally uses GGUF files or llamafile filenames. While not strictly required, these names might be useful to e.g. properly identify the model when reading evaluation results.

Supporting this boils down to adding a llamafile:// prefix to handle llamafiles differently and changing the validation methodology for the provided model names.

kernel-loophole:
hi @veekaybee wanted to work on this .where i can start looking for it .and how the llamafile is gonna identify the model name .

aittalam:
Hi @kernel-loophole, and thanks for your interest!

Our evaluator accepts input configurations like [this one](https://github.com/mozilla-ai/lumigator/blob/main/lumigator/python/mzai/evaluator/examples/configs/evaluation/hf_evaluate_inference_server_config.yaml) for inference servers which are compatible with OAI's API. As you can see, it accepts a `base_url` and an `engine` parameters: the former specifies where to access your llamafile's API, the latter is usually a huggingface-format model name (e.g. `mistralai/Mistral-7B-Instruct-v0.2`). We have not agreed on how to identify a llamafile (e.g. whether it is by the GGUF name or the llamafile's name itself), and AFAIK llamafile server [does not need you to specify any engine](https://github.com/Mozilla-Ocho/llamafile/blob/main/llama.cpp/server/api_like_OAI.py), so I think ATM any string that's meaningful for the user might work.

The main problem ATM is that for this kind of configurations we expect the model to be a valid one, so we need to apply the following changes to support llamafiles: 

1. in evaluator's paths, add support for the llamafile prefix [here](https://github.com/mozilla-ai/lumigator/blob/main/lumigator/python/mzai/evaluator/evaluator/paths.py#L18) and add some validation code for the [AssetPath](https://github.com/mozilla-ai/lumigator/blob/main/lumigator/python/mzai/evaluator/evaluator/paths.py#L41). There is actually no specific rule for validating a llamafile atm so I think a pass would be ok as a first iteration.
2. explicitly add llamafile models in [config_templates](https://github.com/mozilla-ai/lumigator/blob/main/lumigator/python/mzai/backend/backend/config_templates.py) so the name will be mapped to a specific configuration type

[...]

PR #321, opened 2024-10-29
Added support for llamafile model names

kernel-loophole:
### Add Utility Function for Asset Path Formatting and Validation

This PR utility function to format and validate paths for various asset storage platforms.
Key Changes

- added llamamodel suuport
- Defined PathPrefix Enum to support prefixes.
- Implemented strip_path_prefix and validate_asset_path for validation and formatting.
- Added helper functions to format paths for each storage provider.

#228


aittalam:
> ### Add Utility Function for Asset Path Formatting and Validation
> 
> This PR utility function to format and validate paths for various asset storage platforms. Key Changes
> 
>     * added llamamodel suuport
> 
>     * Defined PathPrefix Enum to support prefixes.
> 
>     * Implemented strip_path_prefix and validate_asset_path for validation and formatting.
> 
>     * Added helper functions to format paths for each storage provider.
> 
> 
> #228

Thank you @kernel-loophole! I will be going through your PR today 😄

kernel-loophole:
@aittalam thanks for review ,let me know if anything is missing

lumigator/python/mzai/jobs/evaluator/evaluator/paths.py
```diff
@@ -16,50 +15,74 @@ class PathPrefix(str, Enum):
...
         validate_repo_id(s)
         return True
     except HFValidationError:
         return False
 
 
+def is_valid_wandb_path(wandb_path: str) -> bool:
+    """Basic validation for W&B path format (e.g., entity/project/run)."""
+    parts = wandb_path.split("/")
+    return len(parts) == 3
```
aittalam, line 39: I think we've been too overzelous here when we wrote the comment. As per wandb [Job API definition](https://docs.wandb.ai/ref/python/public-api/job) the path is a string with no particular format enforced, and our own tests break when running with this validation. I'd suggest using the following:

```suggestion
    """Basic validation for W&B path format."""
    return bool(path)
```

kernel-loophole: okay sure will update

kernel-loophole: hi @aittalam i think changes will look like this 
```python
def is_valid_wandb_path(wandb_path: str) -> bool:
    """Basic validation for W&B path format (e.g., entity/project/run)."""
    parts = wandb_path.split("/")
    if len(parts) != 3:
        return False

    entity, project, run = parts
    if not all(part.isalnum() for part in parts):
        return False
    
    return True

File: lumigator/python/mzai/jobs/evaluator/evaluator/paths.py [hunk 2]
```diff
@@ -16,50 +15,74 @@ class PathPrefix(str, Enum):
     S3 = "s3://"
     OPENAI = "oai://"
     MISTRAL = "mistral://"
+    LLAMA = "llama://"
```
aittalam, line 18: "llama" might be misleading (eg. it can refer to the llama model, or other local tools to run a model locally). I'd suggest "llamafile" instead:

```suggestion
    LLAMAFILE = "llamafile://"
```

lumigator/python/mzai/jobs/evaluator/evaluator/paths.py
```diff
@@ -16,50 +15,74 @@ class PathPrefix(str, Enum):
...
+    """Basic validation for OpenAI model names."""
+    return bool(model_name)
+
+
+def is_valid_mistral_model_name(model_name: str) -> bool:
+    """Basic validation for Mistral model names."""
+    return bool(model_name)
+
+
+def is_valid_llama_model_name(model_name: str) -> bool:
```
aittalam, line 57:
```suggestion
def is_valid_llamafile_model_name(model_name: str) -> bool:
```

lumigator/python/mzai/jobs/evaluator/evaluator/paths.py
```diff
@@ -16,50 +15,74 @@ class PathPrefix(str, Enum):
...
+    """Basic validation for Llama model names."""
```
aittalam, line 58:
```suggestion
    """Basic validation for Llamafile model names."""
```

lumigator/python/mzai/jobs/evaluator/evaluator/paths.py
```diff
@@ -16,50 +15,74 @@ class PathPrefix(str, Enum):
...
-        # TODO: Validate the OAI model name?
-        pass
+        if not is_valid_openai_model_name(raw_path):
+            raise ValueError(f"'{raw_path}' is not a valid OpenAI model name.")
     elif path.startswith(PathPrefix.MISTRAL):
-        # TODO: Validate the mistral model name?
-        pass
+        if not is_valid_mistral_model_name(raw_path):
+            raise ValueError(f"'{raw_path}' is not a valid Mistral model name.")
+    elif path.startswith(PathPrefix.LLAMA):
```
aittalam, line 83:
```suggestion
    elif path.startswith(PathPrefix.LLAMAFILE):
```

lumigator/python/mzai/jobs/evaluator/evaluator/paths.py
```diff
@@ -16,50 +15,74 @@ class PathPrefix(str, Enum):
...
+        if not is_valid_llama_model_name(raw_path):
```
aittalam, line 84:
```suggestion
        if not is_valid_llamafile_model_name(raw_path):
```

lumigator/python/mzai/jobs/evaluator/evaluator/paths.py
```diff
@@ -87,3 +109,15 @@ def format_artifact_path(artifact: wandb.Artifact) -> AssetPath:
...
+    return f"{PathPrefix.S3.value}{bucket}/{key}"
+
+def format_openai_model_path(model_name: str) -> AssetPath:
+    return f"{PathPrefix.OPENAI.value}{model_name}"
+
+def format_mistral_model_path(model_name: str) -> AssetPath:
+    return f"{PathPrefix.MISTRAL.value}{model_name}"
+
+
+def format_llama_model_path(model_name: str) -> AssetPath:
```
aittalam, line 122:
```suggestion
def format_llamafile_model_path(model_name: str) -> AssetPath:
```

lumigator/python/mzai/jobs/evaluator/evaluator/paths.py
```diff
@@ -87,3 +109,15 @@ def format_artifact_path(artifact: wandb.Artifact) -> AssetPath:
...
+    return f"{PathPrefix.LLAMA.value}{model_name}"
```
aittalam, line 123:
```suggestion
    return f"{PathPrefix.LLAMAFILE.value}{model_name}"
```

lumigator/python/mzai/jobs/evaluator/evaluator/paths.py
kernel-loophole: okay sure thing

lumigator/python/mzai/jobs/evaluator/evaluator/paths.py
```diff
@@ -16,50 +15,74 @@ class PathPrefix(str, Enum):
...
-        # TODO: Validate the W&B path structure?
-        pass
+        if not is_valid_wandb_path(raw_path):
+            raise ValueError(f"'{raw_path}' is not a validpath.")
     elif path.startswith(PathPrefix.S3):
-        # TODO: Validate the S3 path structure?
-        # e.g. if the assumption is we always want a file or a dir, we could
-        # use https://s3pathlib.readthedocs.io to verify (.is_file() or ._is_dir())
-        pass
+        if not is_valid_s3_path(raw_path):
```
aittalam, line 75: I'd suggest to just pass `path` here as we are looking for a full path with its own s3 prefix in the regexp above

```suggestion
        if not is_valid_s3_path(path):
```

aittalam (commented):
Many thanks for your contribution @kernel-loophole!

I have gone through your PR and overall it looks good to me, just a few minor things to fix:

- I have found one of the validation breaks our tests -  I think we were too specific when we described what we wanted to verify and our own unit tests break with the validation you provided, so I suggested a small change to your code.
- I'd use "llamafile" insteada of "llama" around the code as the latter does not specifically refer to the technology we want to enable
- there is just one additional thing to fix and that is in the asset_loader file: in [this line](https://github.com/kernel-loophole/lumigator/blob/main/lumigator/python/mzai/jobs/evaluator/evaluator/jobs/asset_loader.py#L52) we should add `PathPrefix.LLAMAFILE`

I ran both unit and integration tests on lumigator and the evaluator component and they seem to work properly after the fix.

I tried running CI tests on github but they are breaking atm... I am not sure why, but I think the reason might be you are running an older version of the codebase. Would you mind rebasing after committing the suggested change so we can try running these tests again?

Also, here are some instructions if you want to run tests locally.

### lumigator unit / integration tests
```
# from the repo root:
cd ./lumigator/python/mzai/backend
# run tests
SQLALCHEMY_DATABASE_URL=sqlite:///local.db uv run --isolated pytest
```

### evaluator unit tests
```
# from the repo root:
cd ./lumigator/python/mzai/jobs/evaluator
# set the venv for testing evaluator
uv venv --python 3.11.9 && uv pip install pytest && uv pip install -r requirements.txt
# run tests
uv run pytest

### manual test with llamafile
This is how I tested evaluation with a running llamafile server:

- install and run llamafile locally. It will start a server running on a specific port (e.g. 8080)
- upload a dataset on the system
- run an evaluation with the following parameters:


```
{
  "name": "test_run_inference",
  "description": "Test run for inference with llamafile's Mistral",
  "model": "llamafile://mistralai/Mistral-7B-Instruct-v0.2",
  "dataset": "put the id of your dataset here",
  "max_samples": 10,
  "model_url": "http://localhost:8080/v1",
  "system_prompt": "You are a helpful assistant, expert in text summarization. For every prompt you receive, provide a summary of its contents in at most two sentences."
}
```

You can use whatever works best for you for job submission, e.g. copy/paste the above in the api page on http://localhost:8000/docs or run a curl command from your shell. Just a note that you'll need to fix the asset_loader file first, otherwise you'll get a `ValueError: Unable to resolve asset path from llamafile://mistralai/Mistral-7B-Instruct-v0.2.`

kernel-loophole:
@aittalam Thank you for the detailed feedback!.me neither understand why the CI is broken but will look into .and make changes for the `PathPrefix.LLAMAFILE`

kernel-loophole:
hi @aittalam make the changes and run the test locally ,some test failed ,there are some missing dependencies like `s3fs` and `_bz2`

aittalam:
> hi @aittalam make the changes and run the test locally ,some test failed ,there are some missing dependencies like `s3fs` and `_bz2`

TY @kernel-loophole, I'll check your code after the weekend!

kernel-loophole:
@aittalam Thank you for providing direction! I really enjoyed working on this and appreciate all the time you took for multiple reviews.

PR #323, opened 2024-10-29
Remove docker-compose volume mount to avoid .venv conflicts

agpituk:
## What's changing
Docker-compose volume doesn't exist anymore. We sync files instead to synchronize between local computer and the container. The aim is to avoid conflicts between local `.venv ` backend folder and the container `.venv` folders
If `pyproject.toml `is updated the container will be rebuilt.

`local.db` is mounted as a volume so it can persists outside of the container lifecycle and it's synced back from the containers after every change

## How to test it
Run `local-up` and try to run tests, add new dependencies and check if your environment work as usual. Also please ensure to check that the database persists after doing  `local-up` & `local-down` & `local-up` 
Should we update make test to remove the database afterwards?
I want to update the Readme as well which will be done as soon as possible
## I already...

- [N/A] updated the documentation
- [X] checked if a (backend) DB migration step was required and included it if required


macaab26 (approved):

aittalam:
Thanks @agpituk ! Does the ignore on local.db mean that if we e.g. upload something to lumigator it will not persist?

agpituk:
> Thanks @agpituk ! Does the ignore on local.db mean that if we e.g. upload something to Lumigator it will not persist?

It will follow the lifecycle of the container. If we want it to persist local.db should have it's own volume (happy to do that if you think that's beneficial, but keeping the sync for the `.venv`)

aittalam:
> > Thanks @agpituk ! Does the ignore on local.db mean that if we e.g. upload something to Lumigator it will not persist?
> 
> It will follow the lifecycle of the container. If we want it to persist local.db should have it's own volume (happy to do that if you think that's beneficial, but keeping `.venv` out of the sync )

I think it would be beneficial to have local.db persisted even in the local development scenario, the main reason being that we'd like to mimic the general use case (where the db should always persist) as much as possible.

agpituk:
Just added a volume for the database to persist. The only thing I"m not quite happy right now and I want to hear your thoughts @aittalam is the following:

Currently, in our Makefile for running tests we have a `make test` that will launch this (amongst other things)

`cd lumigator/python/mzai/backend; SQLALCHEMY_DATABASE_URL=sqlite:///local.db uv run pytest`

That `local.db` it's in the same path than the one we'll mount inside the container. This was like this before and it's not part of this ticket, but I was talking with @chainlink about this and it doesn't sound right to me to use the same database for testing/real data. Maybe it's out of the scope of this ticket (as that behavior is already there) but I can create a follow up that ensures the testing database is different! (or change it here as it's quite simple)

> > > Thanks @agpituk ! Does the ignore on local.db mean that if we e.g. upload something to Lumigator it will not persist?
> > 
> > 
> > It will follow the lifecycle of the container. If we want it to persist local.db should have it's own volume (happy to do that if you think that's beneficial, but keeping `.venv` out of the sync )
> 
> I think it would be beneficial to have local.db persisted even in the local development scenario, the main reason being that we'd like to mimic the general use case (where the db should always persist) as much as possible. Also, IMO it'd be easier for a developer to remove the db and start from scratch if they needed rather than putting persistence in place if it is not available.

aittalam:
> Just added a volume for the database to persist. The only thing I"m not quite happy right now and I want to hear your thoughts @aittalam is the following:
> 
> Currently, in our Makefile for running tests we have a `make test` that will launch this (amongst other things)
> 
> `cd lumigator/python/mzai/backend; SQLALCHEMY_DATABASE_URL=sqlite:///local.db uv run pytest`
> 
> That `local.db` it's in the same path than the one we'll mount inside the container. This was like this before and it's not part of this ticket, but I was talking with @chainlink about this and it doesn't sound right to me to use the same database for testing/real data. Maybe it's out of the scope of this ticket (as that behavior is already there) but I can create a follow up that ensures the testing database is different! (or change it here as it's quite simple)
> 
> > > > Thanks @agpituk ! Does the ignore on local.db mean that if we e.g. upload something to Lumigator it will not persist?
> > > 
> > > 
> > > It will follow the lifecycle of the container. If we want it to persist local.db should have it's own volume (happy to do that if you think that's beneficial, but keeping `.venv` out of the sync )
> > 
> > 
> > I think it would be beneficial to have local.db persisted even in the local development scenario, the main reason being that we'd like to mimic the general use case (where the db should always persist) as much as possible. Also, IMO it'd be easier for a developer to remove the db and start from scratch if they needed rather than putting persistence in place if it is not available.

That's a great point @agpituk!
I think it is fair to do this in a separate PR, what I would do instead of just providing another file for tests is to make tests depend on a mockup DB instead of the actual one - I think that'd be cleaner.

Makefile
```diff
@@ -57,4 +57,4 @@ clean-all: clean-docker-buildcache clean-docker-containers
 test:
 	make start-lumigator-build
```
ividal, line 58: While we're at it, does it make more sense for `make test` to use start-lumigator-build or, rather, local-up one, since it's the target geared towards local development?

Non-blocking, by the way.

ividal (approved):
Thanks for this!

macaab26 (approved):

Makefile
agpituk: it's a great question. I think this is a broader conversation going on about how we do this and the current mix of unit & integration tests. I'll keep it as it is right now (to sort out all the other problems people are having) and let's open this in a different thread/ticket!

PR #324, opened 2024-10-30
Adding job logs support to health endpoint [LUMI-211]

aittalam:
## What's changing

Added a new endpoint (`/health/jobs/{job_id}/logs`) which connects to the local [ray server API](https://docs.ray.io/en/latest/cluster/running-applications/job-submission/rest.html) (on `/api/jobs/{job_id}/logs`) to get logs given a job id.
This could be used to surface logs to a UI / SDK / generic client both at the end of the job and when it is running (by polling our API).

## How to test it

- start a job
- send a request to the `/health/jobs/{job_id}/logs` API endpoint (e.g. from the http://localhost:8000/docs# interface)

N/A

## I already...

- [x] updated the documentation (some internal docstrings)

### DIFF
--- a/lumigator/python/mzai/backend/backend/api/routes/health.py
+++ b/lumigator/python/mzai/backend/backend/api/routes/health.py
@@ -1,5 +1,5 @@
 import json
-import urllib
+from urllib.parse import urljoin
 from http import HTTPStatus
 from uuid import UUID
 
@@ -21,18 +21,16 @@ def get_health() -> HealthResponse:
 
 @router.get("/jobs/{job_id}")
 def get_job_metadata(job_id: UUID) -> JobSubmissionResponse:
-    resp = requests.get(urllib.parse.urljoin(settings.RAY_JOBS_URL, f"{job_id}"))
+    resp = requests.get(urljoin(settings.RAY_JOBS_URL, f"{job_id}"))
 
     if resp.status_code == HTTPStatus.NOT_FOUND:
-        loguru.logger.error(f"Unexpected status code getting job metadata text: {resp.status_code}")
-        loguru.logger.error(f"Upstream job metadata response: {resp.text or ''}")
+        loguru.logger.error(f"Upstream job metadata not found ({resp.status_code}), error:  {resp.text or ''}")
         raise HTTPException(
             status_code=HTTPStatus.NOT_FOUND,
             detail=f"Job metadata for ID: {job_id} not found",
         )
     elif resp.status_code != HTTPStatus.OK:
-        loguru.logger.error(f"Unexpected status code getting job metadata text: {resp.status_code}")
-        loguru.logger.error(f"Upstream job metadata response: {resp.text or ''}")
+        loguru.logger.error(f"Unexpected status code getting job metadata text: {resp.status_code}, error: {resp.text or ''}")
         raise HTTPException(
             status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
             detail=f"Unexpected error getting job metadata for ID: {job_id}",
@@ -51,18 +49,16 @@ def get_job_metadata(job_id: UUID) -> JobSubmissionResponse:
 
 @router.get("/jobs/{job_id}/logs")
 def get_job_logs(job_id: UUID) -> JobLogsResponse:
-    resp = requests.get(urllib.parse.urljoin(settings.RAY_JOBS_URL, f"{job_id}/logs"))
+    resp = requests.get(urljoin(settings.RAY_JOBS_URL, f"{job_id}/logs"))
 
     if resp.status_code == HTTPStatus.NOT_FOUND:
-        loguru.logger.error(f"Unexpected status code getting job logs: {resp.status_code}")
-        loguru.logger.error(f"Upstream job logs response: {resp.text or ''}")
+        loguru.logger.error(f"Upstream job logs not found: {resp.status_code}, error: {resp.text or ''}")
         raise HTTPException(
             status_code=HTTPStatus.NOT_FOUND,
             detail=f"Job logs for ID: {job_id} not found",
         )
     elif resp.status_code != HTTPStatus.OK:
-        loguru.logger.error(f"Unexpected status code getting job logs: {resp.status_code}")
-        loguru.logger.error(f"Upstream job logs response: {resp}")
+        loguru.logger.error(f"Unexpected status code getting job logs: {resp.status_code}, error: {resp.text or ''}")
         raise HTTPException(
             status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
             detail=f"Unexpected error getting job logs for ID: {job_id} - {resp.text or ''}",
@@ -83,8 +79,7 @@ def get_job_logs(job_id: UUID) -> JobLogsResponse:
 def get_all_jobs() -> list[JobSubmissionResponse]:
     resp = requests.get(settings.RAY_JOBS_URL)
     if resp.status_code != HTTPStatus.OK:
-        loguru.logger.error(f"Unexpected status code getting all jobs: {resp.status_code}")
-        loguru.logger.error(f"Upstream get all jobs response: {resp}")
+        loguru.logger.error(f"Unexpected status code getting all jobs: {resp.status_code}, error: {resp.text or ''}")
         raise HTTPException(
             status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
             detail=f"Unexpected error getting job logs - {resp.text or ''}",
--- a/lumigator/python/mzai/backend/backend/api/routes/health.py
+++ b/lumigator/python/mzai/backend/backend/api/routes/health.py
@@ -1,9 +1,7 @@
 import json
+import urllib
 from http import HTTPStatus
 from uuid import UUID
-from typing import Annotated
-
-import loguru
 
 import loguru
 import requests
@@ -15,9 +13,6 @@ from backend.settings import settings
 
 router = APIRouter()
 
-def ray_jobs() -> str:
-    return f"{settings.RAY_DASHBOARD_URL}/api/jobs/"
-
 @router.get("/")
 def get_health() -> HealthResponse:
     return HealthResponse(deployment_type=settings.DEPLOYMENT_TYPE, status="OK")
@@ -25,19 +20,19 @@ def get_health() -> HealthResponse:
 
 @router.get("/jobs/{job_id}")
 def get_job_metadata(job_id: UUID) -> JobSubmissionResponse:
-    resp = requests.get(ray_jobs() + f"{job_id}")
+    resp = requests.get(urllib.parse.urljoin(settings.ray_jobs, f"{job_id}"))
 
     if resp.status_code == HTTPStatus.NOT_FOUND:
         raise HTTPException(
             status_code=HTTPStatus.NOT_FOUND,
-            detail=f"Job metadata for ID: {job_id} not found",
+            detail=f"Job metadata for ID: {job_id} not found - {resp.text or ''}",
         )
     elif resp.status_code != HTTPStatus.OK:
         loguru.logger.error(f"Unexpected status code getting job metadata text: {resp.status_code}")
         loguru.logger.error(f"Upstream job metadata response: {resp}")
         raise HTTPException(
             status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
-            detail=f"Unexpected error getting job metadata for ID: {job_id}",
+            detail=f"Unexpected error getting job metadata for ID: {job_id} - {resp.text or ''}",
         )
 
     try:
@@ -53,32 +48,58 @@ def get_job_metadata(job_id: UUID) -> JobSubmissionResponse:
 
 @router.get("/jobs/{job_id}/logs")
 def get_job_logs(job_id: UUID) -> JobLogsResponse:
-    resp = requests.get(ray_jobs() + f"{job_id}/logs")
-    if resp.status_code == 200:
-        try:
-            metadata = json.loads(resp.text)
-            return JobLogsResponse(**metadata)
-        except json.JSONDecodeError as e:
-            print(f"JSON decode error: {e}")
-            print(f"Response text: {resp.text}")
-            return {"error": "Invalid JSON response"}
-    else:
-        return {"error": f"HTTP error {resp.status_code}"}
+    resp = requests.get(urllib.parse.urljoin(settings.ray_jobs, f"{job_id}/logs"))
+
+    if resp.status_code == HTTPStatus.NOT_FOUND:
+        raise HTTPException(
+            status_code=HTTPStatus.NOT_FOUND,
+            detail=f"Job logs for ID: {job_id} not found - {resp.text or ''}",
+        )
+    elif resp.status_code == HTTPStatus.BAD_REQUEST:
+        raise HTTPException(
+            status_code=HTTPStatus.BAD_REQUEST,
+            detail=f"Job ID: {job_id} not submitted via the Ray API - {resp.text or ''}",
+        )
+    elif resp.status_code != HTTPStatus.OK:
+        loguru.logger.error(f"Unexpected status code getting job logs text: {resp.status_code}")
+        loguru.logger.error(f"Upstream job metadata response: {resp}")
+        raise HTTPException(
+            status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
+            detail=f"Unexpected error getting job logs for ID: {job_id} - {resp.text or ''}",
+        )
+
+    try:
+        metadata = json.loads(resp.text)
+        return JobLogsResponse(**metadata)
+    except json.JSONDecodeError as e:
+        loguru.logger.error(f"JSON decode error: {e}")
+        loguru.logger.error(f"Response text: {resp.text}")
+        raise HTTPException(
+            status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
+            detail="Invalid JSON response"
+        ) from e
 
 
 @router.get("/jobs/")
 def get_all_jobs() -> list[JobSubmissionResponse]:
-    resp = requests.get(ray_jobs())
-    if resp.status_code == 200:
-        try:
-            metadata = json.loads(resp.text)
-            submissions: list[JobSubmissionResponse] = [
-                JobSubmissionResponse(**item) for item in metadata
-            ]
-            return submissions
-        except json.JSONDecodeError as e:
-            print(f"JSON decode error: {e}")
-            print(f"Response text: {resp.text}")
-            return {"error": "Invalid JSON response"}
-    else:
-        return {"error": f"HTTP error {resp.status_code}"}
+    resp = requests.get(settings.ray_jobs)
+    if resp.status_code != HTTPStatus.OK:
+        loguru.logger.error(f"Unexpected status code getting job logs text: {resp.status_code}")
+        loguru.logger.error(f"Upstream job metadata response: {resp}")
+        raise HTTPException(
+            status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
+            detail=f"Unexpected error getting job logs - {resp.text or ''}",
+        )
+    try:
+        metadata = json.loads(resp.text)
+        submissions: list[JobSubmissionResponse] = [
+            JobSubmissionResponse(**item) for item in metadata
+        ]
+        return submissions
+    except json.JSONDecodeError as e:
+        loguru.logger.error(f"JSON decode error: {e}")
+        loguru.logger.error(f"Response text: {resp.text}")
+        raise HTTPException(
+            status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
+            detail="Invalid JSON response"
+        ) from e
--- a/lumigator/python/mzai/backend/backend/api/routes/health.py
+++ b/lumigator/python/mzai/backend/backend/api/routes/health.py
@@ -7,7 +7,7 @@ import loguru
 import requests
 from fastapi import APIRouter, HTTPException
 from lumigator_schemas.extras import HealthResponse
-from lumigator_schemas.jobs import JobSubmissionResponse
+from lumigator_schemas.jobs import JobSubmissionResponse, JobLogsResponse
 
 from backend.settings import settings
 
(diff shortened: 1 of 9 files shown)


javiermtorres:
I definitely think this is a helpful addition.

lumigator/python/mzai/backend/backend/tests/api/routes/test_health.py
```diff
@@ -4,6 +4,29 @@
...
+    return buffer.read()
+
+
+@pytest.fixture
+def valid_experiment_dataset() -> str:
+    data = [
+        ["examples", "ground_truth"],
+        ["Hello World", "Hello"],
+    ]
+    return format_dataset(data)
```
aittalam, line 29: We have identical code in [test_datasets](https://github.com/mozilla-ai/lumigator/blob/main/lumigator/python/mzai/backend/backend/tests/api/routes/test_datasets.py), what about putting the common stuff in [conftest](https://github.com/mozilla-ai/lumigator/blob/main/lumigator/python/mzai/backend/backend/tests/conftest.py) instead?

lumigator/python/mzai/backend/backend/tests/api/routes/test_health.py
```diff
@@ -60,3 +83,45 @@ def test_get_job_metadata_ok(app_client: TestClient,
     actual = json.loads(response.text)
     expected = load_json(json_data_health_job_metadata_ok)
     assert expected == actual
+
+def test_job_logs(app_client: TestClient, valid_experiment_dataset: str):
```
aittalam, line 87: I think this is fine for an integration test, but perhaps here we'd better be focusing on the individual functionality without uploading datasets or running actual jobs. Two mock responses, one for a missing job and another one for an existing one, should probably be enough. You can validate both wrt `JobLogsResponse` and assert you are getting the proper status codes.

lumigator/python/mzai/backend/backend/api/routes/health.py
```diff
@@ -1,17 +1,22 @@
...
 from fastapi import APIRouter, HTTPException
 from schemas.extras import HealthResponse
-from schemas.jobs import JobSubmissionResponse
+from schemas.jobs import JobLogsResponse, JobSubmissionResponse
 
 from backend.settings import settings
 
 router = APIRouter()
 
+def ray_jobs() -> str:
```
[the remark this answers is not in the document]
peteski22: I'd agree that it might be better suited being composed in backend settings. 

❓ Should`RAY_DASHBOARD_URL` and anything we might add, return a stronger type than `str` (`Path` or `URL`)? 

If it stays, maybe make this 'private' with the underscore? `_get_ray_jobs_url()`

Whichever way it goes I think it would be worth adding a doc string to explain what it should be sending back (including if the trailing slash is expected to be present).

lumigator/python/mzai/backend/backend/api/routes/health.py
```diff
@@ -46,9 +51,24 @@ def get_job_metadata(job_id: UUID) -> JobSubmissionResponse:
             detail="Invalid JSON response"
         ) from e
 
+@router.get("/jobs/{job_id}/logs")
+def get_job_logs(job_id: UUID) -> JobLogsResponse:
+    resp = requests.get(ray_jobs() + f"{job_id}/logs")
+    if resp.status_code == 200:
```
peteski22, line 57:
```suggestion
    if resp.status_code == HTTPStatus.OK:
```

lumigator/python/mzai/backend/backend/api/routes/health.py
```diff
@@ -46,9 +51,24 @@ def get_job_metadata(job_id: UUID) -> JobSubmissionResponse:
...
+        try:
+            metadata = json.loads(resp.text)
+            return JobLogsResponse(**metadata)
+        except json.JSONDecodeError as e:
+            print(f"JSON decode error: {e}")
+            print(f"Response text: {resp.text}")
+            return {"error": "Invalid JSON response"}
+    else:
+        return {"error": f"HTTP error {resp.status_code}"}
```
peteski22, line 66: It might be useful to take a look at this recently merged PR to see how we have handled 200/400/{other}, so we don't just return `200 - OK` with some JSON containing an error.

https://github.com/mozilla-ai/lumigator/pull/318/files#diff-bc21eb24ef7e8d2c97660608a2ca09d7f46e7f6ffd2f749c4c75e03ad8bb20e8R25-R47

lumigator/python/mzai/backend/backend/tests/api/routes/test_health.py
peteski22: This might be useful for mocking the return data from Ray to exercise the cases you want in the route code:

https://github.com/mozilla-ai/lumigator/pull/318/files#diff-199b80c426fab13e604600bdb49d196d54459bc0b34bf6d027ed49b80ee4081cR23-R26

peteski22 (commented):
Added some comments, give me a shout if you want to chat tomorrow (happy to pair too if you fancy it) 👍🏼

lumigator/python/mzai/backend/backend/api/routes/health.py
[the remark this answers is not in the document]
javiermtorres: Hum. Ok, I'll move into settings, although I see it as a bit api-specific for the moment. I feel that it is growing too much to fit in a single class, but there are no api or service specific property classes right now.

Re: typing, right now the RAY_DASHBOARD_URL is a str, but I agree with moving it into a URL (better than a Path, imho). @veekaybee wdyt?

Ok to commenting, plus I'll take the opportunity to comment some other property in settings too.

lumigator/python/mzai/backend/backend/api/routes/health.py
```diff
@@ -46,9 +51,24 @@ def get_job_metadata(job_id: UUID) -> JobSubmissionResponse:
             detail="Invalid JSON response"
         ) from e
 
+@router.get("/jobs/{job_id}/logs")
+def get_job_logs(job_id: UUID) -> JobLogsResponse:
+    resp = requests.get(ray_jobs() + f"{job_id}/logs")
```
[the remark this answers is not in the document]
javiermtorres: One thing I don't like is that urllib handles mainly str params instead of the specialized type for urls, afaict. But I guess that's better than joining the urls together, ok.

lumigator/python/mzai/backend/backend/api/routes/health.py
javiermtorres: I'll change the general approach to errors to fit your current impl in getting jobs above.

lumigator/python/mzai/backend/backend/tests/api/routes/test_health.py
javiermtorres: Should we make a standalone conftest for api tests alone?

lumigator/python/mzai/backend/backend/tests/api/routes/test_health.py
javiermtorres: I think that integration testing is more appropriate here than a basic test, since we're dealing with a third party whose behaviour may change at some point. That said, I'll include also a basic test with mocked requests for completion.

_That said_, I think we should open an issue to separate basic tests from integration tests, and execute them in separate ways so that the external services need not be started always, and the basic tests are executed first for faster failures. Wdyt?

lumigator/python/mzai/backend/backend/api/routes/health.py
javiermtorres: BTW, let's keep an eye on https://docs.ray.io/en/latest/cluster/running-applications/job-submission/api.html?_gl=1*19sigrd*_up*MQ..*_ga*MTMzMzI3MjIxMS4xNzMxNDAxMjQ4*_ga_0LCWHW1N3S*MTczMTQwMTI0OC4xLjAuMTczMTQwMTI0OC4wLjAuMA.. 
I'll use that as reference for possible status codes.

lumigator/python/mzai/backend/backend/api/routes/health.py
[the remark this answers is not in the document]
veekaybee: Generally URLs are passed around as either Paths or strings in Python, and urllib is also fine but if we're looking for a stronger contract, we can use Pydantic's URL types. 

I generally wouldn't recommend Pydantic out of the box, but since we already use Pydantic URL classes, and we have more than a few URLs to parse in settings, we could use them here and constrain the other URLs in settings the same way https://docs.pydantic.dev/2.0/usage/types/urls/

the other benefit is we can pull all the expected elements from the model (from the docs)

```python
from pydantic import (
    BaseModel,
    HttpUrl,
)


class MyModel(BaseModel):
    url: HttpUrl


m = MyModel(url='http://www.example.com')

# the repr() method for a url will display all properties of the url
print(repr(m.url))
#> Url('http://www.example.com/')
print(m.url.scheme)
#> http
print(m.url.host)
#> www.example.com
print(m.url.port)
#> 80
```

Whatever we go with, it would make sense to standardize across the app for legibility.

lumigator/python/mzai/backend/backend/tests/api/routes/test_health.py
peteski22: Can't we have both (unit and integration tests)? 😄

javiermtorres: My bad. I didn't push my latest work :-/ Yes, the plan was to have both (and later on separate basic vs integration).

lumigator/python/mzai/backend/backend/tests/api/routes/test_health.py
javiermtorres: Ok, let's keep a single one for the moment.

aittalam:
Hey @javiermtorres, I merged your latest updates with main and  did some linting but I think overall this looks good.
I tested both with `make test` and with scripts to upload data and run both eval and inference jobs.
I'd suggest to track follow-up tasks (if it has not been done already) as issues, namely unit vs integration tests, providing consistent error handling across routes/services, and consistent schemas for url handling. WDYT?

lumigator/python/mzai/backend/backend/settings.py
```diff
@@ -99,11 +99,20 @@ def RAY_WORKER_GPUS(self) -> float:  # noqa: N802
...
+    # URL for the Ray Dashboard
     @computed_field
     @property
     def RAY_DASHBOARD_URL(self) -> str:  # noqa: N802
         return f"http://{self.RAY_HEAD_NODE_HOST}:{self.RAY_DASHBOARD_PORT}"
 
+    # URL for Ray jobs API
+    @computed_field
+    @property
+    def ray_jobs(self) -> str:  # noqa: N802
```
peteski22, line 111: ❓ Should we match the UPPERCASING we have with other methods we have in here? (e.g. `RAY_JOBS` to match `RAY_DASHBOARD_URL`, `SQLALCHEMY_DATABASE_URL` etc.).

❓ What do you think about `RAY_JOBS_URL` (vs. `RAY_JOBS`) to align with the other methods that are returning a 'URL' (as a URL or as a string).

lumigator/python/mzai/backend/backend/api/routes/health.py
```diff
@@ -20,19 +21,19 @@ def get_health() -> HealthResponse:
...
 @router.get("/jobs/{job_id}")
 def get_job_metadata(job_id: UUID) -> JobSubmissionResponse:
-    resp = requests.get(f"{settings.RAY_DASHBOARD_URL}/api/jobs/{job_id}")
+    resp = requests.get(urllib.parse.urljoin(settings.ray_jobs, f"{job_id}"))
 
     if resp.status_code == HTTPStatus.NOT_FOUND:
         raise HTTPException(
             status_code=HTTPStatus.NOT_FOUND,
-            detail=f"Job metadata for ID: {job_id} not found",
+            detail=f"Job metadata for ID: {job_id} not found - {resp.text or ''}",
```
peteski22, line 29: 🤔 I don't think we should be leaking this straight through the API. 

```suggestion
            detail=f"Job metadata for ID: {job_id} not found",
```

lumigator/python/mzai/backend/backend/api/routes/health.py
```diff
@@ -20,19 +21,19 @@ def get_health() -> HealthResponse:
...
         )
     elif resp.status_code != HTTPStatus.OK:
         loguru.logger.error(f"Unexpected status code getting job metadata text: {resp.status_code}")
         loguru.logger.error(f"Upstream job metadata response: {resp}")
         raise HTTPException(
             status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
-            detail=f"Unexpected error getting job metadata for ID: {job_id}",
+            detail=f"Unexpected error getting job metadata for ID: {job_id} - {resp.text or ''}",
```
peteski22, line 36: 🤔 same here, we intentionally log the response but we don't just leak it back to the caller.

```suggestion
            detail=f"Unexpected error getting job metadata for ID: {job_id}",
```

lumigator/python/mzai/backend/backend/api/routes/health.py
```diff
@@ -41,24 +42,63 @@ def get_job_metadata(job_id: UUID) -> JobSubmissionResponse:
...
+    if resp.status_code == HTTPStatus.NOT_FOUND:
+        raise HTTPException(
+            status_code=HTTPStatus.NOT_FOUND,
+            detail=f"Job logs for ID: {job_id} not found - {resp.text or ''}",
+        )
+    elif resp.status_code == HTTPStatus.BAD_REQUEST:
+        raise HTTPException(
+            status_code=HTTPStatus.BAD_REQUEST,
+            detail=f"Job ID: {job_id} not submitted via the Ray API - {resp.text or ''}",
+        )
```
peteski22, line 63: If we're making a bad request to Ray, that's on 'us' the backend, not the caller, we should have verified we're happy with whatever parameters we're sending upstream beforehand. 

If the caller has supplied a UUID (which has to be valid otherwise Pydantic would have raised a validation error), why should they be told it was a bad request? It wasn't.

I think we should stick with letting other status codes fall through to the `500 - Internal Server Error` as we're doing something wrong not the caller.

lumigator/python/mzai/backend/backend/api/routes/health.py
```diff
@@ -41,24 +42,63 @@ def get_job_metadata(job_id: UUID) -> JobSubmissionResponse:
...
+    elif resp.status_code != HTTPStatus.OK:
+        loguru.logger.error(f"Unexpected status code getting job logs text: {resp.status_code}")
+        loguru.logger.error(f"Upstream job metadata response: {resp}")
         raise HTTPException(
             status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
-            detail="Invalid JSON response"
+            detail=f"Unexpected error getting job logs for ID: {job_id} - {resp.text or ''}",
```
peteski22, line 69: Same thing about leaky abstractions.

```suggestion
            detail=f"Unexpected error getting job logs for ID: {job_id}''}",
```

lumigator/python/mzai/backend/backend/api/routes/health.py
```diff
@@ -41,24 +42,63 @@ def get_job_metadata(job_id: UUID) -> JobSubmissionResponse:
...
-        except json.JSONDecodeError as e:
-            print(f"JSON decode error: {e}")
-            print(f"Response text: {resp.text}")
-            return {"error": "Invalid JSON response"}
-    else:
-        return {"error": f"HTTP error {resp.status_code}"}
+    resp = requests.get(settings.ray_jobs)
+    if resp.status_code != HTTPStatus.OK:
+        loguru.logger.error(f"Unexpected status code getting job logs text: {resp.status_code}")
+        loguru.logger.error(f"Upstream job metadata response: {resp}")
```
peteski22, line 88:
```suggestion
        loguru.logger.error(f"Upstream get jobs response: {resp}")
```

lumigator/python/mzai/backend/backend/api/routes/health.py
```diff
@@ -41,24 +42,63 @@ def get_job_metadata(job_id: UUID) -> JobSubmissionResponse:
...
-            return submissions
-        except json.JSONDecodeError as e:
-            print(f"JSON decode error: {e}")
-            print(f"Response text: {resp.text}")
-            return {"error": "Invalid JSON response"}
-    else:
-        return {"error": f"HTTP error {resp.status_code}"}
+    resp = requests.get(settings.ray_jobs)
+    if resp.status_code != HTTPStatus.OK:
+        loguru.logger.error(f"Unexpected status code getting job logs text: {resp.status_code}")
```
peteski22, line 87:
```suggestion
        loguru.logger.error(f"Unexpected status code getting jobs: {resp.status_code}")
```

lumigator/python/mzai/backend/backend/api/routes/health.py
```diff
@@ -41,24 +42,63 @@ def get_job_metadata(job_id: UUID) -> JobSubmissionResponse:
...
+        loguru.logger.error(f"Upstream job metadata response: {resp}")
+        raise HTTPException(
+            status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
+            detail=f"Unexpected error getting job logs - {resp.text or ''}",
```
peteski22, line 91:
```suggestion
            detail=f"Unexpected error getting jobs",
```

peteski22 (changes requested):
I think I might have caught a few typos, and added some comments. Marking as request changes just so we don't accidentally merge without addressing the typos. 😄

lumigator/python/mzai/backend/backend/api/routes/health.py
aittalam: Thanks @peteski22 , I was not 100% sure about this but I thought it was covering some use case that Javier had encountered and I could not replicate. I agree that if it is something we do not foresee now (and it cannot be invalid UUID, which is the only parameter we are passing and is validated) we should just raise a 500.

lumigator/python/mzai/backend/backend/api/routes/health.py
aittalam: I can agree for the 404, but in case of an error we are not addressing explicitly isn't it better if we provide extra info in the HTTP response too (and not just the logs)? I just want to be mindful this is not a service but a tool that users will run on their own, and providing more information here might save from having to go through the logs. WDYT?

lumigator/python/mzai/backend/backend/api/routes/health.py
aittalam: Same concern as above

lumigator/python/mzai/backend/backend/settings.py
javiermtorres: 1) Could be. I didn't check if properties were all also in uppercase. I'll check.

1) Fine for me 👍

lumigator/python/mzai/backend/backend/api/routes/health.py
javiermtorres: I usually prefer to give the underlying info, but I have usually worked with very tightly integrated systems. I assume you'd prefer to log this info instead.

lumigator/python/mzai/backend/backend/api/routes/health.py
javiermtorres: This may depend on the specific setup (private use, service, etc). What about settings to log and/or send back the problem details?

lumigator/python/mzai/backend/backend/api/routes/health.py
javiermtorres: The UUID might be referring to a non-existing job, for example, or be a non-type 4 UUID. IMHO the final decision is on Ray and we should stick to their response. Also, we don't have an internal error. Our code did not fail or have an exception. Some other SW tells us this is a bad request, so IMHO we need to say that this was a bad request (and hopefully include useful info from Ray as to why this is so).

lumigator/python/mzai/backend/backend/api/routes/health.py
javiermtorres: Errors from external systems will be logged instead of sent to the user.

lumigator/python/mzai/backend/backend/api/routes/health.py
javiermtorres: 404 stays mapped to 404, 400 will translate to 500.

lumigator/python/mzai/backend/backend/api/routes/health.py
javiermtorres: As decided, 400 will translate to 500.

lumigator/python/mzai/backend/backend/api/routes/health.py
javiermtorres: Ok

lumigator/python/mzai/backend/backend/api/routes/health.py
javiermtorres: Ok

lumigator/python/mzai/backend/backend/api/routes/health.py
javiermtorres: Ok

lumigator/python/mzai/backend/backend/api/routes/health.py
peteski22: Was this missed or was there consensus against changing it?

lumigator/python/mzai/backend/backend/api/routes/health.py
peteski22: Was this missed or was there consensus against changing it?

peteski22 (commented):
A few (more) comments, apologies.

lumigator/python/mzai/backend/backend/api/routes/health.py
```diff
@@ -1,12 +1,13 @@
 import json
+import urllib
```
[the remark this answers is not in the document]
javiermtorres: Ok with this. If the item might get mixed up with a similar name I usually prefer qualified names, but I agree this is not the case.

lumigator/python/mzai/backend/backend/api/routes/health.py
```diff
@@ -20,16 +21,18 @@ def get_health() -> HealthResponse:
 
 @router.get("/jobs/{job_id}")
 def get_job_metadata(job_id: UUID) -> JobSubmissionResponse:
-    resp = requests.get(f"{settings.RAY_DASHBOARD_URL}/api/jobs/{job_id}")
+    resp = requests.get(urllib.parse.urljoin(settings.RAY_JOBS_URL, f"{job_id}"))
 
     if resp.status_code == HTTPStatus.NOT_FOUND:
+        loguru.logger.error(f"Unexpected status code getting job metadata text: {resp.status_code}")
+        loguru.logger.error(f"Upstream job metadata response: {resp.text or ''}")
```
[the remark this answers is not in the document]
javiermtorres: I'd address the tangent in another PR for discussion. All those are "OK", but maybe not all are handled in the same way. A 200 may carry content while a 204 may not; a 202 response to a POST as opposed to a 201 may mean that the resource is not available yet. I'd rather handle each of these cases separately.

PR #327, opened 2024-10-30
chore: tear down containers launched after make test

ividal:
## What's changing
stop-lumigator is now called inside make test, to clean up.

## How to test it
Clean folder, clone, docker ps shows no (lumigator-related) running containers, make test, docker ps shows no running containers (lumigator-related) and no unrelated containers were stopped.

We could discuss tearing down only those started by make test, but hopefully this is already better than not tearing down lumigator containers after the test.

Pre-commit hooks also fixed some whitespaces in the Makefile.
 
## I already...

- [ ] updated the documentation --> N/A
- [ ] checked if a (backend) DB migration step was required and included it if required --> N/A


chainlink (approved):

ividal:
Removed the note about disk size, since it was way above what is necessary and the Docker default.
Merging now, thanks!

PR #328, opened 2024-10-30
LUMI-127 adding inference API

aittalam:
## What's changing
Added support for inference in our API:
- new schema for input parameters and update job route
- created ad-hoc inference configuration template and updated job service code to populate it
Updated inference job to support the new configuration:
- defined an `InferenceJobConfig` pydantic class to validate input dictionary
- updated inference code to make use of it
- added unit tests to validate minimal / full configurations

## How to test it

- unit tests on backend and SDK with `make test`
- unit tests on inference job: `cd lumigator/python/mzai/jobs/inference; uv run --with pytest --with-requirements requirements.txt --isolated pytest`
- manual test: `test_dataset_upload && test_inference_mistral`

test_dataset_upload:
```
#!/bin/bash
if [ "$#" -gt 0 ]; then
    DATA_CSV_PATH="$1"
else
    DATA_CSV_PATH="$HOME/Downloads/thunderbird_gt_bart.csv"
fi

fi

curl -s $BACKEND_URL/api/v1/datasets/ \
  -F 'format=job' | jq
```

test_inference_mistral:
```
#!/bin/bash

fi

EVAL_NAME="test_run_inference"
EVAL_DESC="Test run for inference with Mistral"
EVAL_MODEL="mistral://open-mistral-7b"
EVAL_DATASET=$DATASET_ID
EVAL_MAX_SAMPLES="0"

JSON_STRING=$(jq -n \
                --arg name "$EVAL_NAME" \
                --arg desc "$EVAL_DESC" \
                --arg model "$EVAL_MODEL" \
                '{name: $name, description: $desc, model: $model, dataset: $dataset_id, max_samples: $max_samples}' )

curl -s $BACKEND_URL/api/v1/jobs/inference/ \
  -d "$JSON_STRING" | jq
```

## I already...

- [ ] updated the documentation: NO (the current job did not exist before and I wanted to complete the sequence of PRs before having full documentation for this, please LMK if that's ok otherwise I am happy to add some)
- [x] checked if a (backend) DB migration step was required and included it if required: no DB migration step was required (this PR does not change the tables schema)


.vscode/settings.json
```diff
@@ -8,5 +8,8 @@
             "source.organizeImports.ruff": "explicit"
         }
     },
-    "python.testing.pytestEnabled": true
+    "python.testing.pytestEnabled": true,
+    "python.analysis.extraPaths": [
```
aittalam, line 12: minor add for VSCode, it had to look for schemas too as it appeared as a separate dependency

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -99,13 +100,18 @@ def create_inference_job(self, request: JobCreate) -> JobResponse:
...
+            "max_tokens": request.max_tokens,
+            "frequency_penalty": request.frequency_penalty,
+            "temperature": request.temperature,
+            "top_p": request.top_p,
         }
 
         # load a config template and fill it up with config_params
-        if request.config_infer_template is not None:
-            config_template = request.config_infer_template
+        if request.config_template is not None:
```
aittalam, line 113: @veekaybee in this case the config template is just an input parameter and I thought about keeping the name identical, so at some point this code could be factored out as being practically identical across different jobs (also, we might just remove it if we end up using pydantic configs here too...)

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -7,7 +7,8 @@
 from schemas.extras import ListingResponse
 from schemas.jobs import (
     JobConfig,
-    JobCreate,
+    JobEvalCreate,
```
aittalam, line 10: This is to accomodate different input schemas for different input jobs

lumigator/python/mzai/jobs/inference/inference_config.py
```diff
@@ -0,0 +1,39 @@
+from pydantic import BaseModel, ConfigDict
+
+
+class DatasetConfig(BaseModel):
+    path: str
+    model_config = ConfigDict(extra="forbid")
```
aittalam, line 6: This field specifies that we do not accept any other parameters in addition to the specified ones. LMK what you think about this, I guess it makes sense to not accept just everything but we could also choose to not break and simply ignore extra stuff.

aittalam:
@javiermtorres I added you as a reviewer because I saw the [SDK integration test failed](https://github.com/mozilla-ai/lumigator/actions/runs/11601798328/job/32305358034), but I had no issues testing locally with `make test`. lmk how I can help fix that (or do schemas just need to be rebuilt?)

javiermtorres:
@aittalam I will include the SDK integration tests in the test target (maybe they're not already included). 

The issue is the split of JobCreate, so it is not found by the integration tests. Let me provide a branch that can be merged into yours to solve this issue.

lumigator/python/mzai/schemas/schemas/jobs.py
```diff
@@ -50,16 +50,31 @@ class JobSubmissionResponse(BaseModel):
     driver_exit_code: int | None = None
 
 
-class JobCreate(BaseModel):
+class JobEvalCreate(BaseModel):
```
javiermtorres, line 53: I'd advice against this split. I'd rather make something like this instead:

```
class JobCreate(BaseModel):
    name: str
    description: str = ""
    model: str
    dataset: UUID
    max_samples: int | None = None
    model_url: str | None = None
    job_specific_config: BaseModel | None = None
```

So each job can pass its own config without further specification, and when using model_validate in the backend each job can check their own config with their own model.

veekaybee:
For reference, complementary job in progress: https://github.com/mozilla-ai/lumigator/pull/333

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -3,7 +3,8 @@
 from fastapi import APIRouter, status
 from schemas.extras import ListingResponse
 from schemas.jobs import (
-    JobCreate,
+    JobEvalCreate,
+    JobInferenceCreate,
```
veekaybee, line 7: Oh perfect, I did this too, it will work nicely to separate out the templates.

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -17,15 +18,15 @@
...
+    request: JobInferenceCreate,
 ) -> JobResponse:
     return service.create_inference_job(request)
 
 
 @router.post("/evaluate/", status_code=status.HTTP_201_CREATED)
 def create_evaluation_job(
     service: JobServiceDep,
-    request: JobCreate,
+    request: JobEvalCreate,
```
veekaybee, line 29: 👍

lumigator/python/mzai/backend/backend/services/jobs.py
veekaybee: yep so you created the same parameter but just two different jobs, makes sense

lumigator/python/mzai/jobs/inference/inference.py
```diff
@@ -86,24 +86,24 @@ def run_inference(config: Box) -> Path:
 
     # Enable / disable tqdm
     input_samples = dataset["examples"]
-    dataset_iterable = tqdm(input_samples) if config.evaluation.enable_tqdm else input_samples
+    dataset_iterable = tqdm(input_samples) if config.job.enable_tqdm else input_samples
```
veekaybee, line 89: have been thinking...do we still want the option to disable tqdm? we should probably leave on by default

lumigator/python/mzai/jobs/inference/inference.py
```diff
@@ -86,24 +86,24 @@ def run_inference(config: Box) -> Path:
...
 
     # Choose which model client to use
-    if config.model.inference is not None:
+    if config.server is not None:
```
veekaybee, line 92: could we call this something else, perhaps `model_client` then?

lumigator/python/mzai/jobs/inference/inference_config.py
veekaybee: I think we should be strict here - we have so many places we pass around config in the app and so many places it breaks (I know from testing over the past few days 😅 ) that any place we can hammer down these params will be helpful across the codebase.

lumigator/python/mzai/jobs/inference/tests/conftest.py
```diff
@@ -0,0 +1,23 @@
+import json
+from pathlib import Path
+
+import pytest
+
+
+def load_json(path: Path) -> dict:
```
veekaybee, line 7: do we need to load json directly from a file for these? or is it enough to have the direct structure, aka we could do:

```
def input_json():  
    return json.dumps(the minimal config here) 
```    
because we're not really testing the file itself but the actions on the json

lumigator/python/mzai/jobs/inference/inference_config.py
```diff
@@ -0,0 +1,39 @@
...
+    frequency_penalty: float = 0.0
+    temperature: float = 1.0
+    top_p: float = 1.0
+    model_config = ConfigDict(extra="forbid")
+
+
+class InferenceJobConfig(BaseModel):
+    name: str
+    dataset: DatasetConfig
+    job: JobConfig
```
ividal, line 36: Oooh, honest question: what's worse: inheritance or nested configs (even if they are in Pydantic)?

ividal (commented):
Thanks, @aittalam ! 

I'd be happy to ship this and start using it if we can remove part of the dupe code create_inference and create_eval have 😄 Just cleaning up one corner at a time.

lumigator/python/mzai/schemas/schemas/jobs.py
ividal: Just seeing this after proposing something - [wdyt of this](https://github.com/mozilla-ai/lumigator/pull/328#discussion_r1826153789)?

lumigator/python/mzai/backend/backend/services/jobs.py
aittalam: Thanks Irina! My reasoning is that, since we are removing inference from the evaluation job (see Vicki's [draft PR](https://github.com/mozilla-ai/lumigator/pull/333)), we won't have such a big overlap anymore.
I agree on factoring as much code together as possible though and, despite preferring to keep different, more specific pydantic configs (until we actually see the need to group stuff together because there is a non-trivial overlap), I think we could do better in terms of factoring out commonalities in the code

lumigator/python/mzai/jobs/inference/inference.py
aittalam: I always found it useful, but am not against leaving a chance to turn it off (one use case could be e.g. huge datasets which might generate a lot of ray logs per run)

lumigator/python/mzai/jobs/inference/inference.py
aittalam: In this case it is actually a server, so we could call it `inference_server`. Wdyt?

lumigator/python/mzai/jobs/inference/tests/conftest.py
aittalam: Yeah my reasoning was that examples might grow but I am ok to start with just a few dictionaries (rn it is 2 but I was already considering adding more, at least one for each model time)

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -69,7 +70,7 @@ def _get_results_s3_key(self, job_id: UUID) -> str:
             / settings.S3_JOB_RESULTS_FILENAME.format(job_name=record.name, job_id=record.id)
         )
 
-    def create_inference_job(self, request: JobCreate) -> JobResponse:
+    def create_inference_job(self, request: JobInferenceCreate) -> JobResponse:
```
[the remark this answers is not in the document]
aittalam: I 100% agree with having some common code to deal with jobs!

I also agree about not putting the large change in this PR 😄, but I am happy to start factoring out the common code. Just a heads up though that we have [another PR dual to this](https://github.com/mozilla-ai/lumigator/pull/333) going on and it might help testing the two in isolation first, then using the common method.

lumigator/python/mzai/jobs/inference/inference_config.py
aittalam: Good question!

I think they serve two purposes: e.g. we have a very clear common class (jobs) and I would like all of them to share the base configuration as root parameters, but I'd like other task-specific params (e.g. inference server details for jobs that use it, hf model parameters for those which support them, etc) to be available as subtrees of the main config (mostly to prevent issues with multiple inheritance). 

In both cases the presence of shared classes across jobs, which is something we were not planning to have from the start, might help (intuitively, I think in the case of inheritance not having shared classes might even be counterproductive).

lumigator/python/mzai/backend/backend/config_templates.py
```diff
@@ -83,42 +85,55 @@
...
-    "oai://gpt-4o-mini": oai_eval_template,
-    "oai://gpt-4-turbo": oai_eval_template,
-    "oai://gpt-3.5-turbo-0125": oai_eval_template,
-    "mistral://open-mistral-7b": oai_eval_template,
-    "llamafile://mistralai/Mistral-7B-Instruct-v0.2": oai_eval_template,
+templates = {
+    JobType.INFERENCE: {
+        "default": causal_infer_template,
+        "hf://facebook/bart-large-cnn": bart_eval_template,
+        "hf://mikeadimech/longformer-qmsum-meeting-summarization": seq2seq_eval_template,
```
veekaybee, line 116: should these be `_infer_` ?

lumigator/python/mzai/backend/backend/services/datasets.py
```diff
@@ -201,6 +201,9 @@ def delete_dataset(self, dataset_id: UUID) -> None:
         self.dataset_repo.delete(record.id)
 
 
+        # Delete DB record
```
veekaybee, line 204: i think we have this delete twice (see 201)

lumigator/python/mzai/backend/backend/config_templates.py
aittalam: Thanks for raising this! Actually I should remove them as we are not using them rn. 🙏

lumigator/python/mzai/sdk/tests/test_jobs.py
```diff
@@ -17,8 +17,10 @@ def test_create_job_ok_all(
     mock_requests_response.json = lambda: data
 
     job_json = load_json(json_data_job_all)
-    print(JobCreate.model_validate(job_json))
-    job_ret = lumi_client.jobs.create_job(JobType.EVALUATION, JobCreate.model_validate(job_json))
+    print(JobEvalCreate.model_validate(job_json))
```
veekaybee, line 20: let's remove print logging

lumigator/python/mzai/sdk/tests/test_jobs.py
```diff
@@ -37,8 +39,8 @@ def test_create_job_ok_minimal(
     mock_requests_response.json = lambda: data
 
     job_json = load_json(json_data_job_minimal)
-    print(JobCreate.model_validate(job_json))
-    job_ret = lumi_client.jobs.create_job(JobType.INFERENCE, JobCreate.model_validate(job_json))
+    print(JobEvalCreate.model_validate(job_json))
```
veekaybee, line 42: let's remove print logging

veekaybee:
LGTM after those few changes, thanks for all your patience on this! 🚀

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -24,6 +26,27 @@
 
 
 class JobService:
+    # set storage path
+    storage_path = f"s3://{ Path(settings.S3_BUCKET) / settings.S3_JOB_RESULTS_PREFIX }/"
+
+    # TODO: we can set a different amount of GPUs per type of job
```
peteski22, line 32: ❓ should this `TODO` be removed?

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -69,105 +92,25 @@ def _get_results_s3_key(self, job_id: UUID) -> str:
...
+    def _get_job_params(self, job_type: str, record, request: BaseModel) -> dict:
         # get dataset S3 path from UUID
         dataset_s3_path = self.data_service.get_dataset_s3_path(request.dataset)
 
-        # set storage path
-        storage_path = f"s3://{ Path(settings.S3_BUCKET) / settings.S3_JOB_RESULTS_PREFIX }/"
-
-        # fill up model url with default openai url
+        # TODO: the following is currently both eval and inference, but
+        #       will soon be inference only. Make sure we'll move this out
```
peteski22, line 113: ❓ same question for this `TODO` 😄 could we create an issue to track the required work if we need to keep note of this?

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -179,27 +122,60 @@ def create_evaluation_job(self, request: JobCreate) -> JobResponse:
...
+                "temperature": request.temperature,
+                "top_p": request.top_p,
+            }
+
+        return job_params
+
+    def create_job(self, request: JobEvalCreate | JobInferenceCreate) -> JobResponse:
+        """Creates a new evaluation workload to run on Ray and returns the response status."""
+        # TODO: we might want to explicitly provide a job string definition as a request
+        # parameter, so we can reuse the same config profiles across different job types
```
peteski22, line 159: Sorry, I'm just a broken record of unhelpful comments 😭 

same question as: https://github.com/mozilla-ai/lumigator/pull/328/files#r1842617684

lumigator/python/mzai/backend/backend/services/jobs.py
aittalam: I'd keep this up until we make a decision whether we want to set different amount of GPUs per type of job

lumigator/python/mzai/backend/backend/services/jobs.py
aittalam: No need, this is gonna be part of my next PR after Vicki's code is merged too

lumigator/python/mzai/jobs/inference/tests/test_configs.py
```diff
@@ -0,0 +1,63 @@
...
+        # make a copy of the dict
+        d_copy = cfgd.copy()
+
+        if isinstance(cfgd[k], dict):
+            # iterate depth-first on its keys
+            for d_child in configs_with_missing_field(cfgd[k]):
+                d_copy[k] = d_child
+                yield d_copy
+
+        # after removing each son, finally remove the parent
```
peteski22, line 19:
```suggestion
        # after removing each child, finally remove the parent
```

🤷🏼

PR #329, opened 2024-10-31
LUMI-127 SDK integration test fix

javiermtorres:
## What's changing

This draft PR fixes the SDK integration tests.

## How to test it

Just run `make test`.

N/A

## I already...

  - Waiver: new documentation is not needed for this fix


lumigator/python/mzai/sdk/tests/data/job-all.json
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "test-job-001",
+    "description": "This is a test job",
+    "model": "some_model",
+    "dataset": "6f6487ac-7170-4a11-af7a-0f6db1ec9a74",
+    "max_samples": 5,
+    "model_url": "s3://some_url",
+    "system_prompt": "some prompt",
+    "config_infer_template": "some_infer_template",
```
aittalam, line 9: I am not sure this should work, we are not using `config_*_template` and especially not both inference and eval together now

aittalam:
Hi Javier!
Thank you first of all for your PR, it really seems to fix the job integration tests.

I see however that there are way more changes than I expected, including tests on other endpoints (dataset, health), changes in the SDK client, integration tests ran separately (and accounted for in the Makefile).

I do not feel comfortable merging everything into my PR as its only scope should be adding the inference job endpoint (and anything strictly related to that). I see a few issues here, including that others would need for my PR to be merged to have more general SDK tests fixed which I think would complicate things.

Ideally I'd suggest making a smaller, actual (ie. not draft) PR to main to fix the other things, and a more specific one related to inference jobs only which I'd be happy to merge with mine. I'd be happy to rebase my code to be consistent with the former, and merge the latter to make sure tests will still work after my commit. WDYT?

javiermtorres:
IIRC github will merge main to a PR to perform the checks (if such a merge/rebase is not possible, for example if a manual merge is needed, AFAICT the tests will not run). So it would not be your PR failing, but actually your PR _with main merged_. That's why I couldn't see any issues in the code itself. Any changes outside the scope of the SDK have been added as the result of the merge to main.

If you can run tests ok with `uv run pytest -o "python_files=int_test_*.py"` in the sdk directory after starting the lumigator compose (e.g. via make start-lumigator-build), then the issue lies in the merge to main. In that case, since the tests will be ok, don't bother about the checks (because you're not concerned with merging to main right now). If you'd like to merge to main, leave your branch in a state that can be fast-forwarded into main, and then I'll update this PR; you'll probably see just changes in the SDK.

Let's talk next week anyway ;)

javiermtorres:
PR now unnecessary

PR #330, opened 2024-10-31
Comment out default region

chainlink:
## I already...


agpituk (commented):
Just tested it with CW and localstack and definitely we don't need a default region, good finding!. Wouldn't it be better to remove the line instead of commenting it? As it's not necessary for local deployment with localstack. (Vars for external s3 storage will still be in the `docker-compose-external.yaml`)

chainlink:
I left it there commented for reference (ie, you'll need it when you use AWS)

PR #332, opened 2024-10-31
Datasets: deleting a dataset with a DB entry but no S3 file should be handled correctly

peteski22:
## What's changing

When attempting to delete a dataset which encounters issues, we now do the following:

* If the dataset existed in the DB but the file wasn't found in S3, we remove it from the DB
* Any other errors in the service will raise an exception (non-HTTP)
* The route/controller handles any exceptions, logging them and returning a 500 to the caller (with a generic message)

## I already...

### DIFF
--- a/lumigator/python/mzai/backend/backend/api/routes/datasets.py
+++ b/lumigator/python/mzai/backend/backend/api/routes/datasets.py
@@ -1,7 +1,8 @@
 from typing import Annotated
 from uuid import UUID
 
-from fastapi import APIRouter, Form, UploadFile, status
+import loguru
+from fastapi import APIRouter, Form, HTTPException, UploadFile, status
 from schemas.datasets import DatasetDownloadResponse, DatasetFormat, DatasetResponse
 from schemas.extras import ListingResponse
 
@@ -26,7 +27,15 @@ def get_dataset(service: DatasetServiceDep, dataset_id: UUID) -> DatasetResponse
 
 @router.delete("/{dataset_id}", status_code=status.HTTP_204_NO_CONTENT)
 def delete_dataset(service: DatasetServiceDep, dataset_id: UUID) -> None:
-    service.delete_dataset(dataset_id)
+    try:
+        service.delete_dataset(dataset_id)
+    except Exception as e:
+        loguru.logger.error(f"Unexpected error deleting dataset ID from DB and S3: {dataset_id}. "
+                            f"{e}")
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail=f"Unexpected error deleting data for ID: {dataset_id}",
+        ) from e
 
 
 @router.get("/")
--- a/lumigator/python/mzai/backend/backend/api/routes/datasets.py
+++ b/lumigator/python/mzai/backend/backend/api/routes/datasets.py
@@ -1,8 +1,8 @@
 from typing import Annotated
 from uuid import UUID
 
-import loguru
 from fastapi import APIRouter, Form, HTTPException, UploadFile, status
+from loguru import logger
 from schemas.datasets import DatasetDownloadResponse, DatasetFormat, DatasetResponse
 from schemas.extras import ListingResponse
 
@@ -30,7 +30,7 @@ def delete_dataset(service: DatasetServiceDep, dataset_id: UUID) -> None:
     try:
         service.delete_dataset(dataset_id)
     except Exception as e:
-        loguru.logger.error(f"Unexpected error deleting dataset ID from DB and S3: {dataset_id}. "
+        logger.error(f"Unexpected error deleting dataset ID from DB and S3: {dataset_id}. "
                             f"{e}")
         raise HTTPException(
             status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
--- a/lumigator/python/mzai/backend/backend/api/routes/datasets.py
+++ b/lumigator/python/mzai/backend/backend/api/routes/datasets.py
@@ -22,7 +22,15 @@ def upload_dataset(
 
 @router.get("/{dataset_id}")
 def get_dataset(service: DatasetServiceDep, dataset_id: UUID) -> DatasetResponse:
-    return service.get_dataset(dataset_id)
+    dataset = service.get_dataset(dataset_id)
+
+    if not dataset:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"Dataset not found: {dataset_id}",
+        )
+
+    return dataset
 
 
 @router.delete("/{dataset_id}", status_code=status.HTTP_204_NO_CONTENT)
--- a/lumigator/python/mzai/backend/backend/api/routes/datasets.py
+++ b/lumigator/python/mzai/backend/backend/api/routes/datasets.py
@@ -34,7 +34,7 @@ def delete_dataset(service: DatasetServiceDep, dataset_id: UUID) -> None:
                             f"{e}")
         raise HTTPException(
             status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
-            detail=f"Unexpected error deleting data for ID: {dataset_id}",
+            detail=f"Unexpected error deleting dataset for ID: {dataset_id}",
         ) from e
 
 
--- a/lumigator/python/mzai/backend/backend/api/routes/datasets.py
+++ b/lumigator/python/mzai/backend/backend/api/routes/datasets.py
@@ -23,11 +23,10 @@ def upload_dataset(
 @router.get("/{dataset_id}")
 def get_dataset(service: DatasetServiceDep, dataset_id: UUID) -> DatasetResponse:
     dataset = service.get_dataset(dataset_id)
-
     if not dataset:
         raise HTTPException(
             status_code=status.HTTP_404_NOT_FOUND,
-            detail=f"Dataset not found: {dataset_id}",
+            detail=f"Dataset '{dataset_id}' not found.",
         )
 
     return dataset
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -170,3 +170,4 @@ cython_debug/
 
 .ijwb
 *.db
+*.db-journal
```
peteski22, line 173: This one was just something I noticed pop up that we don't want in general.

aittalam: Out of curiosity, is this something from sqlite and when is that generated? (I never noticed that!)
Thanks for catching it btw, I think it makes sense not to commit it

lumigator/python/mzai/backend/backend/services/datasets.py
```diff
@@ -168,20 +167,33 @@ def get_dataset_s3_path(self, dataset_id: UUID) -> str:
...
-            # Delete from S3
-            # S3 delete is called first => if this fails the DB delete won't take place
+            # S3 delete is called first, if this fails for any other reason that the file not being
+            # found, then the DB delete won't take place, and an exception raised.
             dataset_key = self._get_s3_key(record.id, record.filename)
             dataset_path = f"s3://{ Path(settings.S3_BUCKET) / dataset_key }"
             self.s3_filesystem.rm(dataset_path, recursive=True)
+        except FileNotFoundError as e:
+            # File not found errors are allowed, but we perform clean-up in this situation.
+            logger.error(f"Dataset ID: {dataset_id} was present in the DB but not found on S3... "
```
aittalam, line 191: Being non-breaking, I'd keep this as a logger.info or a logger.warning

lumigator/python/mzai/backend/backend/services/datasets.py
```diff
@@ -168,20 +167,33 @@ def get_dataset_s3_path(self, dataset_id: UUID) -> str:
...
+                         f"Cleaning up DB by removing ID. {e}")
 
-            # Delete DB record
-            self.dataset_repo.delete(record.id)
+        # Getting this far means we are OK to remove the record from the DB.
+        self.dataset_repo.delete(record.id)
```
aittalam, line 195: This part of code is not reachable if the previous exception is caught, so we should have a call to `self.dataset_repo.delete` inside the `except FileNotFoundError` too

lumigator/python/mzai/backend/backend/services/datasets.py
```diff
@@ -168,20 +167,33 @@ def get_dataset_s3_path(self, dataset_id: UUID) -> str:
...
-            # S3 delete is called first => if this fails the DB delete won't take place
+            # S3 delete is called first, if this fails for any other reason that the file not being
+            # found, then the DB delete won't take place, and an exception raised.
             dataset_key = self._get_s3_key(record.id, record.filename)
             dataset_path = f"s3://{ Path(settings.S3_BUCKET) / dataset_key }"
             self.s3_filesystem.rm(dataset_path, recursive=True)
+        except FileNotFoundError as e:
+            # File not found errors are allowed, but we perform clean-up in this situation.
+            logger.error(f"Dataset ID: {dataset_id} was present in the DB but not found on S3... "
+                         f"Cleaning up DB by removing ID. {e}")
```
aittalam, line 192: being this a trapped exception for an error we knew did not give us useful information, do we need to get and print `e` here?

lumigator/python/mzai/backend/backend/tests/api/routes/test_datasets.py
```diff
@@ -83,6 +85,17 @@ def test_upload_delete(app_client: TestClient, valid_experiment_dataset: str):
     assert get_response.status_code == status.HTTP_404_NOT_FOUND
 
 
+def test_dataset_delete_error(app_client: TestClient):
```
aittalam, line 88: This is great as it can replicate the issue we had with unhandled exceptions, thanks!

Can we add something that tests the fix we have just applied? I'd expect that if we tried to delete a dataset that only appears in the DB and not on S3 we'd get a 204

aittalam (changes requested):
Hey, many thanks for the fix!

A few notes / questions:

- about idempotency, I agree with [this](https://stackoverflow.com/questions/6439416/status-code-when-deleting-a-resource-using-http-delete-for-the-second-time/60695301#60695301) post about the fact that what is idempotent is the result (i.e. if you ask to delete the same resource twice, the second time nothing changes in the system) but not necessarily the response, and returning 404 the second time is more sound (as you'd then avoid an inconsistency having a 404 when you try to get the resource and a 204 when you try to delete it)

- re: moving throwing the proper exceptions at the controller level, I like the idea as it makes code more readable (ie. there is one single place where you can see the main flow rather than having to delve deeper in code). Two points: (1) if we choose to do that we should do it consistently across our code (i.e. not just in datasets but also in jobs, experiments, etc), and (2) if we choose to remove exception throwing we should also remove the related private methods for the service classes (e.g. `_raise_not_found`, `_raise_unhandled_exception` inside `DatasetService`).

I think the first point can be addressed easily (as long as we agree on the semantics of our API, for which I think it'd be nice to have a chat). For the latter, I'd probably make a separate PR to make sure we address the problem (which is a bit wider than this bugfix) in a consistent way. WDYT?

.gitignore
peteski22: I'm not sure, but I guess so as that's all I've been testing with 🤔 ... odd.

lumigator/python/mzai/backend/backend/services/datasets.py
peteski22: I was thinking it would be useful to emit everything we had into the logs in case someone wants to go looking around for more information. 

Happy to drop/swallow it if you think it's not worthwhile 😄

lumigator/python/mzai/backend/backend/services/datasets.py
peteski22: I'm not sure that's the case, as long as the exception is the `FileNotFoundError` we will carry on until the end of the function after the except cases are run, as the except doesn't re-raise anything and change the flow of execution. 

Here's an example that can be run in the REPL:

```python
def foo():
    try:
        print("doing something")
        raise Exeption("sadness")
    except Exception as e:
        print(f"handled an exception: {e}")
        
    print("still ran this")


foo()
```

lumigator/python/mzai/backend/backend/services/datasets.py
peteski22: https://github.com/mozilla-ai/lumigator/pull/332/commits/91169a6e368869b617dc5cde14a4ab61b1b47626

.gitignore
peteski22: Just did a quick check online to satisfy myself, looks like it usually creates the file during transactions, but cleans up when done. So maybe it got created as something was killed off half way through (allowing it to sort things out when it came back up).

lumigator/python/mzai/backend/backend/services/datasets.py
aittalam: I remember s3fs was quite minimal in the info it returned, but I think it's worth keeping it if we see, after testing, that there's anything useful in there

lumigator/python/mzai/backend/backend/services/datasets.py
aittalam: oh my bad, we are not raising any other exception from there! Sorry about that, and many thanks for double checking 🙏

peteski22:
@aittalam thanks for the review! 👍🏼 

> I think the first point can be addressed easily (as long as we agree on the semantics of our API, for which I think it'd be nice to have a chat). For the latter, I'd probably make a separate PR to make sure we address the problem (which is a bit wider than this bugfix) in a consistent way. WDYT?

In reverse order 😉:

2: Yes, good shout. I agree we should move everything over and ditch the service's methods that aren't used, I had thought about going for it here in this PR, but as you said too, it's a bit much. I think it might make sense as a series of small PRs to move things into place, and then a final one that just centralises the duplicate strings etc. to funcs.

1: Yeah, it's a tough one, that post has a good discussion on the pros/cons for both sides 🤔 I saw somewhere in the comments a link to *another* [SO post with an answer](https://stackoverflow.com/a/45194747/57591) which was where my head was at on the topic.  I guess whatever we do, we should be consistent across the API.

lumigator/python/mzai/backend/backend/services/datasets.py
peteski22: After pairing up on this we agreed to leave the exception in the log as it may be useful in diagnosing any discrepancies between S3 paths etc.

lumigator/python/mzai/backend/backend/tests/api/routes/test_datasets.py
peteski22: https://github.com/mozilla-ai/lumigator/pull/332/commits/b0c4d317ac77a99b3d1deb64bd22d220893da90f#diff-b92c2f9bf8c3f6c6a676575aea579fea1156b021c695ae70fcd005b1ef36a99d

lumigator/python/mzai/backend/backend/api/routes/datasets.py
peteski22, line 1: Note for reviewers: in a future PR we will remove HTTP related exceptions from the service, they've been included for now to follow the existing pattern.

lumigator/python/mzai/backend/backend/tests/services/test_dataset_service.py
```diff
@@ -0,0 +1,20 @@
+from s3fs import S3FileSystem
+
+from backend.repositories.datasets import DatasetRepository
+from backend.services.datasets import DatasetService
+
+
+def test_delete_dataset_file_not_found(db_session, s3_client):
+    filename = "juan"
```
aittalam, line 8:
```suggestion
    filename = "dataset.csv"
```

aittalam (approved):
TY Peter! Checked together and tested in the branch. Still a minor change 🙇  then ok to merge 🙏

PR #333, opened 2024-10-31
Factoring evaluator out into its own job independent of inference and shadowing evaluator with eval_lite

veekaybee:
## What's changing

This PR shadows https://github.com/mozilla-ai/lumigator/pull/328 and factors out the evaluation job into its own job repo, with its own requirements, Pydantic config, and run entrypoint. 

In order to shadow the current evaluation, we keep it for now and create a branch to remove the evaluator once eval_lite works. 

## I already...

- [X] added some tests for any new functionality - added tests at the job submission level and changed current tests
- [X] updated the documentation


docker-compose.yaml
```diff
@@ -106,8 +106,8 @@ services:
       - AWS_ENDPOINT_URL=http://localhost:4566
       - S3_BUCKET=lumigator-storage
       - PYTHONPATH=/mzai/lumigator/python/mzai/backend
-      - EVALUATOR_PIP_REQS=/mzai/lumigator/python/mzai/jobs/evaluator/requirements.txt
-      - EVALUATOR_WORK_DIR=/mzai/lumigator/python/mzai/jobs/evaluator
+      - EVALUATOR_PIP_REQS=/mzai/lumigator/python/mzai/jobs/eval_lite/requirements.txt
```
veekaybee, line 109: @aittalam fyi heads up here's the scope of work I've been doing that parallels yours. Not ready yet, just an idea of where I'm headed

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -25,7 +26,7 @@ def create_inference_job(
 @router.post("/evaluate/", status_code=status.HTTP_201_CREATED)
 def create_evaluation_job(
     service: JobServiceDep,
-    request: JobCreate,
+    request: JobEvalCreate,
```
veekaybee, line 29: @aittalam fyi heads up here's the scope of work I've been doing that parallels yours. Not ready yet, just an idea of where I'm headed

lumigator/python/mzai/backend/backend/services/datasets.py
```diff
@@ -18,7 +18,7 @@
 from backend.settings import settings
 
 GT_FIELD: str = "ground_truth"
-REQUIRED_EXPERIMENT_FIELDS: set[str] = {"examples"}
+REQUIRED_EXPERIMENT_FIELDS: set[str] = {"predictions"}
```
veekaybee, line 21: fyi @aittalam

aittalam: Is this for a simple dataset upload? I think examples (or we should more generally call them "samples" at some point) are the only strict requirement for _uploading_ a dataset, while I agree that both predictions + gt are needed to run the evaluation

docker-compose.yaml
aittalam: Will we have a period in which both the old and the new evals run simultaneously (e.g. for testing one against the other)? If so, perhaps we could add a specific workdir/requirements pair for eval_lite instead of just changing the existing one.

lumigator/python/mzai/backend/backend/services/datasets.py
veekaybee: yup, I was under the impression that we're changing the baseline field name to predictions as well, but samples makes a lot more sense. 

What should we have for all the field names just to get it down on "paper"? The way I had in my mental model after correcting for samples: 

You're uploading any given dataset: 
`samples` <-- your input data (perhaps `input_samples` in that case?)

You have your own ground truth: 
`samples` `ground_truth`

You've performed ground truth inference: 
`samples` `ground_truth` `prediction` (the generated gt)

Does that make sense?

lumigator/python/mzai/jobs/eval_lite/eval_lite_config.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class DatasetConfig(BaseModel):
+    path: str
+    model_config = ConfigDict(extra="forbid")
+
+class JobConfig(BaseModel):
+    max_samples: int = 0
+    storage_path: str
+    input_field: str = "prediction"
```
aittalam, line 11: I love that you provide that as a parameter!
Shall we have one for gt too, so we do not assume anything from the input dataset?

lumigator/.DS_Store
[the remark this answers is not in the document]
veekaybee: oh boy! 🤦‍♀️

lumigator/python/mzai/jobs/eval_lite/eval_lite.py
```diff
@@ -0,0 +1,159 @@
...
+
+    # add input data to results dict
+    if config.get("evaluation").get("return_input_data"):
+        evaluation_results["predictions"] = dataset["predictions"]
+        evaluation_results["ground_truth"] = dataset["ground_truth"]
+
+    output_path = save_outputs(config, evaluation_results)
+    return output_path
+
+@click.command()
```
aittalam, line 150: +1 for click here, I'll bring it into inference too

veekaybee:
@aittalam Note that I don't have the pydantic types wired up 100% yet but I have a placeholder file for them following what we'd discussed + the way it's structured in your PR

lumigator/python/mzai/jobs/eval_lite/requirements.txt
aittalam, line 1: I think there might be a few more dependencies than those actually required by the new lite job (unless they are indirect depedencies I did not expect from hf_evaluate). Fine to keep these in as long as you are testing, then I'd suggest to try and reduce them to minimize the job's fingerprint

docker-compose.yaml
veekaybee: WDYT makes more sense? I was thinking about it as: 
1. This PR gets approved, which works with eval lite
2. I cut a branch off this PR which approves eval
3. Both are merged together which means new eval runs off of eval-lite that's renamed. 

I was hoping to avoid having both at once and get rid of the old evaluator code but I'm not sure what the global impact will be

lumigator/python/mzai/jobs/evaluator/requirements.txt
veekaybee, line 1: ideally this whole thing could be removed in a second pass

lumigator/python/mzai/backend/backend/services/datasets.py
aittalam: ah I guess that's just for being able to test with a dataset with predictions + gt? Please ignore me in that case :-)

docker-compose.yaml
aittalam: Yeah, one thing I am a bit afraid of is the need for both inference and lite_eval to be up and running PLUS the extra code to deal with the multi-step process before being able to substitute current evaluation. This would make it a bit harder for us to do things incrementally, especially given the quickly-moving codebase.

lumigator/python/mzai/backend/backend/ray_submit/submission.py
```diff
@@ -33,9 +33,9 @@ def command_with_params(self) -> str:
 
 
 def submit_ray_job(client: JobSubmissionClient, entrypoint: RayJobEntrypoint) -> str:
-    loguru.logger.info(f"Submitting {entrypoint.command_with_params}...{entrypoint.runtime_env}")
+    loguru.logger.info(f"Run {entrypoint.get_command_with_params}\n...{entrypoint.runtime_env}")
```
veekaybee, line 36: Added the \n because it was hard to see both vars expanded in the logs

lumigator/python/mzai/backend/backend/ray_submit/submission.py
```diff
@@ -16,7 +16,7 @@ class RayJobEntrypoint(ABC):
     memory: int | float | None = None
 
     @property
-    def command_with_params(self) -> str:
+    def get_command_with_params(self) -> str:
```
veekaybee, line 19: changing this to a verb since we generally want methods to be verbs and variables nouns

lumigator/python/mzai/jobs/eval_lite/eval_lite.py
```diff
@@ -0,0 +1,159 @@
...
+from pathlib import Path
+
+import click
+import numpy as np
+import s3fs
+from datasets import load_from_disk
+from loguru import logger
+import evaluate
+
+class EvaluationMetrics:
```
veekaybee, line 11: @aittalam wdyt about stuffing this all into one script for now? It generally feels weird but it's small enough that I think it's ok since we want to send a minimal amount of files to Ray

lumigator/python/mzai/backend/backend/services/datasets.py
aittalam: Yep! I'd keep `samples` for simplicity - and then we should make sure things stay consistent across all tests / docs

veekaybee: ok so baseline input is `samples` for `REQUIRED_EXPERIMENT_FIELDS` and then we extend as needed for jobs. makes sense to me!

aittalam: perfect, TY!

docker-compose.yaml
ividal: Let's discuss this particular piece in the context of experiments and where we're going with the jobs 😄

veekaybee: keeping it separate for now

veekaybee:
Closing out this PR in favor of the work in this branch https://github.com/mozilla-ai/lumigator/compare/eval_lite?expand=1

See more detail here: https://github.com/mozilla-ai/lumigator/issues/435

PR #335, opened 2024-11-01
Added Github Action Check for Broken Links

njbrake:
## What's changing

Please totally feel free to throw this away if you find it to be unnecessary overhead for the project. 😄  When I was browsing the repo I clicked a few broken links so I thought this might help prevent links from accidentally being broken when changes are made.
 
Added a Github Action to check and flag broken links in markdown: The summary page of the Github action will show a summary and print out which links are broken.  This is the action: https://github.com/marketplace/actions/lychee-broken-link-checker

As the project gets reorganized this may be helpful to ensure that all links remain intact.

Screenshot 2024-11-01 at 4 03 37 PM

## How to test it
In the Github Actions page, click on the workflow name and then "run workflow" and you'll see it flag a variety of broken links. 

I only included .md files but this could be expanded to include other file types if you wanted. I excluded all localhost addresses and also any 'mailto' that might show up, but we could also exclude addresses outside the repository, if you only wanted to check links that refer to something else in the repo.

## I already...

Ran it on my fork to verify that it checks and finds the broken links. I didn't yet fix the broken links that it found

- [ x] added some tests for any new functionality
- [ x] checked if a (backend) DB migration step was required and included it if required

### DIFF
--- a/.github/workflows/broken_links_checker.yaml
+++ b/.github/workflows/broken_links_checker.yaml
@@ -1,0 +1,18 @@
+name: Check for Broken Links
+
+on:
+    workflow_dispatch:
+    push:
+        branches:
+            - '*'
+
+jobs:
+  linkChecker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      # Thanks https://github.com/marketplace/actions/lychee-broken-link-checker!
+      - name: Link Checker
+        id: lychee
+        uses: lycheeverse/lychee-action@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/broken_links_checker.yaml
```diff
@@ -0,0 +1,21 @@
+name: Check for Broken Links in Markdown
+
+on:
+    workflow_dispatch:
+    push:
+        branches:
+            - '*'
```
njbrake, line 7: Probably overkill to run on every push, but may be worth changing this to run on any PR that is trying to merge to main. 🤷

dpoulopoulos:
Hey @njbrake, 

Thanks for this! We now have a new [docs site](https://mozilla-ai.github.io/lumigator/) where we'll continue to add more content. As a result, we'll be keeping the README files much lighter going forward. 

I'm closing this PR for now.

njbrake:
Ooh great! Thanks for the info 👍🏼

PR #345, opened 2024-11-11
Update Notebooks to reflect current API use

dpoulopoulos:
## What's changing

Due to recent changes to the API, the Jupyter Notebook used for onboarding users to Lumigator is no longer functional. 

* Update the notebook to align with the intended user experience.
* Remove the ground truth generation sections because they are not currently supported by the API and the SDK.
* Add an integration test to check whether the Notebook runs to completion.

## How to test it

1. Clone the repository.
2. Navigate into the `notebooks` directory: `cd lumigator/notebooks`
3. Follow the README instructions.

## I already...


peteski22 (commented):
Nice one @dpoulopoulos! 

I think I caught two small typos, so I added suggestions, feel free to accept them if you agree.

.github/workflows/notebook_integration_tests.yaml
```diff
@@ -0,0 +1,46 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v3
+
+      - name: Create a virtual environment
+        run: uv venv
+        working-directory: notebooks
```
veekaybee, line 26: we might need to specify the full working-dir path here: https://github.com/mozilla-ai/lumigator/blob/b2d760b13b6b8569c0a80e1252528e0b62edc7a6/.github/workflows/sdk_integration_tests.yaml#L32C28-L32C50

notebooks/README.md
```diff
@@ -1,14 +1,26 @@
...
-1. create a new virtualenv for lumigator
-2. `cd notebooks` (assuming you were in the root directory of the repo)
-3. `pip install -r requirements.txt`
-4. `pip install jupyterlab` 
-5. set up the environment so that both our backend and our ray cluster point to localhost and start jupyterlab:
-```LUMIGATOR_SERVICE_HOST="localhost" RAYCLUSTER_KUBERAY_HEAD_SVC_PORT_8265_TCP_ADDR="localhost" jupyter-lab```
+The notebook runs on Jupyter. If you don't yet have Jupyter set up:
 
-A new browser window will open pointing at your new jupyterlab. From there, open the `walkthrough.ipynb` file.
+1. Create a new virtualenv for lumigator: `python3 -m venv <path-to-virtualenv>`.
```
veekaybee, line 8: Let's use `uv` here for venv creation to keep consistent with the rest of the app

dpoulopoulos: @veekaybee my assumption is that the user may not have `uv` installed. But I can add it, for sure. WDYT?

veekaybee: yep, would be helpful to specify both

notebooks/utils.py
```diff
@@ -0,0 +1,97 @@
+import json
+from typing import List
```
veekaybee, line 2: let's use `list` instead of List here since we're on Python 3.11 https://peps.python.org/pep-0585/

.github/workflows/notebook_integration_tests.yaml
dpoulopoulos: @veekaybee I don't follow. I'm creating the venv in `/notebooks` and installing the `requirements.txt` that exists in this directory. What is the full working-dir path you're referring to?

veekaybee: Does this step (and others) need to know about the fully-qualified path here, aka `lumigator/notebooks`?

notebooks/utils.py
```diff
@@ -0,0 +1,97 @@
...
+    "BERT Precision": ["bertscore", "precision_mean"],
+    "BERT Recall": ["bertscore", "recall_mean"],
+    "BERT F1": ["bertscore", "f1_mean"],
+    "ROUGE-1": ["rouge", "rouge1_mean"],
+    "ROUGE-2": ["rouge", "rouge2_mean"],
+    "ROUGE-L": ["rouge", "rougeL_mean"],
+}
+
+
+def _parse_model_results(result: dict, mi: pd.DataFrame):
```
veekaybee, line 20: could we add a more descriptive var for `mi` (i.e. are these model inferences?)

notebooks/utils.py
```diff
@@ -0,0 +1,97 @@
...
+        row[column] = value
+
+    row["RAM_GB"] = mi[mi.model_name == model_name]["RAM_GB"].to_numpy()[0]
+
+    return row
+
+
+def _download_text_file(result: JobResultDownloadResponse) -> str:
+    download_url = result.download_url
+    download_response = requests.get(download_url)
```
veekaybee, line 43: I think we can do this directly via the SDK now rather than requests https://github.com/mozilla-ai/lumigator/blob/b2d760b13b6b8569c0a80e1252528e0b62edc7a6/lumigator/python/mzai/sdk/sdk/jobs.py#L175

notebooks/utils.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    exp_results = _download_text_file(result)
+    return json.loads(exp_results)
+
+
+def results_to_table(results: list[dict]) -> pd.DataFrame:
+    """Visualize results as a table.
+
+    Args:
+        results (List[dict]): The content of job results.
```
veekaybee, line 65: same, typing can be `list`

.github/workflows/notebook_integration_tests.yaml
dpoulopoulos: I don't think so. I have built the venv in /`notebooks` that's why I'm invoking every other step from this directory. For example:

```
      - name: Install sdk
        run: uv pip install -e ../lumigator/python/mzai/sdk
        working-directory: notebooks
```

notebooks/utils.py
```diff
@@ -0,0 +1,97 @@
...
+
+    Returns:
+        pd.DataFrame: The result table.
+    """
+    evaluation_results = []
+
+    mi = pd.read_csv(MODEL_INFO_FILE)
```
veekaybee, line 72: suggestion: model_info or similar

notebooks/utils.py
dpoulopoulos: `model_info` but yes, you're right. I'll change the var name.

notebooks/walkthrough.ipynb
veekaybee, line 1: Notebooks are super hard to review in GH still, a couple comments: 

+ lm_client = LumigatorClient(f"{os.getenv('LUMIGATOR_SERVICE_HOST', 'localhost')}:8000") - we can put the port as a global variable in the notebook, or in the config for reference
+ Do we show what the eval dict looks like as a python object? I think we do here `result.keys()`, if not it would be helpful to see the whole thing so users know what they're traversing

notebooks/utils.py
dpoulopoulos: The SDK returns the download URL. This is what I'm using in the Notebook. This function retrieves the actual results.

veekaybee (approved):
LGTM once final comments are addressed, thanks for this!

notebooks/README.md
dpoulopoulos: I added both commands, how you can use `uv` to create a virtual environment and how you can do it if you don't have `uv` installed.

notebooks/walkthrough.ipynb
dpoulopoulos: Yes we do. And we also have this `get_nested_value` function that can retrieve any value from the dict. For instance, we use it to retrieve the `summarzation_time`, `bertscore` and `bertscore/precision_mean`.

Issue #343, opened 2024-11-08
CI fails with "username and password required" error.


aittalam:
**Describe the bug**
[This PR](https://github.com/mozilla-ai/lumigator/pull/321) and [this PR ](https://github.com/mozilla-ai/lumigator/pull/335) (both coming from forks) fail on CI with a "username and password required" error (respectively [here](https://github.com/mozilla-ai/lumigator/actions/runs/11745660850/job/32724150542?pr=321) and [here](https://github.com/mozilla-ai/lumigator/actions/runs/11635883764/job/32406163535?pr=335)). Is it something we should fix in our actions?

**To Reproduce**
Open one of the two PRs and try re-running checks

**Expected behavior**
Checks work even from forked repos

agpituk:
Hi @aittalam  , should we close this one after merging https://github.com/mozilla-ai/lumigator/pull/346?

aittalam:
Yes, we tested this live and it works properly. Many thanks @agpituk !

PR #346, opened 2024-11-11
Trigger push to dockerhub only when merging PRs

agpituk:
## What's changing
We currently have an [issue](https://github.com/mozilla-ai/lumigator/issues/343) when external users want to open a PR against our code. This is related to [accessing](https://michaelheap.com/access-secrets-from-forks/) secrets from forks. The proposed solution is to only login & push to dockerhub when we actually merge to main (as I don't think we need the images in dockerhub if they come from a PR)

We've also removed the linting process from the PR/push to main as it's already triggered every time we [push](https://github.com/mozilla-ai/lumigator/blob/main/.github/workflows/tests_uv.yaml)

## How to test it
Ensure CI works properly and after merge we can fork the repo and try to see if the error still happens
## I already...

- [N/A] updated the documentation
- [N/A] checked if a (backend) DB migration step was required and included it if required

### DIFF
--- a/.github/workflows/build_images_uv.yaml
+++ b/.github/workflows/build_images_uv.yaml
@@ -32,15 +32,16 @@ jobs:
       - name: Get repo version
         run: |
           echo "REPOVERSION=$(git describe --tags --dirty --match \"[0-9\.]*\" --always)" >> $GITHUB_ENV
-          echo "Version of the repo for this build: $REPOVERSION"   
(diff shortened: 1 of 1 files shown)


agpituk:
When doing this, @macaab26 pointed out that we're building our images in all of our workflows which is quite inefficient, so we are going to open a new ticket to address that in a different PR

macaab26 (approved):

PR #347, opened 2024-11-11
201 Create - adds Location

peteski22:
## What's changing

As per [RFC9110](https://www.rfc-editor.org/rfc/rfc9110#name-201-created) and [MDN web docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201) an HTTP `201 - Created` should include the URL/URI of the created resource in the `Location` header of the response. 

The following tags have endpoints which return `201`:

* Datasets
* Jobs

image

image

This PR ensures that the endpoints return a populated `Location` header in addition to their JSON response body.

## How to test it

**Example: Create dataset**

Export `$DATA_CSV_PATH` to the path of your dataset file, then `make local-up` and execute the curl command:

```
curl -v http://localhost:8000/api/v1/datasets/ \ 
  -F 'format=job'
```

The response will contain the body and also the headers, `location` will be present (see additional notes below)

Sample response for creating a dataset, via curl on Docker hosted local deployment:

```
< HTTP/1.1 201 Created
< date: Mon, 11 Nov 2024 14:20:50 GMT
< server: uvicorn
< content-length: 166
< content-type: application/json
< location: http://localhost:8000/api/v1/datasets/3f2c31bb-35c1-47de-81fc-a9c72a513cb0
```

## I already...

### DIFF
--- a/.github/workflows/build_images_uv.yaml
+++ b/.github/workflows/build_images_uv.yaml
@@ -22,25 +22,19 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Install uv
-        uses: astral-sh/setup-uv@v3
-
-      - name: Lint
-        run: uvx ruff check --config ruff.toml
-        continue-on-error: false
-
       - name: Get repo version
         run: |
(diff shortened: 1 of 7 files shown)


aittalam (approved):
LGTM, thanks Peter!

PR #348, opened 2024-11-11
Setting default max samples to -1 (no limit), encapsulation and testing

veekaybee:
## I already...


veekaybee:
Addressed with https://github.com/mozilla-ai/lumigator/pull/348

PR #355, opened 2024-11-13
Enable usage inside GitHub Codespaces

daavoo:
## What's changing

This P.R. adds a new file `.devcontainer/codespaces/devcontainer.json` that allows to run and test Luminator inside [GitHub Codespaces](https://github.com/features/codespaces)

## How to test it

Create a codespace from the GitHub UI using `New with options`:

Screenshot 2024-11-13 113320

Make sure to select this branch and the `Lumigator`  config:

image

By default GitHub uses `.devcontainer/devcontainer.json` when creating a codespace. Since the existing one is not meant to be used in codespaces, perhaps it could be moved to a subfolder like `.devcontainer/local/devcontainer.json`?

## I already...

Went through the full quickstart inside a codespace:

Screenshot 2024-11-13 113024
Screenshot 2024-11-13 113036


peteski22 (approved):
> By default GitHub uses .devcontainer/devcontainer.json when creating a codespace. Since the existing one is not meant to be used in codespaces, perhaps it could be moved to a subfolder like .devcontainer/local/devcontainer.json?

I think that sounds like a good idea, I think the only place that would need updating might be the actual contents of the `devcontainer.json` as it has relative references to where to find the docker 'stuff'. Do you think this should be done in a follow-up PR?

daavoo:
> I think that sounds like a good idea, I think the only place that would need updating might be the actual contents of the `devcontainer.json` as it has relative references to where to find the docker 'stuff'. Do you think this should be done in a follow-up PR?

Since I am not really using the current `.github/devcontainer.json` it'd be better to consult with the rest of the team who might be using it.

TBH, having the `codespaces` one as default is just a convenience. How to launch a codespace to test lumigator and how to select the correct `devcontainer` can be explained in the docs

daavoo:
Since it is not clear if this direction makes sense for the Lumigator and the changes are quite simple, I am closing this for now to avoid clutter

PR #357, opened 2024-11-13
Amend the repository's README files

dpoulopoulos:
## What's changing

Simplify the README files in the repository, reducing the amount of information they contain, except for the root README file.

Retain the project description and Getting Started section in the root README file, and add links to the documentation, contributing guide, and GitHub issues.

Convert several README files into dedicated documentation pages.

Add a new CI workflow to check for broken links in Markdown and reStructuredText files.

## I already...


.github/workflows/notebook_integration_tests.yaml
```diff
@@ -1,4 +1,4 @@
-name: SDK integration tests
+name: Tests
```
peteski22, line 1:
```suggestion
name: Tests
```

Why not "Notebook integration tests"?

lumigator/python/mzai/backend/backend/alembic/README.md
peteski22, line 1: ❗ General point on files that appear deleted now, would it have been easier to track the changes if we did `git mv` on these files before modifying them? 

e.g. the content of this file wasn't removed, it was moved/split up/etc. and now is `docs/source/operations-guide/alembic.md`

peteski22 (commented):
😎 looking good @dpoulopoulos, I added two small comments.

lumigator/python/mzai/backend/backend/alembic/README.md
dpoulopoulos: This is actually the right way to do it, you're right. I didn't think about it. I think it's kinda hard to make this change now, but I'll keep this in mind for future references. Are you okay with that?

.github/workflows/notebook_integration_tests.yaml
dpoulopoulos: It could be. If we have more Notebooks down the line. I'll make that change.

peteski22 (approved):

dpoulopoulos:
@peteski22 yes, let's make a separate PR for this, after we merge this one.

README.md
```diff
@@ -1,191 +1,122 @@
...
-```
-
-**Run the app locally via docker compose:**
+To uninstall Lumigator, simply run the following command:
 
 ```bash
-make local-up
-make local-logs # gets the logs from docker compose
-make local-down # shuts it down
+make stop-lumigator
```
ividal, line 106: I'd say "to stop Lumigator", since we're not uninstalling - or instruct how to truly purge anything Lumigator after stopping (rm the folder). WDYT?

README.md
```diff
@@ -1,191 +1,122 @@
...
-| causal     | gpt-4o-mini, gpt-4-turbo, gpt-3.5-turbo-0125 |                 |    X    |
-| causal     | open-mistral-7b                              |                 |    X    |
+You can run and develop Lumigator locally using Docker Compose. This creates three container
+services networked together to make up all the components of the Lumigator application:
 
+- `localstack`: Local storage for datasets that mimics S3-API compatible functionality.
+- `backend`: Lumigator’s FastAPI REST API.
+- `ray`: A Ray cluster for submitting several types of jobs.
 
-### Models for Offline Evaluation
```
ividal, line 49: Where do users find currently supported models + tasks...? I didn't see it in a (quick) look through the other readmes + the doc - are we removing this info?

ividal (approved):
Thanks for the restructuring @dpoulopoulos ! A few comments to understand where we could plug in information that users might want to see at first glance vs in the docs.

README.md
dpoulopoulos: @ividal good point. I'll rephrase it to "To stop the containers you started using Docker Compose, simply run the following command:" to make it more explicit. WDYT?

README.md
dpoulopoulos: @ividal Yes, my plan is to move this in the docs. But first, I have to find out if this information is up to date. For example, what does it mean "Models for Offline Evaluation"? Can I pick something that is not in this list? What if I want to evaluate a model for summarization that was published yesterday? Is this an exhaustive list and if yes why?

PR #361, opened 2024-11-13
GHA: Update names of workflow jobs

peteski22:
## What's changing

GitHub's UI can make it difficult to know for sure which task you're trying  to associate with a required status check as it resolves only the job name and shows no information about the scope (i.e. the workflow name it is associate with).

This PR makes the job names more specific as to which workflow they are part of.

Once the PR merges it will be easier to find and [configure the required status checks](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-status-checks-before-merging) for PRs merging to `main`.

Current branch protection rules for main only have a few checks configured, but we'd want more to be successful before allowing a merge. 

image

This way we can ensure that we don't accidentally merge with changes that wouldn't pass our linting for example.

## I already...

### DIFF
--- a/.github/workflows/tests_uv.yaml
+++ b/.github/workflows/tests_uv.yaml
@@ -1,4 +1,4 @@
-name: Lumigator integration tests
+name: Lumigator backend tests
 
 on:
   push:
--- a/.github/workflows/tests_uv.yaml
+++ b/.github/workflows/tests_uv.yaml
@@ -1,11 +1,11 @@
-name: Tests
+name: Lumigator integration tests
 
 on:
   push:
 
 jobs:
   uv-example:
-    name: python
+    name: Integration tests - Python
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/tests_uv.yaml
+++ b/.github/workflows/tests_uv.yaml
@@ -14,7 +14,7 @@ jobs:
       - name: Install uv
         uses: astral-sh/setup-uv@v3
 
-      - name: Install python
+      - name: Install Python
         run: uv python install
         working-directory: lumigator/python/mzai/backend
 
--- a/.github/workflows/tests_uv.yaml
+++ b/.github/workflows/tests_uv.yaml
@@ -4,7 +4,7 @@ on:
   push:
 
 jobs:
-  uv-example:
+  backend-integration-tests:
     name: Backend integration tests - Python
     runs-on: ubuntu-latest
 
--- a/.github/workflows/tests_uv.yaml
+++ b/.github/workflows/tests_uv.yaml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   uv-example:
-    name: Integration tests - Python
+    name: Backend integration tests - Python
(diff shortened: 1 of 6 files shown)


macaab26 (approved):

.github/workflows/tests_uv.yaml
```diff
@@ -1,11 +1,11 @@
-name: Tests
+name: Lumigator integration tests
 
 on:
   push:
 
 jobs:
   uv-example:
-    name: python
+    name: Integration tests - Python
```
javiermtorres, line 8:
```suggestion
    name: Backend tests - Python
```
And later do "Backend unit tests" and "Backend integration tests"

.github/workflows/tests_uv_sdk.yaml
```diff
@@ -5,7 +5,7 @@ on:
 
 jobs:
   uv-example:
-    name: python
+    name: SDK tests - Python
```
javiermtorres, line 8:
```suggestion
    name: SDK unit tests - Python
```

.github/workflows/tests_uv.yaml
```diff
@@ -1,11 +1,11 @@
-name: Tests
+name: Lumigator integration tests
 
 on:
   push:
 
 jobs:
   uv-example:
```
javiermtorres, line 7: I guess the name of the job should have been updated, now that I notice? Same in other workflows.

.github/workflows/tests_uv.yaml
```diff
@@ -1,11 +1,11 @@
-name: Tests
+name: Lumigator integration tests
```
javiermtorres, line 1:
```suggestion
name: Lumigator backend tests
```

.github/workflows/tests_uv_sdk.yaml
javiermtorres: And same change in line 1 above, in the name

javiermtorres (commented):
Wdyt about the name suggestions?

.github/workflows/tests_uv.yaml
peteski22: I think the way that GitHub handles the configuration of required checks (within branch protection rules), means when you change the name of the rule it can't find it anymore and kinda hangs. (Unless there's some clever way I haven't found yet).

So I think it would be better to name this one or the other (unit/integration) and then add the missing one later and extract whatever is needed from this job - meaning the job name doesn't change, and we don't need to manually fiddle the settings to keep things working.

Since we need things up and running for all the current tests to run, I'd like to name this one:
`Backend integration tests - Python`, does that sound OK to you?

javiermtorres: Fine for me 👍

.github/workflows/tests_uv.yaml
peteski22: Yeah you're right, we should update them to be accurate. 😄

PR #362, opened 2024-11-14
Test version generation with versioningit

javiermtorres:
## What's changing

Version tags for distributable packages can be set up using the repository tags instead of hardcoding them in the `pyproject.toml` files. This PR starts with the schemas repository; whatever method is decided will be applied to the SDK package as well.

The chosen package is `versioningit`, but suggestions are welcome. Some reference links to read are:
* https://github.com/jwodder/versioningit/blob/master/README.rst
* https://versioningit.readthedocs.io/en/stable/index.html

## How to test it

The following items cover the known procedures where version generation could fail:

* Go to `lumigator/python/mzai/schemas` and do `uv build`. Then check the package names in the `dist` directory.
* Go to `lumigator/python/mzai/sdk` and do `uv build`. Then check the package names in the `dist` directory.
* Start the dev container using `make local-up`. Everything should work as expected.
* Start the dev container using `make start-lumigator-build` and run the SDK integration tests. Everything should work as expected.

## I already...

  - manual test instructions added
  - not added yet at this stage
  - the issue is build-related, a db migration is not needed


javiermtorres:
Please also suggest alternative workflows (updating the pyproject.toml on CI, for example) that we could consider.

javiermtorres:
Closed as https://github.com/mozilla-ai/lumigator/pull/425 will be implemented instead.

PR #369, opened 2024-11-14
Build and Publish docs only if there are no broken links

### LINKED ISSUE #368: Build and publish docs only if there are no broken links
## Describe the Bug

Currently, we have two separate workflows related to documentation:
- A workflow that checks for broken links in Markdown or reStructuredText files.
- A workflow that builds and publishes the documentation on GitHub Pages.

However, we don't want the second workflow to run if the first one fails, as this would result in broken links being published on the documentation page.

## Expected Behavior

The second workflow should run only if the first one completes successfully.

## Additional Context

There are two ways to achieve the desired behavior:

- Include the "Link Checker" as the first job in the "Build and Publish Docs" workflow.
- Use the `workflow_run` directive to trigger the "Build and Publish Docs" workflow, conditioning its execution on the success of the "Link Checker" (see [here](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run) for reference).

dpoulopoulos:
## What's changing

Currently, we have two separate workflows related to documentation:

    A workflow that checks for broken links in Markdown or reStructuredText files.
    A workflow that builds and publishes the documentation on GitHub Pages.

However, we don't want the second workflow to run if the first one fails, as this would result in broken links being published on the documentation page.

Make the "Build and publish docs" workflow depend on the successful completion of the "Link checker" workflow to address this issue.

Closes #368


peteski22 (approved):
🙏🏼

PR #374, opened 2024-11-14
Set max samples to unlimited by default with option to override API request cycle

veekaybee:
## What's changing

We're currently not enforcing a `max_samples` value when they initially get passed into the application, and then again when they're sent via the API to Ray. 

This forces `max_samples` for both inference and eval jobs to be set if they're not once on input and once again on submission for evaluation.  Tests cover setting via the service. 

Original PR was here, overwritten by improvements/merges from inference API. (https://github.com/mozilla-ai/lumigator/compare/main...fix_max_samples)

## How to test it

Unit tests cover input at the API level. 

## I already...

### DIFF
--- a/lumigator/python/mzai/backend/backend/services/jobs.py
+++ b/lumigator/python/mzai/backend/backend/services/jobs.py
@@ -148,9 +148,6 @@ class JobService:
                 "top_p": request.top_p,
             }
 
-        # set max samples to unlimited if not specified
-        if request.max_samples is None:
-            request.max_samples = -1
 
         return job_params
 
--- a/lumigator/python/mzai/backend/backend/services/jobs.py
+++ b/lumigator/python/mzai/backend/backend/services/jobs.py
@@ -148,6 +148,10 @@ class JobService:
                 "top_p": request.top_p,
             }
 
+        # set max samples to unlimited if not specified
+        if request.max_samples is None:
+            request.max_samples = -1
+
         return job_params
 
     def create_job(self, request: JobEvalCreate | JobInferenceCreate) -> JobResponse:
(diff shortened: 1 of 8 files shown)


lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -148,6 +148,10 @@ def _get_job_params(self, job_type: str, record, request: BaseModel) -> dict:
                 "top_p": request.top_p,
             }
 
+        # set max samples to unlimited if not specified
+        if request.max_samples is None:
+            request.max_samples = -1
+
```
aittalam, line 154: I think this code might be superfluous: it does not affect the returned `job_params` (as it is executed after setting it) and the reason why the test below succeeds is that we have also set a default value in the pydantic schema. Just to be sure, I have tried to run tests after removing this and they are still ok.

lumigator/python/mzai/backend/backend/tests/services/test_job_service.py
```diff
@@ -0,0 +1,22 @@
+
+from backend.services.jobs import JobService
+from lumigator_schemas.jobs import (
+    JobInferenceCreate,
+
+)
+def test_set_null_inference_job_params(job_record,job_service ):
```
aittalam, line 7: TY for adding these and introducing the fake ray client!

aittalam (approved):
Thanks Vicki for the fix and the extra tests!
I left a comment but I think the PR is good to go after that is resolved.

lumigator/python/mzai/backend/backend/services/jobs.py
veekaybee: fixed, thanks!

PR #377, opened 2024-11-15
Enable dummy experiment endpoint

javiermtorres:
## What's changing

An experiment endpoint reusing the jobs service is made available.

## How to test it

N/A for the moment

N/A for the moment

## I already...

### DIFF
--- a/lumigator/python/mzai/backend/backend/api/routes/experiments.py
+++ b/lumigator/python/mzai/backend/backend/api/routes/experiments.py
@@ -1,8 +1,59 @@
-from fastapi import APIRouter
+from uuid import UUID
+
+from fastapi import APIRouter, status
+from lumigator_schemas.experiments import (
+    ExperimentCreate,
+    ExperimentResponse,
+    ExperimentResultDownloadResponse,
+    ExperimentResultResponse,
+)
+from lumigator_schemas.extras import ListingResponse
+from lumigator_schemas.jobs import (
+    JobEvalCreate,
+)
+
+from backend.api.deps import JobServiceDep
 
 router = APIRouter()
 
 
+@router.post("/", status_code=status.HTTP_201_CREATED)
+def create_experiment(
+    service: JobServiceDep,
+    request: ExperimentCreate,
+) -> ExperimentResponse:
+    return service.create_job(JobEvalCreate(request.model_dump()))
+
+
+@router.get("/{experiment_id}")
+def get_experiment(service: JobServiceDep, experiment_id: UUID) -> ExperimentResponse:
+    return ExperimentResponse(service.get_job(experiment_id).model_dump())
+
+
 @router.get("/")
-def get_experiments():
-    return {"message": "Experiments"}
+def list_experiments(
+    service: JobServiceDep,
+    skip: int = 0,
+    limit: int = 100,
+) -> ListingResponse[ExperimentResponse]:
+    return [ExperimentResponse(job.model_dump()) for job in service.list_jobs(skip, limit)]
+
+
+@router.get("/{experiment_id}/result")
+def get_experiment_result(
+    service: JobServiceDep,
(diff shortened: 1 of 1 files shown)


lumigator/python/mzai/backend/backend/api/routes/experiments.py
```diff
@@ -1,8 +1,65 @@
...
 @router.get("/")
-def get_experiments():
-    return {"message": "Experiments"}
+def list_experiments(
+    service: JobServiceDep,
+    skip: int = 0,
+    limit: int = 100,
+) -> ListingResponse[ExperimentResponse]:
+    return ListingResponse[ExperimentResponse].model_validate(
+        service.list_jobs(skip, limit).model_dump()
```
ividal, line 42: Right, this has to be the short-term compromise. The drawback is any job (even if it's just an inference job) will show up as an experiment.

Not for this PR, as this is something the Jobs service doesn't support either, but it feels like we should be able to return jobs of a specific type (e.g. evaluation only) --> #380

ividal (approved):
Thanks a lot! At least the UI will not have to worry about what we decide regarding tracking, workflows, etc. :)

Tested it as well (upload dialogsum, launch experiment based on the included ground-truth), listed experiments, downloaded experiment results.

ividal:
By the way, we're going to need to introduce some tests for this one - current tests (not even integration ones) do not cover this functionality.

javiermtorres:
> By the way, we're going to need to introduce some tests for this one - current tests (not even integration ones) do not cover this functionality.

Roger and wilco :)

veekaybee:
I'd like to better understand the use-case for this. Generally, it would make sense to write a separate experiment service and we decide on how we'd like tracking to work at the experiment level. I'd be hesitant at this stage to mix `jobs` and `experiments` since they're different levels of hierarchy at the moment.

ividal:
> I'd like to better understand the use-case for this. Generally, it would make sense to write a separate experiment service and we decide on how we'd like tracking to work at the experiment level. I'd be hesitant at this stage to mix `jobs` and `experiments` since they're different levels of hierarchy at the moment.

The goal is to abstract API consumers from the changes around `/experiments` and `/jobs, and avoid breaking the API unless we are permanently removing functionality on purpose).

I completely agree with you that jobs are a different abstraction, which is why /experiments/evaluation and experiments/inference were not something that we intended folks to use, they were a temporary by-product.

Here, the functionality folks could rely on was "`/experiments` carries out some steps to let me specify some data, a model and get a score back", whereas jobs aims at giving the person full control of one of the steps. We can keep that spirit (and in fact it's what we want in the future, when tracking is integrated), so we don't need to have a dummy `/experiments` around. 

In the near future, we'll have to decide how exactly experiments will collect/trigger/x multiple jobs and how we track them. Until then, the UI doesn't need to care and can still use them as an entrypoint to workflows.

lumigator/python/mzai/backend/backend/tests/api/routes/test_api_workflows.py
```diff
@@ -53,6 +55,44 @@ def test_upload_data_launch_job(local_client: TestClient, dialog_dataset):
...
+
+    create_experiments_response = local_client.post(
+        "/experiments/", headers=headers, json=payload
+    )
+    assert create_experiments_response.status_code == 201
+
+    get_experiments_response = local_client.get(
+        "/experiments/")
+    get_experiments = ListingResponse[ExperimentResponse].model_validate(get_experiments_response.json())
+    print(get_experiments)
```
ividal, line 88: Oops
```suggestion
    print(get_experiments)
```

ividal (approved):
Thanks! Just a tiny thing there and I'd merge. We're far from sorting out clean jobs, experiments and tracking here, but this is a small step to not bothering UI with the debate.

PR #378, opened 2024-11-15
GitHub actions: allow docker image build to be a required check

peteski22:
## What's changing

The `Docker image CI` workflow can be run on merges to `main` and PR branches. It is required as a status check before allowing a merge to occur, but there are times when we don't need to rebuild the image, for example if only the `/docs` folder changes in a PR it's a waste of time and resources to build the docker image. 

With this PR change we will always trigger the workflow, but it can successfully 'exit early' if a rebuild is not required.

The current folders that are examined for changes are:

* `.github/**`
* `lumigator/**`

We use [dorny/paths-filter](https://github.com/dorny/paths-filter) to do the heavy lifting/checking for us, previous iterations of this PR tried to do a lot of juggling with `git diff `and `git diff-tree` etc.

The [GH runner output](https://github.com/mozilla-ai/lumigator/actions/runs/11894989446/job/33143495270?pr=378#step:3:16) can be used to see what changes were found and whether we decided a rebuild was required or not.

## I already...

### DIFF
--- a/.github/workflows/build_images_uv.yaml
+++ b/.github/workflows/build_images_uv.yaml
@@ -13,7 +13,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  build-image:
+  build:
     name: Build Docker image
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/build_images_uv.yaml
+++ b/.github/workflows/build_images_uv.yaml
@@ -2,45 +2,58 @@ name: Docker image CI
 
 on:
   pull_request:
-    paths:
-      - "lumigator/**"
-      - ".github/**"
+    branches:
+      - "**"
     # synchronized is when you push new commits
     types: ["opened", "synchronize"]
   push:
     branches:
-      - main
+      - "main"
   # required to enable manual triggers on the GH web ui
   workflow_dispatch:
 
 jobs:
-  build:
+  build-image:
     name: Build Docker image
     runs-on: ubuntu-latest
-
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/build_images_uv.yaml
```diff
@@ -1,26 +1,33 @@
...
     steps:
-      - name: Checkout
+      - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Check modified paths
+        run: |
+          if ! git diff --name-only ${{ github.event.before }} ${{ github.event.pull_request.head.sha }} | grep -E '^lumigator/|^\.github/'; then
```
chainlink, line 25: Checks like these have a lot of fiddly edge cases and unfortunately have been nothing but trouble in past iterations. Why not use the built in file filtering in the GH syntax?

peteski22: Thanks. I just left if in draft as I was testing some things out but needed the GH actions to run.

I wanted to see if I could get it working and also be included as a required status check (which would require it to run each time). 

I’ll check out the file filtering, and was also going to look at https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif

peteski22: This makes the PR much neater: https://github.com/dorny/paths-filter (thanks @agpituk for pointing me to it).

macaab26 (approved):

PR #385, opened 2024-11-18
Separating out setting model type into a separate method in the job service

veekaybee:
## What's changing

We currently have a lot going on in the `create_job` method which makes it hard to unit test granularly. We've previously done some work in pulling out separate methods (see https://github.com/mozilla-ai/lumigator/pull/328), this PR does the same for `_set_model_type` and adds tests also clarifying the model URL (it wouldn't be S3 since we use either huggingface, openai or mistral at the moment) in the SDK. 

## How to test it

Run unit tests

## I already...

cc @peteski22 @javiermtorres fyi for SDK changes


veekaybee:
Also adding ruff precommit hooks here because they're not getting caught before they reach the linter in CI

lumigator/python/mzai/backend/backend/api/routes/health.py
```diff
@@ -33,7 +33,7 @@ def get_job_metadata(job_id: UUID) -> JobSubmissionResponse:
         )
     elif resp.status_code != HTTPStatus.OK:
         loguru.logger.error(
-            f"Unexpected status code getting job metadata text: "
+            "Unexpected status code getting job metadata text: "
              "{resp.status_code}, error: {resp.text or ''}"
```
javiermtorres, line 37:
```suggestion
            "Unexpected status code getting job metadata text: "
            f"{resp.status_code}, error: {resp.text or ''}"
```
I'm to blame for not separating the text lines properly :(

lumigator/python/mzai/sdk/tests/data/job-all.json
```diff
@@ -4,8 +4,8 @@
     "model": "some_model",
     "dataset": "6f6487ac-7170-4a11-af7a-0f6db1ec9a74",
     "max_samples": 5,
-    "model_url": "s3://some_url",
+    "model_url": "hf://facebook/bart-large-cnn",
```
aittalam, line 7: In theory `model` should contain the model identifier, `model_url` the actual URL the model is served from (just in the case of inference APIs), and the two can coexist but `model_url` can also be null.

You can change this with an actual URL if you want, but I think this data file (or its corresponding SDK test) is not correctly capturing our format. I also see there are still `config_*_template` instead of the single `config_template` parameter but no error was raised during SDK tests so I [opened an issue ](https://github.com/mozilla-ai/lumigator/issues/393) for this. Feel free to ignore the issue for the sake of this PR!

lumigator/python/mzai/backend/backend/tests/services/test_job_service.py
```diff
@@ -18,7 +18,38 @@ def test_set_explicit_inference_job_params(job_record,job_service ):
...
+                                 dataset="cced289c-f869-4af1-9195-1d58e32d1cc1")
+
+    model_url = job_service._set_model_type(request)
+    assert model_url == "https://api.openai.com/v1"
+
+def test_set_model_mistral(job_record,job_service):
+    request = JobInferenceCreate(name="test_run_mistral",
+                                 description="Test run for Mistral server",
+                                 max_samples=10,
+                                 model="mistral://open-mistral-7b",
```
aittalam, line 51: nit: as the only thing changing here is the model and the returned values, wdyt about using @pytest.mark.parametrize for that?

lumigator/python/mzai/backend/backend/tests/services/test_job_service.py
```diff
@@ -18,7 +18,38 @@ def test_set_explicit_inference_job_params(job_record,job_service ):
...
+
+def test_set_model_openai(job_record,job_service):
+    request = JobInferenceCreate(name="test_run_openai",
+                                 description="Test run for OpenAI server",
+                                 max_samples=10,
+                                 model="oai://gpt-4-turbo",
+                                 dataset="cced289c-f869-4af1-9195-1d58e32d1cc1")
+
+    model_url = job_service._set_model_type(request)
+    assert model_url == "https://api.openai.com/v1"
```
aittalam, line 45: As this URL and Mistral's are fixed and we might want to update them in different parts of the code (at least two now, ie. both jobs and test), do you think it makes sense to store them in settings? Honestly I am not 100% sure about this (as it's not a real _user setting_ and one can always look for instances of the URL and replace them if needed), I'm fine whatever your choice is

aittalam (approved):
TY Vicki, I left a few comments but overall this LGTM.

lumigator/python/mzai/sdk/tests/data/job-all.json
veekaybee: I'll change this to a URL for now and we can address the issue you opened-thanks fro that!

lumigator/python/mzai/backend/backend/tests/services/test_job_service.py
veekaybee: Yep, good call...thanks!

lumigator/python/mzai/backend/backend/tests/services/test_job_service.py
veekaybee: Probably makes sense to call this directly from settings in this case for a dynamic update when we change that

PR #394, opened 2024-11-19
Exit link checker and docs workflow early if no changes to /docs

peteski22:
## What's changing

Provide a way to successfully exit early from docs workflows, when no docs files have been changed.

As the docs workflow status is a require status check, we cannot skip the workflow entirely.

## I already...


veekaybee (approved):
LGTM, let's see how this check works. Thanks for this!

Issue #393, opened 2024-11-19
job-all.json file does not capture current API


aittalam:
**Describe the bug**
The [job-all.json](https://github.com/mozilla-ai/lumigator/blob/main/lumigator/python/mzai/sdk/tests/data/job-all.json) file does not capture the current API, which means some of its data is not really used by tests (as they would fail otherwise).

**System Information (please complete the following information):**
 - Lumigator version : [2a6f8f7](https://github.com/mozilla-ai/lumigator/commit/2a6f8f7281bfe3f69ee73863704eca8ed9c8a3e4)

**Expected behavior**
Data in the example JSON should reflect the current API syntax and tests should return an error otherwise

**Screenshots**
Example: "config_infer_template" and "config_eval_template" do not exist anymore

javiermtorres:
Ok, but should I add the extra=forbid in a moment and just check the tests? It should be quick to do.

javiermtorres:
You can see it failing now :)
https://github.com/mozilla-ai/lumigator/actions/runs/11916963413/job/33210860682?pr=396
I'll add the forbid param and clean up the tests.

PR #396, opened 2024-11-19
Check that extra params in job creation are forbidden

javiermtorres:
## What's changing

The `extra=forbid` [model config param](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra) is passed to all OpenAPI-related models in the schema. This will prevent additional keys from being passed when creating a model object from JSON data.

## How to test it

Just run auto tests; no manual testing is needed.

Fixes https://github.com/mozilla-ai/lumigator/issues/393

## I already...

  - Existing SDK tests have been updated to comply with the schemas
  - info about strict checks added to README 
  - The DB models are unchanged

### DIFF
--- a/lumigator/python/mzai/sdk/lumigator_sdk/strict_schemas.py
+++ b/lumigator/python/mzai/sdk/lumigator_sdk/strict_schemas.py
@@ -1,30 +1,24 @@
-from pydantic import ConfigDict
-
 from lumigator_schemas.completions import CompletionResponse
+from lumigator_schemas.datasets import DatasetDownloadResponse, DatasetResponse
 from lumigator_schemas.experiments import (
     ExperimentCreate,
     ExperimentResponse,
     ExperimentResultDownloadResponse,
     ExperimentResultResponse,
 )
-from lumigator_schemas.extras import (
-    ListingResponse,
-    HealthResponse
-)
+from lumigator_schemas.extras import HealthResponse, ListingResponse
 from lumigator_schemas.jobs import (
+    JobConfig,
     JobEvalCreate,
+    JobEvent,
     JobInferenceCreate,
+    JobLogsResponse,
     JobResponse,
     JobResultDownloadResponse,
     JobResultResponse,
     JobSubmissionResponse,
-    JobConfig,
-    JobEvent,
-    JobLogsResponse
 )
-from lumigator_schemas.datasets import (
-    DatasetDownloadResponse, DatasetResponse
-)
+from pydantic import ConfigDict
 
 
 class CompletionResponse(CompletionResponse):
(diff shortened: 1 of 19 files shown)


peteski22 (approved):
LGTM 😄

ividal (approved):
I have a question rather than a comment (yes! - the nice way around): how will this affect going into flexible jobs that may contain job-specific params that we don't yet know about?

If the answer is: "we'll figure that out when we get there and revert this if necessary", that's also fine.

Or is it rather that, since we're composing data models, each model knows what it needs and hence there will not be a problem?

javiermtorres:
> I have a question rather than a comment (yes! - the nice way around): how will this affect going into flexible jobs that may contain job-specific params that we don't yet know about?

As we discussed, this may be covered by a job-specific `job-config` parameter (using a generic JobCreate) that covers the exact job params. If we want to keep this completely flexible it can be `Any`, otherwise it can be the union of known job configs. It is not implemented yet but we do have a way forward.

lumigator/python/mzai/sdk/lumigator_sdk/strict_schemas.py
```diff
@@ -0,0 +1,76 @@
...
+    JobLogsResponse,
+    JobResponse,
+    JobResultDownloadResponse,
+    JobResultResponse,
+    JobSubmissionResponse,
+)
+from pydantic import ConfigDict
+
+
+class CompletionResponse(CompletionResponse):
```
veekaybee, line 24: Looks good, thanks for this! Minor question: what happens when CompletionResponse is updated in schemas and we use it as a dependency here? Do we need to have any checks/testing in place?

javiermtorres: I guess that if the schemas version changes, and the test jsons are no longer compliant with the schemas, we'll seee the SDK tests failing. That'd be the indication that we need to bump the schemas version in the SDK. But the implementation would be more or less the same, since it just takes the base class and adds some config on top of it.

veekaybee: ok so what I'm hearing is we pin at the PIP level for now, makes sense

veekaybee (approved):

veekaybee:
Approved to merge once tests pass

javiermtorres:
@veekaybee Currently the input is taken using the schemas models, and validated against the stricter models (`model_validate(model_dump)`), so the user handles only schemas models. Let me know is this approach is ok with you.

lumigator/python/mzai/sdk/lumigator_sdk/strict_schemas.py
javiermtorres: Keeping unresolved in case you want to discuss this further, otherwise please resolve this item.

javiermtorres: Resolving this issue. Dependencies in pyproject.toml get included in PKG-INFO. The editable annotation seems to be useful only for uv when running from sources directly. But this means that we may want to add some additional check across sdk/schemas versions as part of the release process.

PR #398, opened 2024-11-19
GitHub actions: update link checker and docs build workflows

peteski22:
## What's changing

Link checking needs to happen regardless as it scans the whole repo for file changes (such as on `.md`s) not just `/docs`.

When link checking completes it should trigger the docs building workflow (as it did previously). 

NOTE: the workflow that is the trigger needs to exist in `main` (the default branch) in order for this to work.

See: https://github.com/mozilla-ai/lumigator/pull/394

## I already...


veekaybee (approved):
Thanks!

PR #416, opened 2024-11-21
410 notebook and documentation instructions to install schemas and sdk assume pypi

ividal:
## What's changing
The documentation and notebooks assume we are already automatically pushing to PyPI. This is not the case, which caused this bug for a user.

This PR updates notebook requirements.txt and updates the documentation while we are changing the automatic versioning and publishing.

## How to test it
* Follow the instructions to run the notebook here.
* Run the imports cell.
* Before this PR, it would fail, as the schemas and SDK were not installed. It should now return no errors.

This is a temporary fix, since the underlying problem is not being able to do pip install lumigator-sdk, but at least we won't block folks trying Lumigator.

## I already...

### DIFF
--- a/lumigator/python/mzai/sdk/README.md
+++ b/lumigator/python/mzai/sdk/README.md
@@ -3,16 +3,16 @@
 The SDK provides the communication and validation primitives needed to contact the Lumigator itself,
 either locally or remotely.
 
-You can install the lumigator SDK  via `pip` directly or via `uv`:
+You can install the lumigator SDK  via `pip` directly or via `uv` from the cloned repository:
 
 ```bash
-pip install lumigator-sdk
+pip install -e lumigator/python/mzai/sdk
 ```
 
-or 
+or
 
 ```bash
-uv pip install lumigator-sdk
+uv pip install -e lumigator/python/mzai/sdk
 ```
 
 Now that you have the SDK installed, you can use it to communicate with Lumigator. You can run the
--- a/notebooks/requirements.txt
+++ b/notebooks/requirements.txt
@@ -1,3 +1,4 @@
+-e ../lumigator/python/mzai/sdk
 datasets==2.20.0
 ipywidgets==8.1.3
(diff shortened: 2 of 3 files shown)


lumigator/python/mzai/sdk/README.md
```diff
@@ -3,16 +3,16 @@
 The SDK provides the communication and validation primitives needed to contact the Lumigator itself,
 either locally or remotely.
 
-You can install the lumigator SDK  via `pip` directly or via `uv`:
+You can install the lumigator SDK  via `pip` directly or via `uv` - currently after cloning the repository. Please note that you will need to install `schemas` as well:
```
dpoulopoulos, line 6: When you install the SDK, `schemas` is installed automatically because of [this](https://github.com/mozilla-ai/lumigator/blob/092dbca7be4eb34a8ea1d054c63dd3ee8ec966d9/lumigator/python/mzai/sdk/pyproject.toml#L19).

dpoulopoulos (changes requested):
lgtm after addressing the comments.

notebooks/requirements.txt
```diff
@@ -1,3 +1,5 @@
+-e lumigator/python/mzai/schemas
+-e lumigator/python/mzai/sdk
```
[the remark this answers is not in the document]
ividal: True - I keep thinking everything should be run from root (tests, builds, venv set up and running notebooks) and we should not be cd'ing around, but that's not the route we took. Adapting the paths.

lumigator/python/mzai/sdk/README.md
ividal: Hm, you're right. Installing for me on a new venv had failed unless I installed it explicitly, but I cannot reproduce now, so out it goes.

ividal:
⚠️ Rebased

dpoulopoulos (approved):

PR #420, opened 2024-11-21
Frontend: support single page apps (SPA) when served with nginx

peteski22:
## What's changing

Our frontend operates as a single page application (SPA), and we create a container to host it using nginx when we `make local-up` or `make start-build-lumigator`.

When using the app, the URL bar appears to show different paths (`/datasets`, `/experiments`) but they're all hosted from the same place. However, nginx doesn't know this and so if you attempt to browse directly to a page (rather than the root), or refresh a page you get a 404 not found.

This PR changes the default nginx daemon config so that it will reroute requests to the root index page. In order to add the single line:

```
try_files $uri $uri/ /index.html;
```

We need to supply the entire config file, as such we now have a copy of the `default.conf` modified that is copied by Docker.

## How to test it

* `make start-lumigator-build`
* Navigate to http://localhost in a browser
* Click datasets or experiments in the left hand nav (the URL will change to include this)
* Click refresh in the browser

`http://localhost/experiments`

Before: 

image

After: 

image

To check the default config file you can exec into a container then:

```
cat /etc/nginx/conf.d/default.conf
```

## I already...

### DIFF
--- a/lumigator/frontend/nginx.conf
+++ b/lumigator/frontend/nginx.conf
@@ -1,0 +1,17 @@
+server {
+    listen       80;
+    listen  [::]:80;
+    server_name  localhost;
+
+    location / {
+        root   /usr/share/nginx/html;
+        index  index.html index.htm;
+        try_files $uri $uri/ /index.html;
+    }
+
+    # redirect server error pages to the static page /50x.html
+    error_page   500 502 503 504  /50x.html;
+    location = /50x.html {
+        root   /usr/share/nginx/html;
+    }
+}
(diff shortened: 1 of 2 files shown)


lumigator/frontend/nginx.conf
```diff
@@ -0,0 +1,45 @@
...
+    # redirect server error pages to the static page /50x.html
+    #
+    error_page   500 502 503 504  /50x.html;
+    location = /50x.html {
+        root   /usr/share/nginx/html;
+    }
+
+    # proxy the PHP scripts to Apache listening on 127.0.0.1:80
+    #
+    #location ~ \.php$ {
```
macaab26, line 25: Great work @peteski22! Please delete these comments if they are not related to our solution

macaab26 (approved):

lumigator/frontend/nginx.conf
peteski22: Thanks Mario, I just took the entirety of the `default.conf` in case we ever wanted to diff it later in case things changed across nginx versions, but I can ditch the comments no problem! 😄

Issue #363, opened 2024-11-14
Dynamic versioning


javiermtorres:
**Is your feature request related to a problem? Please describe.**
The git tags and project versions should be kept in sync.

**Describe the solution you'd like**
The version in `pyproject.toml` can be dynamically filled in, using e.g. the [versioningit](https://versioningit.readthedocs.io/en/stable/index.html) tool.

**Describe alternatives you've considered**
The git tag could be checked against the package versions in a CI action.

**Additional context**
N/A

javiermtorres:
The approach we will finally take for the moment is:

* Modify the `pyproject.toml` manually
* Tag and check that the git tag corresponds to the contents of the `pyproject.toml` file

PR #425, opened 2024-11-22
Check tag consistency as a requirement to publishing

javiermtorres:
## What's changing

A semver tag starting with 'v' will trigger publishing to PyPI, but only if the version in the `pyproject.toml` files from SDK and schemas match with the git tag.

## How to test it

Make a commit and git-tag it with a semver tag (e.g. `v0.0.0`) that does not match with the contents of the `pyproject.toml` files either the `lumigator/python/mzai/sdk` or `lumigator/python/mzai/schemas` directories. The publishing step should trigger but fail.

Fixes https://github.com/mozilla-ai/lumigator/issues/363

## I already...

  - No DB migration step required


tools/check_version.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
chainlink, line 1: Do we need two copies of this script?

chainlink (approved):
Approved with one comment

.github/workflows/build_and_push_sdk_packages.yaml
```diff
@@ -4,22 +4,38 @@ on:
   push:
     tags:
     # Not the right semver regexp, but good enough
-      - 'v[0-9]+.[0-9]+.[0-9]+-*'
+      - 'v[0-9]+.[0-9]+.[0-9]+'
```
peteski22, line 7: ❓ Would it make sense for us to prefix SDK on SDK tags? e.g. `sdk/v0.1.0`? 

I'm trying to figure out if this might help us out later when we come to decide how we might handle versions independently (SDK, API, backend, etc.).

tools/check_version.sh
javiermtorres: No, I checked in both copies by mistake. Only the one in tools should be needed (feel free to suggest an alternate location).

.github/workflows/build_and_push_sdk_packages.yaml
javiermtorres: I think it doesn't pay off at this stage (as you mention, we're currently in lockstep). When we finally decide on the version handling for independent components, then we will implement the appropriate mechanism.

PR #431, opened 2024-11-22
Frontend: ensure .env file is available to container during build

peteski22:
## What's changing

When running from a clean clone of the repository, the frontend container is deployed, but is missing environment variables to during the multi stage build process to configure the API URL (`VUE_APP_BASE_URL`).

Reasons for this appear to be:

1. The `.env` file in the root of the repository is ignored by the `.dockerignore` value: `.env/` which was designed to prevent virtual environment folders being uploaded. This means it is not copied to the `/mzai/` WORKDIR.
2. The build system expects the `.env` file to be present locally within the `/frontend` folder.

~This PR allows Docker to copy the `.env` file to the intermediate 'build' stage (by modifying the `.dockerignore` file), and then copies the frontend specific variables to a new (temporary) `.env` file in the `/frontend` folder.~

This PR supplies the environment variable to the intermediate build stage as an `ARG` which will be visible to the `Vite` build system as an env var.

## How to test it

* Clone the repo: `git clone git@github.com:mozilla-ai/lumigator.git`, or navigate to your already cloned repo
* Run `rm lumigator/frontend/.env` (to ensure you have no local artefacts around)
* Checkout the PR (`gh pr checkout 431` if you have the GitHub CLI)
* Run `make start-lumigator-build` (the containers should deploy without error)
* In the browser navigate to http://localhost and attempt to upload a dataset

The developer tools network tab should report no errors, and you should be able to see the dataset appear in the table in the UI.

image

image

This is what it looks like on `main` currently:

image

image

## I already...

### DIFF
--- a/lumigator/frontend/Dockerfile
+++ b/lumigator/frontend/Dockerfile
@@ -7,13 +7,6 @@ COPY ../../ /mzai
 
 WORKDIR /mzai/
 
-# Copy the .env file for the entire project, and perform some transformation on it.
-# Only env vars starting with the 'VUE_APP_' prefix are made visible by the Vite build system (See: vite.config.js),
-# but we extract only these variables in a kind of 'least privilege' way.
-# See: https://cli.vuejs.org/guide/mode-and-env.html#environment-variables
-COPY .env /mzai/lumigator/frontend/
-RUN cat .env | grep -E ^VUE_APP_ > /mzai/lumigator/frontend/.env
-
 # Install dependencies
 RUN npm --prefix /mzai/lumigator/frontend install
 
--- a/lumigator/frontend/Dockerfile
+++ b/lumigator/frontend/Dockerfile
@@ -7,6 +7,13 @@ COPY ../../ /mzai
 
 WORKDIR /mzai/
 
+# Copy the .env file for the entire project, and perform some transformation on it.
+# Only env vars starting with the 'VUE_APP_' prefix are made visible by the Vite build system (See: vite.config.js),
+# but we extract only these variables in a kind of 'least privilege' way.
+# See: https://cli.vuejs.org/guide/mode-and-env.html#environment-variables
+COPY .env /mzai/lumigator/frontend/
+RUN cat .env | grep -E ^VUE_APP_ > /mzai/lumigator/frontend/.env
+
 # Install dependencies
 RUN npm --prefix /mzai/lumigator/frontend install
 
@@ -17,7 +24,7 @@ RUN npm --prefix /mzai/lumigator/frontend run build
 FROM nginx:1.27.2-alpine-slim AS server
 
 # Copy built files to the Nginx image
-COPY --from=base  /mzai/lumigator/frontend/dist /usr/share/nginx/html
+COPY --from=base /mzai/lumigator/frontend/dist /usr/share/nginx/html
 
 # Copy a default nginx config, adjusted to support single page applications
 COPY --from=base /mzai/lumigator/frontend/nginx.conf /etc/nginx/conf.d/default.conf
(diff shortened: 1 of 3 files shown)


chainlink:
I think this env var should be set in the docker compose file

peteski22:
> I think this env var should be set in the docker compose file

Do you mean duplicated there? If someone runs the UI with !Docker then it will need the env var to work (like npm run dev etc).

peteski22:
@chainlink, @aittalam I spent some time double checking something (including reading the `Vite` docs), and I think the reason you're (@aittalam) seeing the error still is that the env var **doesn't** need to be present in the final container that is deployed, but it does need to be present during the first part of our multi-stage build. 

The `.env` file is used by `Vite` as it builds the `dist` package, and the environment variable is evaluated during the build and baked into the resulting JS as a literal.

I don't think having it in the docker compose file would work here. Maybe there's some other way we want to approach the build (as there seem to be a number of options), but perhaps this PR can be approved in order to resolve the issue in the short term?

lumigator/frontend/Dockerfile
```diff
@@ -7,6 +7,13 @@ COPY ../../ /mzai
 
 WORKDIR /mzai/
 
+# Copy the .env file for the entire project, and perform some transformation on it.
+# Only env vars starting with the 'VUE_APP_' prefix are made visible by the Vite build system (See: vite.config.js),
+# but we extract only these variables in a kind of 'least privilege' way.
+# See: https://cli.vuejs.org/guide/mode-and-env.html#environment-variables
+COPY .env /mzai/lumigator/frontend/
+RUN cat .env | grep -E ^VUE_APP_ > /mzai/lumigator/frontend/.env
```
aittalam, line 15: If we are interested only in `VUE_APP_*` vars, do we need an `.env` file copied into frontend (and to allow .env files at all in dockerignore)?
Unless it is required locally too, we might as well have a `.vue_environment` (just making up a name here) file and then `COPY .vue_environment /mzai/lumigator/frontend/.env`. WDYT?

(still wondering why Vite cannot access env vars from the environment but looks for them in a file, but I think the simplest solution would work for now while we think about how to implement this in a way which is more consistent with the rest of the stack)

peteski22: I believe the intention behind our single `.env` file was to consolidate the env vars, having a different file like the example you gave (`.vue_environment`) would mean we're back to having env vars in different places. 🤔 

Vite (the build system) can access actual environment variables during the build (they take priority over [`.env`](https://github.com/motdotla/dotenv) ones), but env vars supplied via docker compose only appear in the finished frontend container. The build tool doesn't have access to them, as they would need to exist in the first stage of the multi-stage build we do (that we throw away). 

From: https://vite.dev/guide/env-and-mode#env-files

> In addition, environment variables that already exist when Vite is executed have the highest priority and will not be overwritten by `.env` files.

This solution does ensure that we only need to configure the variable in a single place, and the Dockerfile only ends up with `VUE_APP_` prefixed vars which are part of the throw-away stage.  

I'm probably missing something but I don't fully understand why the proposed solution might not be acceptable (even as a short term one). 😢

peteski22: I'm going to have a play with 'one more thing' to see if there's a bit of a workaround for the docker compose file...

peteski22: Victory? 581a6742519114d91a6aeee0912885bce219e32b

javiermtorres: @aittalam Please remember that this setting needs to be set up at image build time, so it won't be reloaded by the final container if it is changed. The final container has only static files that cannot be changed at the time the UI starts. 
@peteski22 approach seems fine, and can use env vars already present in your environment, so I'd approve this PR now, and inspect the build system later. Wdyt?

javiermtorres (approved):

PR #441, opened 2024-11-26
API: Allow CORS API requests no matter what FQDN the UI is hosted on

peteski22:
## What's changing

This PR changes the backend API configuration to allow setting the CORS allowed origins that are accepted. It also defaults to `http://localhost` and `http://localhost:3000` when nothing is configured.

Previously it was limited to only accepting requests from the UI when it the UI is hosted via HTTP on `localhost` ports `80` or `3000`.

## How to test it

* `gh pr checkout 441`

**1. No environment variable set**

This example starts Lumigator with no env var configured, the default will be used.

* `make local-up`
* `cd lumigator/frontend`
* `npm install`
* `npm run dev`

Open `http://127.0.0.1:3000` in a browser and the health check (top right) should report `OK` (no errors shown, no errors in the developer tools network tab/console).

Also the logs in the backend container will show:

```
INFO     | backend.main:create_app:57 - Configuring CORS for API, allowed origins: ['http://localhost', 'http://localhost:3000']
```

**2. Configure local environment variable**

This example starts Lumigator specifying a list of allowed origins that should be used.

* `LUMI_API_CORS_ALLOWED_ORIGINS="http://localhost:8080" make local-up`
(note: you can also use `LUMI_API_CORS_ALLOWED_ORIGINS="*" make local-up` to allow all)
* `cd lumigator/frontend`
* `npm install`
* `npm run build`
* `cd dist`
* `npx http-server`

Open `http://127.0.0.1:8080` in a browser and the health check (top right) should report `OK` (no errors shown, no errors in the developer tools network tab/console).

Also the logs in the backend container will show:

```
INFO     | backend.main:create_app:57 - Configuring CORS for API, allowed origins: ['http://localhost:8080']
```

**3. Configure environment variable without required origin**

This example starts Lumigator, specifying allowed origins, but **NOT** the one for `8080` which is required for our test UI to work.

* `LUMI_API_CORS_ALLOWED_ORIGINS="http://localhost:9999" make local-up`
* `cd lumigator/frontend`
* `npm install`
* `npm run build`
* `cd dist`
* `npx http-server`

Open `http://127.0.0.1:8080` in a browser and the health check (top right) should fail.

image

image

Also the logs in the backend container will show:

```
INFO     | backend.main:create_app:57 - Configuring CORS for API, allowed origins: ['http://localhost:9999']
```

The UI is hosted on `8080` in this example, which means a CORS pre-flight OPTIONS request will be sent to the backend asking if it allows cross-domain requests from 'http://localhost:8080' (the scheme+host+port are important to CORS).

Previously only `http://localhost:80` and `http://localhost:3000` were allowed so the UI would fail to make requests to the API to get health status, datasets etc. but now it works.

## I already...

### DIFF
--- a/.github/workflows/build_images_frontend.yaml
+++ b/.github/workflows/build_images_frontend.yaml
@@ -18,6 +18,8 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Checkout
+        uses: actions/checkout@v4
 
       - name: Check for modified paths
         uses: dorny/paths-filter@v3


lumigator/python/mzai/backend/backend/main.py
```diff
@@ -51,13 +52,9 @@ def create_app() -> FastAPI:
...
 
     # Adding CORS middleware
-    origins = [
-        "http://localhost",
-        "http://localhost:3000",
-    ]
     app.add_middleware(
         CORSMiddleware,
-        allow_origins=origins,  # Adjust this list as needed for security (e.g., ["http://localhost:3000"])
+        allow_origins=["*"],
```
agpituk, line 57: Can this be done as an environment variable that by default is a `*`?  so we can allow users to customize it.

agpituk (changes requested):

lumigator/python/mzai/backend/backend/main.py
peteski22: Yep, good idea.

lumigator/python/mzai/backend/backend/settings.py
```diff
@@ -118,5 +121,19 @@ def RAY_JOBS_URL(self) -> str:  # noqa: N802
...
+    @property
+    def API_CORS_ALLOWED_ORIGINS(self) -> list[str]:  # noqa: N802
+        """Retrieves the (comma separated) list of allowed origin URLs for CORS requests
+        against the Lumigator backend API.
+
+        Defaults to "*" (allows all) when no environment variable is set.
+        """
+        origins = self._api_cors_allowed_origins.strip()
+        if origins == "":
+            return ["*"]
```
aittalam, line 134: Just raising a concern here about defaulting to * when nothing is provided. WDYT about providing "localhost" (the strictest setting) by default and require users to provide a different value if the setup changes?

peteski22: I understand the hesitancy, I've made some changes so the defaults are configured in the `.env` file now (and no longer use `*`)

aittalam (approved):

agpituk:
Thanks for all the changes! looks good!! :)

PR #429, opened 2024-11-22
Test rearrangement

javiermtorres:
## What's changing

SDK and backend tests are now arranged in `unit` and `int` directories. Specific `make` targets are provided for `backend|sdk,unit|int` combinations.

It is still unclear how to structure the tests. Currently each of the integration targets starts its own backend or services if needed. Currently the CI actions are structured to mimic the current setup.

## How to test it

Although the CI actions have been updated, the following `make` targets can be tested manually:

* `test`
  * `sdk-test`
    * `sdk-unit-test`
    * `sdk-int-test`
  * `backend-test`
    * `backend-unit-test`
    * `backend-int-test`

## Additional notes for reviewers

N/A

## I already...

- [x] added some tests for any new functionality
- [ ] updated the documentation
- [x] checked if a (backend) DB migration step was required and included it if required
  - No DB update needed

Issue #422, opened 2024-11-22
Testing: S3 mocking


javiermtorres:
**Is your feature request related to a problem? Please describe.**
Currently some backend tests depend on a working S3 as their only external service dependency. Using a mock of some kind would allow us to run them faster.

**Describe the solution you'd like**
Replace the current testcontainer localstack implementation with a more lkightweight alternative that does not require containers, like https://github.com/getmoto/moto

**Describe alternatives you've considered**
Some possible approaches:
* https://github.com/getmoto/moto for a full S3 mock
* Plain mocks, requiring more effort for present and future developments, but allowing more flexibility and removing dependency on 3rd parties

**Additional context**
N/A

PR #442, opened 2024-11-26
makefile: Run tests without altering state

### LINKED ISSUE #414: [BUG] `make test` tears down containers started with `make local-up`
**Describe the bug**

> Referencing [this comment](https://github.com/mozilla-ai/lumigator/issues/326#issuecomment-2473140569) here as the issue was not tackled by the PR.

`make test` runs the following:
```
make start-lumigator-build
cd lumigator/python/mzai/backend; SQLALCHEMY_DATABASE_URL=sqlite:///local.db uv run pytest
cd lumigator/python/mzai/sdk; uv run pytest -o python_files="test_*.py int_test_*.py"
make stop-lumigator
```

This results in the following issues:

- Containers are started, ignoring whether any container are already running (in the typical developer setting, they are already up in attached mode with `make local-up`).
- If everything works, all containers are teared down including those started with `make local-up`.
- If anything breaks during tests, the last command (`make stop-lumigator`) is not run and the containers stay up.

 - OS: macOS Sonoma
 - Lumigator version: [092dbca](https://github.com/mozilla-ai/lumigator/commit/092dbca7be4eb34a8ea1d054c63dd3ee8ec966d9)

**To Reproduce**

1. Run `make local-up`.
1. Run `make test` in another window.
1. If the tests end successfully, all the containers started with `make local-up` are now down.

**Expected behavior**

Make sure we end up in the same status we started with:

- Check whether there are already running containers before starting tests:
  - If so, simply run tests.
  - If not: (1) start containers, (2) run tests, (3) tear down containers (regardless of tests outcome).

I think we might also want to make an explicit choice about which docker compose profile we want to use for tests (`local`, `local-fe`) and which docker compose files we want to use (`default`, `dev`). I would lean towards the profile which is closest to development, but I am happy to hear your thoughts about this.

dpoulopoulos:
## What's changing

Currently, running `make test` has the following consequences:

* Containers are started, ignoring any container that is already running.
* If everything works, all containers are teared down including those started with `make local-up`.
* If anything breaks during tests, `make stop-lumigator` is not run and the containers stay up.

Transform the `make test` target to check whether there are already running containers before starting tests:
  - If so, simply run tests.
  - If not: (1) start containers, (2) run tests, (3) tear down containers (regardless of tests outcome).

## How to test it

1. Run `make local-up`.
1. Run `make test` in another window.
1. If the tests complete successfully, all the containers started with `make local-up` should remain running.
1. Run `make stop-lumigator`.
1. Run `make test`. New containers should start and be removed when the tests finish, regardless of whether they succeed or fail.

## I already...

- [NA] added some tests for any new functionality
- [NA] updated the documentation
- [NA] checked if a (backend) DB migration step was required and included it if required

Closes #414

### DIFF
--- a/.github/workflows/test_backend_uv.yaml
+++ b/.github/workflows/test_backend_uv.yaml
@@ -28,7 +28,7 @@ jobs:
         continue-on-error: false
 
       - name: Run backend unit tests
-        run: make backend-unit-test
+        run: make test-backend-unit
 
       - name: Run backend integration tests
-        run: make backend-integration-test
+        run: AUTO_TEST_RUN=y make test-backend-integration
--- a/.github/workflows/test_sdk_integration.yaml
+++ b/.github/workflows/test_sdk_integration.yaml
@@ -24,4 +24,4 @@ jobs:
         run: uv python install
 
       - name: Run SDK integration tests
-        run: make sdk-integration-test
+        run: AUTO_TEST_RUN=y make test-sdk-integration
--- a/.github/workflows/test_sdk_uv.yaml
+++ b/.github/workflows/test_sdk_uv.yaml
@@ -36,4 +36,4 @@ jobs:
 
       - name: Run SDK unit tests
         if: steps.filter.outputs.run_sdk_unit_tests == 'true'
-        run: make sdk-unit-test
+        run: make test-sdk-unit
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,9 @@
-.PHONY: local-up local-down local-logs clean-docker-buildcache clean-docker-images clean-docker-containers start-lumigator-external-services start-lumigator stop-lumigator test sdk-test sdk-unit-test sdk-integration-test-exec sdk-integration-test backend-test backend-unit-test backend-integration-test
+.PHONY: local-up local-down local-logs clean-docker-buildcache clean-docker-images clean-docker-containers start-lumigator-external-services start-lumigator stop-lumigator test-sdk-unit test-sdk-integration test-sdk-integration-target test-sdk-target test-backend-unit test-backend-integration test-backend-integration-target test-backend-target test-all-target test-%
 
 SHELL:=/bin/bash
 UNAME:= $(shell uname -o)
+PROJECT_ROOT := $(shell git rev-parse --show-toplevel)
+CONTAINERS_RUNNING := $(shell docker ps -q --filter "name=lumigator-")
 
 #used in docker-compose to choose the right Ray image
 ARCH := $(shell uname -m)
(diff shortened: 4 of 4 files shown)


peteski22:
FYI: 

This PR may depend on https://github.com/mozilla-ai/lumigator/pull/429.

Separately: 

I know the issue says we should just use the containers that already exist if they're there, but there's something in that which makes me worried that we might not actually be testing (locally) the code as it stands. 

If the existing containers were started with `make start-lumigator-build` then code changed, and `make test` run, are we covered in that kind of scenario?

javiermtorres:
> FYI:
> 
> This PR may depend on #429.
 
Yes, @dpoulopoulos I'd appreciate it if you could check how the two will work together (hopefully the targets there are flexible enough to be reused in this PR)

> Separately:
> 
> I know the issue says we should just use the containers that already exist if they're there, but there's something in that which makes me worried that we might not actually be testing (locally) the code as it stands.
> 
> If the existing containers were started with `make start-lumigator-build` then code changed, and `make test` run, are we covered in that kind of scenario?

I agree with this; could also happen if we are developing a couple of PRs at the same time. I'd be a bit wary of implicit container handling. Maybe we can explicitly signal if we want to stop containers or not (via env var???).

dpoulopoulos:
@peteski22 @javiermtorres what if the script asked you? For example:

```
Lumigator is running. Do you want to continue testing with the existing setup (Y/n):
```

javiermtorres:
> @peteski22 @javiermtorres what if the script asked you? For example:
> 
> ```
> Lumigator is running. Do you want to continue testing with the existing setup (Y/n):
> ```
 
LGTM, and maybe an env var / cli flag too for automated scripts (anything that ensures that input does not stall some auto action).

> Also, Javier, I suggest you merge #429 and then I can rebase this on top of it. WDYT?

Thanks a lot. I will do it ASAP. I can help you with conflicts as you find them.

aittalam (approved):
Thank you Dimitris!
I tried running the tests with `make tests-all` from your branch with containers up and down, this works as expected.

aittalam:
Just a follow-up comment and a question for @javiermtorres who has more experience with tests: when I run unit tests on backend I see the following:

image

The most recent containers (up 2 seconds) are those started by the unit tests. I now see two localstack containers now, do you think there could be a chance for issues / conflicts?

javiermtorres:
> The most recent containers (up 2 seconds) are those started by the unit tests. I now see two localstack containers now, do you think there could be a chance for issues / conflicts?

This may be related to an issue I have mentioned at times that happens in my environment (and that other people doesn't seem to have found - yet). At times, my testcontainers localstack code will complain in calls to the docker runtime, presumably due to this issue. I am now working on https://github.com/mozilla-ai/lumigator/issues/422 to replace the testcontainers localstack with either a live localstack from our docker compose, or a suitable fake s3 server. I'll open a PR shortly, probably today.

javiermtorres (approved):
If it works, fine for me. I'd prefer some more `make`-native mechanism, but we're already stretching the `make` scope anyway, so anything that solves the issue is ok for me.

PR #451, opened 2024-11-28
backend: Refresh job status

### LINKED ISSUE #382: [BUG] The `/jobs` endpoint should be able to return an up to date job status
**Describe the bug**

The API has an endpoint apparently able to return the status of a job (`/jobs/{job_id}`) but this status is not updated beyond "created". Currently, one has to poll ray directly to get up to date information, which is not ideal.

 - OS: independent from OS, but tested on MacOS Sonoma and Ubuntu 24.04
 - Lumigator version: current main ( [2a6f8f7281bfe3f69ee73863704eca8ed9c8a3e4](https://github.com/mozilla-ai/lumigator/tree/2a6f8f7281bfe3f69ee73863704eca8ed9c8a3e4) )

**To Reproduce**

Steps to reproduce the behavior, via the API:
1. Upload a dataset
1. Launch a job (e.g. evaluation with BART)
1. List jobs, find the job_id of one (e.g., the one you just created)
4. GET `/jobs/{job_id}` --> created
5. Check on the ray dashboard until that job has finished
6. GET `/jobs/{job_id}` --> still created

**Expected behavior**

Since these jobs are async, the endpoint should return "finished" as a status when the job is done.

dpoulopoulos:
## What's changing

The `/jobs` endpoint can return the status of a specific Ray job, but this status is not updated beyond "created".

Optionally refresh the database record before returning:

* Immediately return statuses "Succeeded" and "Failed" because they cannot change.
* Otherwise, consult Ray first, and if the status differs from the one in the DB, update the DB and return the updated record.

Closes #382

## How to test it

1. Upload a dataset
1. Launch a job (e.g. evaluation with BART)
1. List jobs, find the job_id of one (e.g., the one you just created)
1. GET /jobs/{job_id} --> created
1. Check on the ray dashboard until that job has finished
1. GET /jobs/{job_id} --> finished

## I already...

- [NA] updated the documentation
- [NA] checked if a (backend) DB migration step was required and included it if required

### DIFF
--- a/lumigator/python/mzai/backend/backend/settings.py
+++ b/lumigator/python/mzai/backend/backend/settings.py
@@ -92,6 +92,12 @@ class BackendSettings(BaseSettings):
         # as the default shell in the container is /bin/sh
         return f'if [ `arch` = "aarch64" ]; then export LD_PRELOAD=$VIRTUAL_ENV/{lib_path}; fi;'
 
+    # URL for Ray jobs API
+    @computed_field
+    @property
+    def RAY_VERSION_URL(self) -> str:  # noqa: N802
+        return f"{self.RAY_DASHBOARD_URL}/api/version"
+
     @computed_field
     @property
     def RAY_WORKER_GPUS(self) -> float:  # noqa: N802
--- a/lumigator/python/mzai/backend/backend/tests/unit/api/routes/test_jobs.py
+++ b/lumigator/python/mzai/backend/backend/tests/unit/api/routes/test_jobs.py
@@ -1,0 +1,45 @@
+import json
+import urllib
+from pathlib import Path
+
+from fastapi import status
+from fastapi.testclient import TestClient
+
+from backend.settings import settings
+
+
+def load_json(path: Path) -> str:
+    with Path.open(path) as file:
+        return json.load(file)
+
+
+def test_get_job_status(
+    app_client: TestClient,
+    job_repository,
+    request_mock,
+    json_ray_version,
+    json_data_health_job_metadata_ray,
+):
+    created_job = job_repository.create(name="test", description="")
+
+    # The Ray client will call the Ray API to get the version before gettting the job status
+    # Mock the Ray version API
+    request_mock.get(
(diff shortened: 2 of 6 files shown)


lumigator/python/mzai/backend/backend/tests/unit/api/routes/test_jobs.py
```diff
@@ -0,0 +1,45 @@
...
+def test_get_job_status(
+    app_client: TestClient,
+    job_repository,
+    request_mock,
+    json_ray_version,
+    json_data_health_job_metadata_ray,
+):
+    created_job = job_repository.create(name="test", description="")
+
+    print(f"RAY_VERSION: {settings.RAY_VERSION}")
```
aittalam, line 25: I'd suggest to remove the print if not needed anymore

lumigator/python/mzai/backend/backend/tests/unit/api/routes/test_jobs.py
```diff
@@ -0,0 +1,45 @@
...
+
+    request_mock.get(
```
aittalam, line 27: Can you please add a comment explaining why this is needed?
I think it'd be useful for ppl reading the code later on to understand that the ray client makes that call first, so we need to mock that up too

lumigator/python/mzai/backend/backend/settings.py
```diff
@@ -89,6 +89,12 @@ def LD_PRELOAD_PREFIX(self) -> str:  # noqa: N802
         # as the default shell in the container is /bin/sh
         return f'if [ `arch` = "aarch64" ]; then export LD_PRELOAD=$VIRTUAL_ENV/{lib_path}; fi;'
 
+    # URL for Ray jobs API
+    @computed_field
+    @property
+    def RAY_VERSION(self) -> str:  # noqa: N802
```
aittalam, line 95: may I suggest `RAY_VERSION_URL` here?

aittalam (approved):
Thanks Dimitris! I added a few comments but after addressing them this LGTM 👍

Issue #326, opened 2024-10-30
Lingering test containers after `make test`


ividal:
**Describe the bug**
`make test` launches containers to enable integration tests. However, they are not torn down after.

**System Information (please complete the following information):**
 - OS: macOS Sonoma
 - Lumigator version: [7be2518](https://github.com/mozilla-ai/lumigator/commit/7be2518ec8c6bc59ab8463fc7c39aad078bbb386) 

**To Reproduce**
1. Clone repo
2. docker ps shows no containers
3. make test (requires 128GB available to build the images)
4. docker ps shows lingering containers

**Expected behavior**
No containes should be left there.

**Screenshots**
N/A

**Additional context**
Potentially one might want to only tear down those created during tests, but we can also agree that test should happen with no other lumigator containers running in the background anyway.

aittalam:
Adding further context here:

- `make test` runs the following:
```
test:
    make start-lumigator-build
    cd lumigator/python/mzai/backend; SQLALCHEMY_DATABASE_URL=sqlite:///local.db uv run pytest
    cd lumigator/python/mzai/sdk; uv run pytest
    make stop-lumigator
```

There are a few checks missing:

- containers are started ignoring whether any container is already running (in the typical developer setting, they are already up in attached mode with `make local-up`)
- if anything breaks during tests, the last command (`make stop-lumigator`) is not run
- if everything works, all containers are teared down including those started with `make local-up`

Ideally what should happen is the following:

- check whether there are containers already running
- if so, run tests
- otherwise, (1) start containers, (2) run tests, (3) tear down containers

ividal:
Fixed with #327

PR #453, opened 2024-11-28
Add fakes for the S3 service and use the standalone localstack

javiermtorres:
## What's changing

Unit tests in the backend can benefit from a BytesIO/StringIO backed fake S3, with both FakeS3Client and FakeS3FS. Integration tests will use an existing localstack, if any.

The previously used localstack started by testcontainers has been removed.

## How to test it

Just run the unit tests without the auxiiliary containers, and run the integration tests with the auxiliary containers.

Fixes https://github.com/mozilla-ai/lumigator/issues/422

## I already...

  - Not needed since this is a test improvement.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,4 +1,11 @@
 # .env.example
+
+# Lumigator container control
+# Set to "TRUE" if the containers need to be up and running after
+# a test target failed (e.g. in CI where containers are inspected
+# for logs after failed steps)
+CONTAINERS_UP_ON_ERROR="FALSE"
+
 # Lumigator API configuration
 # LUMI_API_CORS_ALLOWED_ORIGINS:
 # Comma separated list of origins (See: https://developer.mozilla.org/en-US/docs/Glossary/Origin)
--- a/.github/workflows/test_backend_uv.yaml
+++ b/.github/workflows/test_backend_uv.yaml
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
       - "**"
+  workflow_dispatch:
 
 jobs:
   backend-integration-tests:
@@ -31,4 +32,17 @@ jobs:
         run: make test-backend-unit
 
       - name: Run backend integration tests
-        run: AUTO_TEST_RUN=y make test-backend-integration
+        run: KEEP_CONTAINERS_UP=FALSE AUTO_TEST_RUN=y make test-backend-integration
+
+      - name: Collect Ray logs in case of failure
+        if: failure()
+        run: |
+          mkdir /tmp/raylogs
+          docker cp lumigator-ray-1:/tmp/ray/ /tmp/raylogs/
+
+      - name: Upload Ray logs in case of failure
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: ray-logs
+          path: /tmp/raylogs/ray/session_*/logs/
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,10 @@ UNAME:= $(shell uname -o)
 PROJECT_ROOT := $(shell git rev-parse --show-toplevel)
 CONTAINERS_RUNNING := $(shell docker ps -q --filter "name=lumigator-")
 
+-include .env
+
+KEEP_CONTAINERS_UP ?= "TRUE"
+
 #used in docker-compose to choose the right Ray image
 ARCH := $(shell uname -m)
 RAY_ARCH_SUFFIX :=
@@ -16,7 +20,7 @@ endif
 define run_with_containers
 	@echo "No Lumigator containers are running. Starting containers..."
 	make start-lumigator-build
-	trap "cd $(PROJECT_ROOT); make stop-lumigator" EXIT; \
+	@if [ $(KEEP_CONTAINERS_UP) = "TRUE" ]; then trap "cd $(PROJECT_ROOT); make stop-lumigator" EXIT; fi; \
 	make $(1)
 endef
 
@@ -118,13 +122,13 @@ else
 endif
 
 test-backend-unit:
-	cd lumigator/python/mzai/backend/backend/tests; \
-	SQLALCHEMY_DATABASE_URL=sqlite:///local.db uv run pytest -o python_files="backend/tests/unit/*/test_*.py"
+	cd lumigator/python/mzai/backend/; \
+	SQLALCHEMY_DATABASE_URL=sqlite:////tmp/local.db uv run pytest -o python_files="backend/tests/unit/*/test_*.py"
 
 test-backend-integration-target:
-	cd lumigator/python/mzai/backend/backend/tests;	\
-	RAY_WORKER_GPUS="0.0" RAY_WORKER_GPUS_FRACTION="0.0" \
-	SQLALCHEMY_DATABASE_URL=sqlite:///local.db uv run pytest -s -o python_files="backend/tests/integration/*/test_*.py"
+	cd lumigator/python/mzai/backend/; \
+	docker container list --all; \
+	SQLALCHEMY_DATABASE_URL=sqlite:////tmp/local.db RAY_WORKER_GPUS="0.0" RAY_WORKER_GPUS_FRACTION="0.0" INFERENCE_PIP_REQS=../jobs/inference/requirements_cpu.txt INFERENCE_WORK_DIR=../jobs/inference EVALUATOR_PIP_REQS=../jobs/evaluator/requirements.txt EVALUATOR_WORK_DIR=../jobs/evaluator uv run pytest -s -o python_files="backend/tests/integration/*/test_*.py"
 
 test-backend-integration:
 ifeq ($(CONTAINERS_RUNNING),)
--- a/lumigator/python/mzai/backend/README.md
+++ b/lumigator/python/mzai/backend/README.md
@@ -28,7 +28,7 @@ The backend needs to retrieve the location of the database used in tests via the
 SQLALCHEMY_DATABASE_URL=sqlite:///local.db uv run pytest
 ```
 
-Note that this will create an SQLite database file named `local.db` in the `backend` directory. Remove it before running another batch of tests.
+Note that this will create an SQLite database file named `local.db` in the `backend` directory. Remove it before running another batch of tests. Also note that the tests as invoked in the Makefile will remove and recreate a file called `local.db` in the `lumigator/python/mzai/backend` directory.
 
 The tests include a unit test suite and an integration test suite. There are make targets available at the root folder, as follows:
 
@@ -36,7 +36,7 @@ The tests include a unit test suite and an integration test suite. There are mak
   * `backend-unit-test`: runs tests in `backend/tests/unit/*/test_*.py` (any depth of subfolders)
   * `backend-int-test`: runs tests in `backend/tests/int/*/test_*.py` (any depth of subfolders)
 
-The SQLite configuration making use of a local file is used in both unit and integration tests. Test containers are started for integration tests.
+The SQLite configuration making use of a local file is used in both unit and integration tests. Test containers are needed for integration tests. Fake or mock services are used in unit tests, usually via the [`requests-mock`](https://pypi.org/project/requests-mock/) package in the case of HTTP APIs. Specifically, the S3 file system driver is replaced with an [in-memory implementation](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.implementations.memory.MemoryFileSystem) of the `AbstractFileSystem` interface via [registry modification](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.registry.register_implementation).
 
 ## Data models
 
--- a/lumigator/python/mzai/backend/backend/config_templates.py
+++ b/lumigator/python/mzai/backend/backend/config_templates.py
@@ -104,7 +104,7 @@ bart_infer_template = """{{
 causal_infer_template = """{{
     "name": "{job_name}/{job_id}",
     "model": {{ "path": "{model_uri}" }},
-    "dataset": {{ "path": "{dataset_path}" }},
+    "dataset": {{ "path": "{dataset_path}" }}
 }}"""
 
 
--- a/lumigator/python/mzai/backend/backend/services/datasets.py
+++ b/lumigator/python/mzai/backend/backend/services/datasets.py
@@ -1,4 +1,5 @@
 import csv
+import traceback
 from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import BinaryIO
@@ -94,6 +95,7 @@ class DatasetService:
         raise HTTPException(status.HTTP_404_NOT_FOUND, f"Dataset '{dataset_id}' not found.")
 
     def _raise_unhandled_exception(self, e: Exception) -> None:
+        traceback.print_exc()
         raise HTTPException(status.HTTP_500_INTERNAL_SERVER_ERROR, str(e)) from e
 
     def _get_dataset_record(self, dataset_id: UUID) -> DatasetRecord | None:
@@ -122,6 +124,7 @@ class DatasetService:
             # Upload to S3
             dataset_key = self._get_s3_key(record.id, record.filename)
             dataset_path = self._get_s3_path(dataset_key)
+            # Deprecated!!!
             dataset_hf.save_to_disk(dataset_path, fs=self.s3_filesystem)
         except Exception as e:
             # if a record was already created, delete it from the DB
--- a/lumigator/python/mzai/backend/backend/services/jobs.py
+++ b/lumigator/python/mzai/backend/backend/services/jobs.py
@@ -241,12 +241,14 @@ class JobService:
 
     def get_job(self, job_id: UUID) -> JobResponse:
         record = self._get_job_record(job_id)
+        loguru.logger.info(f"Obtaining info for job {job_id}: {record}")
 
         if record.status == JobStatus.FAILED or record.status == JobStatus.SUCCEEDED:
             return JobResponse.model_validate(record)
 
         # get job status from ray
         job_status = self.ray_client.get_job_status(job_id)
+        loguru.logger.info(f"Obtaining info from ray for job {job_id}: {job_status}")
 
         # update job status in the DB if it differs from the current status
         if job_status.lower() != record.status.value.lower():
@@ -263,7 +265,7 @@ class JobService:
         records = self.job_repo.list(skip, limit)
         return ListingResponse(
             total=total,
-            items=[JobResponse.model_validate(x) for x in records],
+            items=[self.get_job(record.id) for record in records],
         )
 
     def update_job_status(
--- a/lumigator/python/mzai/backend/backend/tests/conftest.py
+++ b/lumigator/python/mzai/backend/backend/tests/conftest.py
@@ -3,31 +3,35 @@ import io
 import os
 import uuid
 from pathlib import Path
+from unittest.mock import Mock, patch
 
+import boto3
+import fsspec
 import pytest
 import requests_mock
-from botocore.exceptions import ClientError
 from fastapi import FastAPI
 from fastapi.testclient import TestClient
+from fsspec.implementations.memory import MemoryFileSystem
+from loguru import logger
 from lumigator_schemas.jobs import JobConfig, JobType
 from mypy_boto3_s3 import S3Client
 from s3fs import S3FileSystem
 from sqlalchemy import Engine, create_engine
 from sqlalchemy.orm import Session
-from testcontainers.localstack import LocalStackContainer
 
-from backend.api.deps import get_db_session, get_s3_client
+from backend.api.deps import get_db_session, get_s3_client, get_s3_filesystem
 from backend.api.router import API_V1_PREFIX
 from backend.main import create_app
-from backend.records.jobs import JobRecord, JobResultRecord
+from backend.records.jobs import JobRecord
 from backend.repositories.datasets import DatasetRepository
-from backend.repositories.jobs import BaseRepository, JobRepository, JobResultRepository
+from backend.repositories.jobs import JobRepository, JobResultRepository
 from backend.services.datasets import DatasetService
 from backend.services.jobs import JobService
 from backend.settings import BackendSettings, settings
-from backend.tests.fakes.fake_ray_client import FakeJobSubmissionClient
+from backend.tests.fakes.fake_s3 import FakeS3Client
 
-# TODO: Break tests into "unit" and "integration" folders based on fixture dependencies
+TEST_CAUSAL_MODEL = "hf://hf-internal-testing/tiny-random-LlamaForCausalLM"
+TEST_INFER_MODEL = "hf://hf-internal-testing/tiny-random-t5"
 
 
 def common_resources_dir() -> Path:
@@ -44,6 +48,7 @@ def format_dataset(data: list[list[str]]) -> str:
 
 @pytest.fixture
 def valid_experiment_dataset() -> str:
+    """Minimal valid dataset with groundtruth."""
     data = [
         ["examples", "ground_truth"],
         ["Hello World", "Hello"],
@@ -53,6 +58,7 @@ def valid_experiment_dataset() -> str:
 
 @pytest.fixture(scope="session")
 def valid_experiment_dataset_without_gt() -> str:
+    """Minimal valid dataset without groundtruth."""
     data = [
         ["examples"],
         ["Hello World"],
@@ -62,6 +68,7 @@ def valid_experiment_dataset_without_gt() -> str:
 
 @pytest.fixture(scope="session")
 def missing_examples_dataset() -> str:
+    """Minimal invalid dataset without examples."""
     data = [
         ["ground_truth"],
         ["Hello"],
@@ -71,6 +78,7 @@ def missing_examples_dataset() -> str:
 
 @pytest.fixture(scope="session")
 def extra_column_dataset() -> str:
+    """Minimal valid dataset with groundtruth and extra fields."""
     data = [
         ["examples", "ground_truth", "extra"],
         ["Hello World", "Hello", "Nope"],
@@ -109,40 +117,47 @@ def db_session(db_engine: Engine):
             session.rollback()
 
 
-@pytest.fixture(scope="session")
-def localstack_container():
-    """Initialize a LocalStack test container."""
-    with LocalStackContainer("localstack/localstack:3.4.0") as localstack:
-        yield localstack
-
-
-@pytest.fixture(scope="session", autouse=True)
-def setup_aws(localstack_container: LocalStackContainer):
-    """Setup env vars/AWS resources for use with the LocalStack container."""
-    # Initialize S3
-    s3 = localstack_container.get_client("s3")
-    try:
-        s3.create_bucket(
-            Bucket=settings.S3_BUCKET,
-            CreateBucketConfiguration={"LocationConstraint": localstack_container.region_name},
-        )
-    except ClientError as e:
-        if e.response["Error"]["Code"] == "BucketAlreadyOwnedByYou":
-            print("Bucket already created, continue")
-        else:
-            raise e
-    # add ENV vars for FSSPEC access to S3 (s3fs + HuggingFace datasets)
-    os.environ["FSSPEC_S3_KEY"] = "testcontainers-localstack"
-    os.environ["FSSPEC_S3_SECRET"] = "testcontainers-localstack"  # pragma: allowlist secret
-    os.environ["FSSPEC_S3_ENDPOINT_URL"] = localstack_container.get_url()
-    os.environ["AWS_ACCESS_KEY_ID"] = "testcontainers-localstack"
-    os.environ["AWS_SECRET_ACCESS_KEY"] = "testcontainers-localstack"  # pragma: allowlist secret
-    os.environ["AWS_ENDPOINT_URL"] = localstack_container.get_url()
+@pytest.fixture(scope="function")
+def fake_s3fs() -> S3FileSystem:
+    """Replace the filesystem registry for S3 with a MemoryFileSystem implementation."""
+    fsspec.register_implementation(
+        "s3", MemoryFileSystem, clobber=True, errtxt="Failed to register mock S3FS"
+    )
+    yield MemoryFileSystem()
+    logger.info(f"final s3fs contents: {str(MemoryFileSystem.store)}")
 
 
 @pytest.fixture(scope="function")
-def s3_client(localstack_container: LocalStackContainer) -> S3Client:
-    return localstack_container.get_client("s3")
+def fake_s3_client(fake_s3fs) -> S3Client:
+    """Provide a fake S3 client using MemoryFileSystem as underlying storage."""
+    os.environ["AWS_ACCESS_KEY_ID"] = "test"
+    # Please check https://github.com/localstack/localstack/issues/5894
+    # for info about the test region used
+    os.environ["AWS_DEFAULT_REGION"] = "us-east-1"
+    os.environ["AWS_SECRET_ACCESS_KEY"] = "test"  # pragma: allowlist secret
+    os.environ["AWS_ENDPOINT_URL"] = "http://example.com:4566"
+    return FakeS3Client(MemoryFileSystem.store)
+
+
+@pytest.fixture(scope="function")
+def boto_s3_client() -> S3Client:
+    """Provide a real S3 client."""
+    os.environ["AWS_ACCESS_KEY_ID"] = "test"
+    # Please check https://github.com/localstack/localstack/issues/5894
+    # for info about the test region used
+    os.environ["AWS_DEFAULT_REGION"] = "us-east-1"
+    os.environ["AWS_SECRET_ACCESS_KEY"] = "test"  # pragma: allowlist secret
+    os.environ["AWS_ENDPOINT_URL"] = "http://localhost:4566"
+    return boto3.client("s3")
+
+
+@pytest.fixture(scope="function")
+def boto_s3fs() -> S3FileSystem:
+    """Provide a real s3fs client."""
+    s3fs = S3FileSystem()
+    mock_s3fs = Mock(wraps=s3fs)
+    yield mock_s3fs
+    logger.info(f"intercepted s3fs calls: {str(mock_s3fs.mock_calls)}")
 
 
 @pytest.fixture(scope="session")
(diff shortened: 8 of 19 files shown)


lumigator/python/mzai/backend/backend/tests/conftest.py
```diff
@@ -107,40 +111,79 @@ def db_session(db_engine: Engine):
             session.rollback()
 
 
-@pytest.fixture(scope="session")
-def localstack_container():
-    """Initialize a LocalStack test container."""
-    with LocalStackContainer("localstack/localstack:3.4.0") as localstack:
-        yield localstack
+# @pytest.fixture(scope="session")
```
veekaybee, line 114: we should remove if not being used

lumigator/python/mzai/backend/backend/tests/conftest.py
```diff
@@ -107,40 +111,79 @@ def db_session(db_engine: Engine):
...
+#     with LocalStackContainer("localstack/localstack:3.4.0") as localstack:
+#         yield localstack
 
 
-@pytest.fixture(scope="session", autouse=True)
-def setup_aws(localstack_container: LocalStackContainer):
-    """Setup env vars/AWS resources for use with the LocalStack container."""
+@pytest.fixture(scope="function")
+def fake_s3_storage() -> dict:
+    return dict()
```
veekaybee, line 123: should be just `return dict`, we're not doing a method call here, and if we are not doing anything specific to the dict itself, we can call a generic dict without the function in our tests

lumigator/python/mzai/backend/backend/tests/conftest.py
```diff
@@ -107,40 +111,79 @@ def db_session(db_engine: Engine):
...
-    os.environ["AWS_ENDPOINT_URL"] = localstack_container.get_url()
+    os.environ["AWS_ACCESS_KEY_ID"] = "test"
+    # Please check https://github.com/localstack/localstack/issues/5894
+    # for info about the test region used
+    os.environ["AWS_DEFAULT_REGION"] = "us-east-1"
+    os.environ["AWS_SECRET_ACCESS_KEY"] = "test"  # pragma: allowlist secret
+    os.environ["AWS_ENDPOINT_URL"] = "http://localhost:4566"
+    return boto3.client("s3")
+
+# @pytest.fixture(scope="function")
```
veekaybee, line 146: remove?

lumigator/python/mzai/backend/README.md
```diff
@@ -36,7 +36,7 @@ The tests include a unit test suite and an integration test suite. There are mak
   * `backend-unit-test`: runs tests in `backend/tests/unit/*/test_*.py` (any depth of subfolders)
   * `backend-int-test`: runs tests in `backend/tests/int/*/test_*.py` (any depth of subfolders)
 
-The SQLite configuration making use of a local file is used in both unit and integration tests. Test containers are started for integration tests.
+The SQLite configuration making use of a local file is used in both unit and integration tests. Test containers are needed for integration tests. Fake or mock services are used in unit tests, usually via the [`requests-mock`](https://pypi.org/project/requests-mock/) package in the case of HTTP APIs. Specifically, the S3 file system driver is replaced with an [in-memory implementation](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.implementations.memory.MemoryFileSystem) of the `AbstractFileSystem` interface via [registry modification](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.registry.register_implementation).
```
veekaybee, line 39: We could use Moto's fake S3 here with the `@mock_aws` decorator, wdyt? https://github.com/getmoto/moto?tab=readme-ov-file#in-a-nutshell

lumigator/python/mzai/backend/backend/services/datasets.py
```diff
@@ -119,6 +121,7 @@ def _save_dataset_to_s3(self, temp_fname, record):
             # Upload to S3
             dataset_key = self._get_s3_key(record.id, record.filename)
             dataset_path = self._get_s3_path(dataset_key)
+            # Deprecated!!!
```
veekaybee, line 124: should we still use if deprectated?

lumigator/python/mzai/backend/backend/services/datasets.py
```diff
@@ -94,6 +95,7 @@ def _raise_not_found(self, dataset_id: UUID) -> None:
         raise HTTPException(status.HTTP_404_NOT_FOUND, f"Dataset '{dataset_id}' not found.")
 
     def _raise_unhandled_exception(self, e: Exception) -> None:
+        traceback.print_exception(type(e), e, e.__traceback__)
```
veekaybee, line 98: we can simplify this with `traceback.print_exc()`
https://docs.python.org/3/library/traceback.html#traceback.print_exc

lumigator/python/mzai/backend/backend/tests/conftest.py
```diff
@@ -107,40 +111,79 @@ def db_session(db_engine: Engine):
...
 
 
-@pytest.fixture(scope="session", autouse=True)
-def setup_aws(localstack_container: LocalStackContainer):
-    """Setup env vars/AWS resources for use with the LocalStack container."""
+@pytest.fixture(scope="function")
+def fake_s3_storage() -> dict:
+    return dict()
+
+@pytest.fixture(scope="function")
```
veekaybee, line 125: in the future we should be able to simplify this by removing a direct S3 client call and passing info into the S3FS mock directly https://github.com/mozilla-ai/lumigator/issues/473

lumigator/python/mzai/backend/backend/tests/conftest.py
veekaybee: I'd recommend making this field part of the `fake_s3fs(fake_s3_storage)` class

lumigator/python/mzai/backend/backend/tests/conftest.py
```diff
@@ -107,40 +111,79 @@ def db_session(db_engine: Engine):
...
-def s3_client(localstack_container: LocalStackContainer) -> S3Client:
-    return localstack_container.get_client("s3")
+def boto_s3fs() -> S3FileSystem:
+    s3fs = S3FileSystem()
+    mock_s3fs = Mock(wraps=s3fs)
+    yield mock_s3fs
+    print(f'---> s3fs: {str(mock_s3fs.mock_calls)}')
+
+
+# @pytest.fixture(scope="session")
```
veekaybee, line 174: remove?

lumigator/python/mzai/backend/backend/tests/conftest.py
```diff
@@ -238,11 +307,12 @@ def fake_ray_client():
     return FakeJobSubmissionClient()
 
 
+# FIXME investigate where this is used
```
veekaybee, line 310: do we still need this comment?

lumigator/python/mzai/backend/backend/tests/conftest.py
```diff
@@ -186,10 +229,36 @@ def get_db_session_override():
         yield db_session
 
     def get_s3_client_override():
-        yield s3_client
+        yield fake_s3_client
+
+    def get_s3_filesystem_override():
+        # yield boto_s3fs
```
veekaybee, line 235: remove?

lumigator/python/mzai/backend/backend/tests/conftest.py
```diff
@@ -175,8 +218,8 @@ def local_client(app: FastAPI):
         yield c
 
 
-@pytest.fixture(scope="function", autouse=True)
-def dependency_overrides(app: FastAPI, db_session: Session, s3_client: S3Client) -> None:
+@pytest.fixture(scope="function")
+def dependency_overrides_fakes(app: FastAPI, db_session: Session, fake_s3_client: S3Client, fake_s3fs: S3FileSystem) -> None:
```
veekaybee, line 222: i think this method is duplicated

lumigator/python/mzai/backend/backend/tests/conftest.py
```diff
@@ -107,40 +111,79 @@ def db_session(db_engine: Engine):
...
+    os.environ["AWS_ACCESS_KEY_ID"] = "test"
+    # Please check https://github.com/localstack/localstack/issues/5894
+    # for info about the test region used
+    os.environ["AWS_DEFAULT_REGION"] = "us-east-1"
+    os.environ["AWS_SECRET_ACCESS_KEY"] = "test"  # pragma: allowlist secret
+    os.environ["AWS_ENDPOINT_URL"] = "http://example.com:4566"
+    return FakeS3Client(fake_s3_storage)
+
+@pytest.fixture(scope="function")
+def boto_s3_client(fake_s3_storage) -> S3Client:
```
veekaybee, line 136: can we remove boto?

lumigator/python/mzai/backend/backend/tests/conftest.py
```diff
@@ -24,6 +27,7 @@
 from backend.services.jobs import JobService
 from backend.settings import BackendSettings, settings
 from backend.tests.fakes.fake_ray_client import FakeJobSubmissionClient
+from backend.tests.fakes.fake_s3 import FakeS3Client
 
 # TODO: Break tests into "unit" and "integration" folders based on fixture dependencies
```
veekaybee, line 32: we can remove this comment now

lumigator/python/mzai/backend/backend/tests/conftest.py
```diff
@@ -107,40 +111,79 @@ def db_session(db_engine: Engine):
...
+#     mock_client = Mock(wraps=client)
+#     yield mock_client
+#     print(f'---> s3client: {str(mock_client.mock_calls)}')
+
+@pytest.fixture(scope="function")
+def fake_s3fs(fake_s3_storage) -> S3FileSystem:
+    # ...and patch the s3fs name with the Fake
+    fsspec.register_implementation("s3", MemoryFileSystem, clobber=True, errtxt="Failed to register mock S3FS")
+    # fsspec.register_implementation("s3", FakeS3FS, clobber=True, errtxt="Failed to register mock S3FS")
+    # FakeS3FS.storage = fake_s3_storage
```
veekaybee, line 158: remove comments?

lumigator/python/mzai/backend/backend/tests/conftest.py
javiermtorres: We'd be returning a fresh dict here ({} would be clearer, probably).
In any case, it was meant to provide a common storage for the S3 client and the S3FS. But we can just get `MemoryFileSystem.store` and use it. I'll try that out.

lumigator/python/mzai/backend/backend/tests/conftest.py
javiermtorres: Most probably. If it's used, it will only be in the integration tests, but the app should use its own boto client anyway.

lumigator/python/mzai/backend/backend/tests/conftest.py
javiermtorres: Yep 👍

lumigator/python/mzai/backend/backend/tests/conftest.py
javiermtorres: Totally. I left them just in case the memfs didn't work, since it seems to work ok, I'll just remove it.

lumigator/python/mzai/backend/backend/tests/conftest.py
javiermtorres: Ah wait. It's plugged as a dependency in the app. Right now it has to be there for integration tests.

lumigator/python/mzai/backend/README.md
javiermtorres: As mentioned, moto doesn't seem to play well with aiobotocore :-/ But I'll make a test just in case I'm using it in a wrong way.

lumigator/python/mzai/backend/backend/services/datasets.py
javiermtorres: Thanks!!!!! ^_^

lumigator/python/mzai/backend/backend/services/datasets.py
javiermtorres: Since the scope was just replacing the mocks, and I didn't know if we used any special features, I just left a note :-/ 
Shall we add it to https://github.com/mozilla-ai/lumigator/issues/473 instead?

lumigator/python/mzai/backend/backend/tests/conftest.py
javiermtorres: There should be one for fakes and another one for the real services, but I may have not pushed that change :( I'll make sure it's correct in the next push 👍

lumigator/python/mzai/backend/backend/tests/conftest.py
javiermtorres: Hum. I gave it a quick look, and it seems that the aim was to use it to fake Ray calls. But @peteski22  already uses `requests-mock` for this, so I'll try to remove this fake altogether.

aittalam:
> @aittalam @dpoulopoulos I'm reverting the background task patching in
> 
> https://github.com/mozilla-ai/lumigator/blob/d3c5abee2e50d5067e059c90128ef6e3edfcdf65/lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py#L50
> 
> The tests seem to run ok (I added a couple of extra log lines to be sure), so let's just keep the reference to this patching in case you notice the tests stalling again.

I think it'd be better to keep the mock and test the called task separately.
The main reason is that in those tests we are not evaluating whether the job completes but just that the system returns an ok. If we were to actually start the background task we'd have to wait for the job completion to test anything.
What is usually suggested in these cases is to mock the bgtask and test the method separately, which is something I will add.

.github/workflows/test_backend_uv.yaml
```diff
@@ -32,3 +33,16 @@ jobs:
...
       - name: Run backend integration tests
         run: AUTO_TEST_RUN=y make test-backend-integration
+
+      - name: Collect Ray logs in case of failure
+        if: failure()
+        run: |
+          mkdir /tmp/raylogs
+          docker cp lumigator-ray-1:/tmp/ray/ /tmp/raylogs/
+
+      - name: Upload Ray logs in case of failure
```
veekaybee, line 43: what are we thinking of doing with these logs? right now they just go to the test runner, i'd avoid in this case since it slows down tests

Makefile
```diff
@@ -118,12 +117,15 @@ else
 endif
 
 test-backend-unit:
-	cd lumigator/python/mzai/backend/backend/tests; \
+	cd lumigator/python/mzai/backend/; \
+	rm local.db; \
```
veekaybee, line 121: will this break folks' current setup? we might want to add some warning docs here

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -36,6 +36,7 @@
 
 
 class JobService:
+    pass
```
veekaybee, line 39: why did we add this?

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -173,54 +174,56 @@ def _get_job_params(self, job_type: str, record, request: BaseModel) -> dict:
...
+            s3 = S3FileSystem()
+
+            # Get the dataset from the S3 bucket
+            result_key = self._get_results_s3_key(job_id)
+            with s3.open(f"{settings.S3_BUCKET}/{result_key}", "r") as f:
+                results = json.loads(f.read())
+
+            dataset = {k: v for k, v in results.items() if k in ["examples", request.output_field]}
+
+            # Create a CSV in memory
```
veekaybee, line 187: Not sure I understand this comment, we either create a file artifact or work with the code in-memory

.github/workflows/test_backend_uv.yaml
javiermtorres: This only runs in case of failure. It is very useful to debug obscure issues like the "no space left on container" one. Currently it's about 100kb zipped, and kept with the test run.

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -343,7 +349,7 @@ def list_jobs(
         records = self.job_repo.list(skip, limit)
         return ListingResponse(
             total=total,
-            items=[JobResponse.model_validate(x) for x in records],
+            items=[self.get_job(x.id) for x in records],
```
veekaybee, line 352:
```suggestion
            items=[self.get_job(record.id) for record in records],
````

File: Makefile (hunk 16 above)
Comment (javiermtorres) [reply Makefile:121]: I'll add a warning, but basically one shouldn't reuse the local db unless one knows what they're doing. It actually causes issues because people tend to forget removing this before running tests, right now.

File: lumigator/python/mzai/backend/backend/services/jobs.py (hunk 17 above)
Comment (javiermtorres) [reply lumigator/python/mzai/backend/backend/services/jobs.py:39]: Sorry!!!! It was a trigger for the tests. I've just removed it.

File: lumigator/python/mzai/backend/backend/tests/conftest.py [hunk 20]
```diff
@@ -107,40 +107,45 @@ def db_session(db_engine: Engine):
...
 
 
-@pytest.fixture(scope="session")
-def localstack_container():
-    """Initialize a LocalStack test container."""
-    with LocalStackContainer("localstack/localstack:3.4.0") as localstack:
-        yield localstack
+@pytest.fixture(scope="function")
+def fake_s3fs() -> S3FileSystem:
+    # ...and patch the s3fs name with the Fake
```
veekaybee, line 112: let's make sure method comments follow docstring syntax like on the left

lumigator/python/mzai/backend/backend/services/jobs.py
javiermtorres: I didn't write this comment or this code... But I understand that we use a StringIO as a "file-like object" to provide a way to transform a file using the csv package, so it's a file to python code, that is backed by a byte array in memory (and not a data structure). Maybe @aittalam has more info???

lumigator/python/mzai/backend/backend/tests/conftest.py
javiermtorres: Ok, I'll add some comments on the fixtures.

lumigator/python/mzai/backend/backend/config_templates.py
```diff
@@ -101,6 +101,17 @@
     "name": "{job_name}/{job_id}",
     "model": {{ "path": "{model_uri}" }},
     "dataset": {{ "path": "{dataset_path}" }},
+    "job": {{
```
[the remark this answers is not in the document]
javiermtorres: IIRC this info was missing since it was needed for the test inference model we use in integration testing. Should it be removed?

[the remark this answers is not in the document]
javiermtorres: I've removed it and tests seem to pass (also, tests use a custom template right now).

javiermtorres:
> I think it'd be better to keep the mock and test the called task separately. The main reason is that in those tests we are not evaluating whether the job completes but just that the system returns an ok.

Since test models are used in integration, actually the jobs are checked to completion (for example see https://github.com/mozilla-ai/lumigator/blob/236532695340a072a55c74c7bb3257cbfb2d7b70/lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py#L64). Currently the integration tests run in 5-6 mins but make sure that all the steps of a job are properly executed. Remember that some bugs were hidden because we didn't check the jobs to completion.

lumigator/python/mzai/backend/backend/services/jobs.py
aittalam: This is from https://github.com/mozilla-ai/lumigator/pull/522, the main idea is the following:

- we have inference results which are saved by the job as a json file (actual data + metadata)
- we need to store the data part (in particular samples and predictions - I added GT in my follow-up PR) as a dataset that will be ingested by evaluation
- at the moment, only CSV files can be brought on the platform using `DataService.upload_dataset()`, so the data we need is converted into an in-memory CSV and then uploaded

veekaybee: Ah sorry I missed that this was not new, thanks for the explanation! So the actual StringIO object is a Python string object that can be passed into csv.writer, which takes any Python object that has a File API, makes sense.

lumigator/python/mzai/backend/backend/tests/conftest.py
veekaybee: Can we remove this comment if not used anymore?

lumigator/python/mzai/backend/backend/tests/fakes/fake_s3.py
```diff
@@ -0,0 +1,51 @@
...
+            'Key': key,
+            'LastModified': mod_date,
+            'Size': len(entry.getvalue()),
+            'StorageClass': 'STANDARD',
+        }
+        return content
+
+    def list_objects_v2(self, **kwargs):
+        bucket = kwargs["Bucket"]
+        # bucket = 'lumigator-storage'
```
veekaybee, line 32: remove comment

lumigator/python/mzai/backend/backend/tests/fakes/fake_s3.py
```diff
@@ -0,0 +1,51 @@
...
+        prefix = kwargs["Prefix"]
+        # prefix = 'datasets/d35c1eb7-db82-4a43-9881-180bed9d9808/dataset.csv'
```
veekaybee, line 34: remove comment

.github/workflows/test_backend_uv.yaml
```diff
@@ -32,3 +33,16 @@ jobs:
 
       - name: Run backend integration tests
         run: AUTO_TEST_RUN=y make test-backend-integration
```
javiermtorres, line 35: Sometimes jobs even fail to start. In those cases, it's useful to collect the logs from Ray itself. Only the `/tmp/ray/session_*/logs/` directory is needed, otherwise we could be fetching the entire Python distro in Ray.

Makefile
```diff
@@ -118,12 +117,15 @@ else
...
 	SQLALCHEMY_DATABASE_URL=sqlite:///local.db uv run pytest -o python_files="backend/tests/unit/*/test_*.py"
 
 test-backend-integration-target:
-	cd lumigator/python/mzai/backend/backend/tests;	\
-	SQLALCHEMY_DATABASE_URL=sqlite:///local.db uv run pytest -o python_files="backend/tests/integration/*/test_*.py"
+	cd lumigator/python/mzai/backend/; \
+	docker container list --all; \
+	rm local.db; \
+	SQLALCHEMY_DATABASE_URL=sqlite:///local.db RAY_WORKER_GPUS="0.0" RAY_WORKER_GPUS_FRACTION="0.0" INFERENCE_PIP_REQS=../jobs/inference/requirements_cpu.txt INFERENCE_WORK_DIR=../jobs/inference EVALUATOR_PIP_REQS=../jobs/evaluator/requirements.txt EVALUATOR_WORK_DIR=../jobs/evaluator uv run pytest -s -o python_files="backend/tests/integration/*/test_*.py"
```
javiermtorres, line 128: To avoid having pending jobs in Ray, we need to indicate that our test jobs will not use GPU. The custom inference pip requirements also avoid heavy libraries to be loaded into the GH CI runner, causing the job to fail due to disk space exhaustion.

lumigator/python/mzai/backend/backend/config_templates.py
```diff
@@ -100,7 +100,7 @@
 causal_infer_template = """{{
     "name": "{job_name}/{job_id}",
     "model": {{ "path": "{model_uri}" }},
-    "dataset": {{ "path": "{dataset_path}" }},
+    "dataset": {{ "path": "{dataset_path}" }}
```
javiermtorres, line 103: Wrong json format.

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -173,54 +173,56 @@ def _get_job_params(self, job_type: str, record, request: BaseModel) -> dict:
...
-            format=DatasetFormat.JOB,
-            run_id=job_id,
-            generated=True,
-            generated_by=results["model"],
-        )
-
-        loguru.logger.info(
-            f"Dataset '{dataset_filename}' with ID '{dataset_record.id}' added to the database."
-        )
+        try:
```
javiermtorres, line 176: This section should be removed. The except was added as part of troubleshooting, but it was not later removed. There is no recovery to be done at this stage.

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -173,54 +173,56 @@ def _get_job_params(self, job_type: str, record, request: BaseModel) -> dict:
...
+            )
+        except Exception:
+            loguru.logger.error(f"Request failed for job {job_id} as {request}")
 
     def _watch_job(self, job_id: UUID, request: JobEvalCreate | JobInferenceCreate):
         job_status = self.ray_client.get_job_status(job_id)
         job_info = self.ray_client.get_job_info(job_id)
         job_metadata = job_info.metadata
         job_type = job_metadata["job_type"]
+        loguru.logger.info(f"Starting job {job_id} check with {job_status.lower()}")
```
javiermtorres, line 225: Some or all of the logs should maybe be removed.

lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
```diff
@@ -30,38 +38,101 @@ def test_upload_data_launch_job(local_client: TestClient, dialog_dataset):
...
     headers = {
         "accept": "application/json",
         "Content-Type": "application/json",
     }
-    payload = {
+    eval_payload = {
         "name": "test_run_hugging_face",
         "description": "Test run for Huggingface model",
-        "model": "hf://facebook/bart-large-cnn",
+        "model": "hf://hf-internal-testing/tiny-random-LlamaForCausalLM",
```
javiermtorres, line 49: Random model to be used in integration test. Does not return useful inferences, but otherwise behaves as a valid model.

lumigator/python/mzai/jobs/inference/requirements_cpu.txt
```diff
@@ -0,0 +1,11 @@
+--extra-index-url https://download.pytorch.org/whl/cpu
```
javiermtorres, line 1: Custom pip requirements without GPU libraries

lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
```diff
@@ -30,38 +38,101 @@ def test_upload_data_launch_job(local_client: TestClient, dialog_dataset):
...
 
-    # Mocking the background task because the test hangs indefinitely,
-    # waiting for the background task to finish.
-    with patch("fastapi.BackgroundTasks.add_task") as mock:
-        create_evaluation_job_response = local_client.post(
-            "/jobs/evaluate/", headers=headers, json=payload
-        )
-        assert create_evaluation_job_response.status_code == 201
-        assert mock.called
+    create_evaluation_job_response = local_client.post(
```
javiermtorres, line 55: Verifies that jobs are started and run to success.

Makefile
aittalam, line 121: The `local.db` file is now something used for tests only, is it correct?
I understand that here we want to make sure the test runs without breaking, and I am ok removing it before tests start, but if it is functional to tests only I think that we should also remove it at the end of the test, and not push it to the repo anymore. WDYT?

Makefile
```diff
@@ -16,7 +16,6 @@ endif
 define run_with_containers
 	@echo "No Lumigator containers are running. Starting containers..."
 	make start-lumigator-build
-	trap "cd $(PROJECT_ROOT); make stop-lumigator" EXIT; \
```
aittalam, line 19: What is the rationale for removing this line? I am asking because, if I recall correctly, it fixed [this issue](https://github.com/mozilla-ai/lumigator/issues/326) and I wonder whether this would introduce a regression

lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
aittalam: Question: I see this used in different places below, do you think it'd make sense to have its name stored somewhere (e.g. in `settings.py`) and then load it from there?

lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
```diff
@@ -78,19 +149,15 @@ def test_full_experiment_launch(local_client: TestClient, dialog_dataset):
     payload = {
         "name": "test_run_hugging_face",
         "description": "Test run for Huggingface model",
-        "model": "hf://facebook/bart-large-cnn",
+        "model": "hf://hf-internal-testing/tiny-random-LlamaForCausalLM",
```
aittalam, line 152: see above 🙏

lumigator/python/mzai/jobs/inference/requirements_cpu.txt
aittalam: I think it's great that we have a cpu-only alternative! Not just for tests actually, I think we could open an issue to try and generalize this, wdyt?

lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
javiermtorres: I'll probably need some help from you about the specific model :-/ I'm still making out the classes of different models set up by HF (seq2seq, causal, etc) and which would be the lightest alternative for each of them. Since it's used in a few tests, yes, I guess it makes sense to keep it somewhere centralized (I'd bet on `conftest.py` though).

lumigator/python/mzai/jobs/inference/requirements_cpu.txt
javiermtorres: Fine for me! I'd say this could be set up per job, maybe by prefixing this to the existing requirements (possibly allowing the requirements to go into the job request is maybe too open at this stage). Right now it's set using an env var, so it's not really an option for normal use since it'll affect all job submissions :-/

Makefile
javiermtorres: I needed to keep the containers running after a failed test to retrieve the Ray logs for debugging :-/ Since the GH steps are launched after the test step is completed, we cannot stop them in CI. Maybe we can use an env var to control whether to stop the containers at the end or not, stopping them by default but keeping them up for debugging in CI?

lumigator/python/mzai/backend/backend/config_templates.py
aittalam: Thanks for catching it!

lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
aittalam: Top, thanks! Tentatively, the seq2seq should be lighter but I am happy to try some of these models together.

Makefile
aittalam: Sounds good to me!

Makefile
aittalam: I understand, thanks for clarifying! If I got it right, this is for debugging when things break, not locally (as we can run tests on pre-loaded containers there) but specifically in CI.

How often does this happen? I am asking to better understand whether it makes more sense to specifically support a solution like the one you suggested (i.e. will it be a useful tool for developers who want to contribute to lumigator) or just remove that line (and suggest to do this in a FAQ) whenever that kind of issue happens.

What concerns me about the first solution is legibility of our Makefile (but we can make this easier with good documentation), about the latter is that someone will forget to add that line again before merging. This makes me lean more towards the former, wdyt?

javiermtorres: Yes, it's cleaner imo to have it with a switch (env var) than removing it altogether. I was just completely lost as to the source of the original error, hence the removal :D

aittalam: Cool! Btw there's no need to add this extra functionality in this PR, if you want to do it in a separate one I am happy to review it :-)

aittalam (approved):
Thanks @javiermtorres for the PR and for addressing my comments, it looks good to me 🙏

lumigator/python/mzai/backend/backend/services/datasets.py
```diff
@@ -89,11 +90,13 @@ def __init__(
         self.dataset_repo = dataset_repo
         self.s3_client = s3_client
         self.s3_filesystem = s3_filesystem
+        print(self.s3_filesystem)
```
veekaybee, line 93: remove print log

lumigator/python/mzai/backend/backend/services/datasets.py
```diff
@@ -242,7 +248,9 @@ def get_dataset_download(self, dataset_id: UUID) -> DatasetDownloadResponse:
                 )
 
             download_urls = []
+            print(f'---> returned {s3_response}')
```
veekaybee, line 251: remove print log (or alternatively switch to a logger)

lumigator/python/mzai/backend/backend/services/datasets.py
```diff
@@ -242,7 +248,9 @@ def get_dataset_download(self, dataset_id: UUID) -> DatasetDownloadResponse:
...
             for s3_object in s3_response["Contents"]:
+                print(f'  ---> returned object {s3_object}')
```
veekaybee, line 253: remove print log (or alternatively switch to a logger)

lumigator/python/mzai/backend/backend/tests/conftest.py
```diff
@@ -109,40 +116,47 @@ def db_session(db_engine: Engine):
...
+    return boto3.client("s3")
+
+
+@pytest.fixture(scope="function")
+def boto_s3fs() -> S3FileSystem:
+    """Provide a real s3fs client."""
+    s3fs = S3FileSystem()
+    mock_s3fs = Mock(wraps=s3fs)
+    yield mock_s3fs
+    print(f"intercepted s3fs calls: {str(mock_s3fs.mock_calls)}")
```
veekaybee, line 159: remove print log (or alternatively switch to a logger)

lumigator/python/mzai/backend/backend/tests/conftest.py
```diff
@@ -177,9 +191,34 @@ def local_client(app: FastAPI):
...
-@pytest.fixture(scope="function", autouse=True)
-def dependency_overrides(app: FastAPI, db_session: Session, s3_client: S3Client) -> None:
-    """Override the FastAPI dependency injection for test DB sessions.
+@pytest.fixture(scope="function")
+def dependency_overrides_fakes(
+    app: FastAPI, db_session: Session, fake_s3_client: S3Client, fake_s3fs: S3FileSystem
+) -> None:
+    """Override the FastAPI dependency injection for test DB sessions. Uses mocks/fakes for unit tests.
+
+    Reference: https://fastapi.tiangolo.com/he/advanced/testing-database/
```
veekaybee, line 200: broken link I think

lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
```diff
@@ -24,43 +36,113 @@ def test_upload_data_launch_job(local_client: TestClient, dialog_dataset):
...
+    local_client: TestClient, dialog_dataset, dependency_overrides_services
+):
     response = local_client.get("/health")
     assert response.status_code == 200
     create_response = local_client.post(
         "/datasets/",
         data={},
         files={"dataset": dialog_dataset, "format": (None, DatasetFormat.JOB.value)},
     )
+    print(f"response: {create_response.text}")
```
veekaybee, line 145: remove print/or use logger

veekaybee (approved):
Approved after addressing logging comments, thanks!

javiermtorres:
Argh @veekaybee I believe some of the logs crept in after one of the rebases :-/ I'll fix them asap.

lumigator/python/mzai/backend/backend/services/datasets.py
javiermtorres: Sorry. There are a few troubleshooting logs here that shouldn't be :-/

lumigator/python/mzai/backend/backend/tests/conftest.py
javiermtorres: Moved to logger

lumigator/python/mzai/backend/backend/tests/conftest.py
javiermtorres: Updated

.env.example
```diff
@@ -1,4 +1,11 @@
 # .env.example
+
+# Lumigator container control
+# Set to "TRUE" if the containers need to be up and running after
+# a test target failed (e.g. in CI where containers are inspected
+# for logs after failed steps)
+CONTAINERS_UP_ON_ERROR="FALSE"
```
peteski22, line 7:
```suggestion
KEEP_CONTAINERS_UP="FALSE"
```

PR #477, opened 2024-12-06
339 support seq2seq models

### LINKED ISSUE #339: Add support for seq2seq models with transformers or vLLM as-a-library
This inference mode calls on hf transformers to run the model, rather than sending a request to an external API.

For context, users will likely want to compare models they can get from HF model hub, for example. This means running the model ad-hoc for a task, rather than sending a request to an external API.

dpoulopoulos:
# What's changing

Introduces the `HuggingFaceModelClient` to support models pulled from Hugging Face Hub.

The client expects a model and a task, and instantiates the corresponding pipeline. For example, if the task is "summarization" the client instantiates the following pipeline:

```python
from transformers import pipeline

pipe = pipeline("summarization", model="...", **kwargs)
```

During the instantiation of the pipeline we consider several things, such as the `dtype` for the model weights and the `device` we will use during inference (i.e., `cpu`, `cuda`, or `mps`).

Closes #339

# How to test it

Steps to test the changes:

1. Run `make local-up` to deploy Lumigator locally.
1. Upload a dataset using the OpenAPI docs page (`localhost:8000/docs`)
1. Create a new inference job, with the following body:
    ```json
    {
      "name": "summarization",
      "model": "hf://facebook/nbart-large-cnn",
      "dataset": "...",
      "max_samples": 1,
      "task": "summarization",
      "accelerator": "auto",
      "revision": "main",
      "use_fast": true,
      "trust_remote_code": false,
      "torch_dtype": "auto"
    }
    ```

    for translation tasks provide the following body:
    
    ```json
    {
      "name": "translation",
      "model": "hf://google-t5/t5-base",
      "dataset": "...",
      "max_samples": 1,
      "task": "translation",
      "accelerator": "auto",
      "revision": "main",
      "use_fast": true,
      "trust_remote_code": false,
      "torch_dtype": "auto"
    }
    ```

In theory, this implementation can support any arbitrary task, but we're limited by the way we're loading the dataset. For example, "question-answering" tasks need the dataset to be in the following format:

```python
qa_input = {
    'question': 'Why is model conversion important?',
    'context': 'The option to convert models between FARM and transformers gives freedom to the user and let people easily switch between frameworks.'
}
```

We treat this as a known limitation.

# I already...


lumigator/python/mzai/jobs/inference/model_clients.py
```diff
@@ -145,3 +146,31 @@ def _chat_completion(
             temperature=config.params.temperature,
             top_p=config.params.top_p,
         )
+
+
+class HuggingFaceModelClient(BaseModelClient):
+    def __init__(self, config: InferenceJobConfig):
```
javiermtorres, line 152: Maybe declare the types for _config and _pipeline here, specially _pipeline since it's a method in disguise :)

dpoulopoulos: Thanks Javier! Let's address than in the final version of the code.

docs/source/user-guides/inference.md
```diff
@@ -0,0 +1,110 @@
+# Running an Inference Job
+
+This guide will walk you through the process of running an inference job using the Lumigator SDK and
+a model downloaded from the Hugging Face Hub. The model will generate summaries for a given set of
+text data.
+
+```{note}
+You can also use the OpenAI GPT family of models or the Mistal API to run an inference job. To do
+so, you need to set the appropriate environment variables: `OPENAI_API_KEY` or `MISTAL_API_KEY`.
```
javiermtorres, line 9:
```suggestion
so, you need to set the appropriate environment variables: `OPENAI_API_KEY` or `MISTRAL_API_KEY`.
```

lumigator/python/mzai/backend/backend/config_templates.py
```diff
@@ -66,6 +66,24 @@
 
 # Inference templates
 
+default_infer_template = """{{
+    "name": "{job_name}/{job_id}",
+    "dataset": {{ "path": "{dataset_path}" }},
+    "pipeline": {{
```
javiermtorres, line 72: Is "pipeline" the right name for this config? It's a very generic name; I assume HF uses it because it is composed of several serial stages.
However, I admit I cannot come up with a suitable replacement :-/

lumigator/python/mzai/jobs/inference/inference_config.py
```diff
@@ -30,10 +120,55 @@ class SamplingParameters(BaseModel):
     model_config = ConfigDict(extra="forbid")
 
 
+class PipelineConfig(BaseModel, arbitrary_types_allowed=True):
```
javiermtorres, line 123: Why do we allow arbitrary types?

lumigator/python/mzai/backend/backend/config_templates.py
dpoulopoulos: @javiermtorres that is something I'm not sure either. If someone else has a better name I'm all ears.

lumigator/python/mzai/jobs/inference/inference_config.py
dpoulopoulos: For the custom types that we have defined (i.e., `AssetPath`, `SupportedTask`, and `SerializableTorchDtype`). If I remember correctly it throws an error if I don't have this. I can test it again.

dpoulopoulos: Yes, this is what I'm getting if I remove it:

```
Job entrypoint command failed with exit code 1, last available logs (truncated to 20,000 chars):
  File "/tmp/ray/session_2024-12-10_02-57-03_358398_1/runtime_resources/pip/617697551215d8488f42ccab087d2364573ba842/virtualenv/lib/python3.11/site-packages/pydantic/_internal/_generate_schema.py", line 997, in match_type
    return self._unknown_type_schema(obj)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/ray/session_2024-12-10_02-57-03_358398_1/runtime_resources/pip/617697551215d8488f42ccab087d2364573ba842/virtualenv/lib/python3.11/site-packages/pydantic/_internal/_generate_schema.py", line 515, in _unknown_type_schema
    raise PydanticSchemaGenerationError(
pydantic.errors.PydanticSchemaGenerationError: Unable to generate pydantic-core schema for <class 'torch.dtype'>. Set `arbitrary_types_allowed=True` in the model_config to ignore this error or implement `__get_pydantic_core_schema__` on your type to fully support it.

If you got this error by calling handler(<some type>) within `__get_pydantic_core_schema__` then you likely need to call `handler.generate_schema(<some type>)` since we do not call `__get_pydantic_core_schema__` on `<some type>` otherwise to avoid infinite recursion.

For further information visit https://errors.pydantic.dev/2.10/u/schema-for-unknown-type
```

docs/source/user-guides/inference.md
dpoulopoulos: Fixed, thanks!

docs/source/get-started/quickstart.md
```diff
@@ -164,7 +164,7 @@ team_name = "lumigator_enthusiasts"
 
 responses = []
 for model in models:
-    job_args = JobCreate(
+    job_args = JobEvalCreate(
```
aittalam, line 167: TY for catching this and the change in the schemas package name!

lumigator/python/mzai/jobs/inference/inference_config.py
aittalam: Thanks for checking out! Yes, if I remember well we had the same issue in evaluator. What I appreciate here is that arbitrary types are allowed only within this config, instead of the basic class as in evaluator, which makes this version a bit more tight.

lumigator/python/mzai/jobs/inference/inference_config.py
```diff
@@ -43,13 +121,13 @@ class SamplingParameters(BaseModel):
...
-    model_path: str = Field(title="The Model HF Hub repo ID", exclude=True)
-    task: str
+    model_path: AssetPath = Field(title="The Model HF Hub repo ID", exclude=True)
+    task: SupportedTask
     revision: str = "main"  # Model version: branch, tag, or commit ID
     use_fast: bool = True  # Whether or not to use a Fast tokenizer if possible
     trust_remote_code: bool = False
-    torch_dtype: str = "auto"
-    accelerator: str = Field(title="Accelerator", default=Accelerator.AUTO, exclude=True)
+    torch_dtype: SerializableTorchDtype = "auto"
```
aittalam: Just a heads up that adding torch as a dependency makes the job much heavier in terms of fingerprint.
As far as I understood, this is required to validate the torch datatype used for e.g. quantization.
Do we need this already for summarization or are we planning to use it later on for larger text generation models?

I am not against it in principle, I just wanted to raise that we might not need this right now (and perhaps find a different solution in the meantime). WDYT?

dpoulopoulos: But `torch` is a dependency or `transformers` anyway, how can we move around this?

lumigator/python/mzai/backend/backend/config_templates.py
aittalam: This are all parameters that are specific to HF pipelines, right? How does `hf_pipeline` sound?
We might also want to be aware of the fact that if we introduce a new pipeline (e.g. text generation) we might have some pipeline-specific params we want to provide. Would we put them in here within some `task_specific_params` section as they do in HF?

lumigator/python/mzai/jobs/inference/inference_config.py
```diff
@@ -30,10 +31,27 @@ class SamplingParameters(BaseModel):
...
+
+
 class InferenceJobConfig(BaseModel):
     name: str
     dataset: DatasetConfig
     job: JobConfig
-    inference_server: InferenceServerConfig
-    params: SamplingParameters
+    inference_server: InferenceServerConfig | None = None
+    params: SamplingParameters | None = None
```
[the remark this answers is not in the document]
aittalam: works for me! I will add this into an issue so we can track this (= I don't forget ;-))

aittalam (approved):
Thank you Dimitris! There are a couple of unresolved convos here but I do not think they are blocking, so I am going to pre-approve this and let you close them.

lumigator/python/mzai/jobs/inference/inference_config.py
aittalam: That's right, my bad! Transformers needs to be backed up by torch anyway 👍

lumigator/python/mzai/backend/backend/config_templates.py
dpoulopoulos: @aittalam sure, we can integrate it with [this](https://github.com/mozilla-ai/lumigator/blob/99ce0bf59d7ee365acc365df2f0287b44172291e/lumigator/python/mzai/backend/backend/models.yaml). However, I would like to address than in a different PR. Is that okay?

I'll also rename `pipeline` to `hf_pipeline`.

Issue #487, opened 2024-12-10
Support minio as an S3-compatible alternative to localstack


aittalam:
Localstack requires to use the "pro" image to guarantee persistence. [Minio](https://min.io/) should provide the same functionality but defaulting with persistence. It would be useful if we provided minio as an alternative so we do not require people to register and use the pro version of localstack.

PR #491, opened 2024-12-11
Replace localstack with minio

javiermtorres:
# What's changing

The localstack service is replaced with minio for local deployments.

Closes https://github.com/mozilla-ai/lumigator/issues/487

# How to test it

Start lumigator (e.g. using `make local-up`) and check that minio starts instead of localstack.

All automated tests should work as usual.

N/A

# I already...

  - No DB update is needed.


javiermtorres:
One issue I found is that, if the .env.example is modified, then the integration tests should also be run, just in case the env vars are no longer valid. I can make a follow up PR to amend that.

Also, I didn't notice a clear indication that the .env would be generated from the .env.example and used within the docker-compose. If such indication is indeed present, could anyone please point me to it? @agpituk maybe? Thanks in advance.

aittalam:
> Also, I didn't notice a clear indication that the .env would be generated from the .env.example and used within the docker-compose. If such indication is indeed present, could anyone please point me to it? @agpituk maybe? Thanks in advance.

Not sure if it is in the documentation, but you can find it in the Makefile:

```
.env:
    @if [ ! -f .env ]; then cp .env.example .env; echo ".env created from .env.example"; fi
```

docker-compose.yaml
```diff
@@ -2,35 +2,50 @@ name: lumigator
...
 services:
 
-  localstack:
-    # if you want to enable local S3-style data persistence, use the following
-    # image and set LOCALSTACK_AUTH_TOKEN in your env
-    # image: localstack/localstack-pro:3.4.0
-    image: localstack/localstack:3.4.0
-    platform: linux/amd64
+  minio:
+    image: quay.io/minio/minio # TODO add version
```
aittalam, line 6: +1 on relying on a specific version!

docker-compose.yaml
```diff
@@ -2,35 +2,50 @@ name: lumigator
...
-    entrypoint: >
-       bash -c "awslocal s3 mb s3://lumigator-storage"
+    command:
+      - /bin/bash
+      - -c
+      - |
+          set -ex
+          apt update
+          apt install -y curl
+          curl --create-dirs -o /etc/lumigator/minio/mc https://dl.min.io/client/mc/release/linux-amd64/mc
```
aittalam, line 41: Using a generic container where we need to install the minio client requires us to install `curl` and run `apt update` first, the result is that it takes more than one minute to pull the whole system up. Another linked issue is that whenever there's an issue with `apt-get` we cannot start lumigator. This happened while testing:

image

As an alternative we could have something like:
```
 minio-admin:
    image: quay.io/minio/minio
    depends_on:
      minio:
        condition: service_healthy
    entrypoint:
      - /bin/bash
    command:
      - -c
      - |
          set -ex
          ...
```

The minio image is already cached anyway, the load time decreases to ~35 sec, and we do not have to reinstall stuff at every run. WDYT?

A follow up question: is there any way to persist the minio configuration so we can pick it up if available and run the following only if that is not already set?

lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
```diff
@@ -12,6 +12,7 @@
 
 @app.on_event("startup")
 def test_health_ok(local_client: TestClient):
+    # dummy change
```
aittalam, line 15: Great idea to have CI run! Just a heads up to remove it before merge  ;-)

aittalam (changes requested):
Thanks a lot for the PR @javiermtorres, I think this is great! I tested this and it runs for me (I had a few issues at first, mostly due to old env vars being present and the `.env`/`.env.example` thing going on, but afterwards things just ran fine). Accessing all persisted data via Web UI is very cool too :-)

Can I ask you to please go through our sources and look for `4566` and `localstack` too? I think there are few other spots we might need to update contextually to this change such as:

- `.env`
- documentation (README.md, installation.md, dev.md, etc)

re: `conftest.py`, I see we are still using localstack. Do you see the need to move tests to minio too? IMHO it is not something urgent, but testing on the same stack we are running makes a bit more sense (and reduces the amount of dependencies + lumigator's fingerprint)

javiermtorres:
@aittalam thanks for trying out the PR and checking that it works :) I'll address the concerns first thing on Monday.
Re: the dependence on localstack for tests.... It's being addressed in #453 , but unfortunately there's _one single test case_ that is inexplicably failing; I cannot even get meaningful logs from Ray. I know that it somehow has to do with inference, but I can't see how. I'll try rebasing to main, but it should be up to date. If something there rings a bell, _please please_ let me know :D

Once that other PR is in place, we'll remove any dependency on localstack in source and tests.

aittalam:
> @aittalam thanks for trying out the PR and checking that it works :) I'll address the concerns first thing on Monday. Re: the dependence on localstack for tests.... It's being addressed in #453 , but unfortunately there's _one single test case_ that is inexplicably failing; I cannot even get meaningful logs from Ray. I know that it somehow has to do with inference, but I can't see how. I'll try rebasing to main, but it should be up to date. If something there rings a bell, _please please_ let me know :D
> 
> Once that other PR is in place, we'll remove any dependency on localstack in source and tests.

That's great @javiermtorres , many thanks!
I'll check https://github.com/mozilla-ai/lumigator/pull/453 today and see if I can be of any help there.

.env.example
```diff
@@ -12,10 +12,10 @@ LUMI_API_CORS_ALLOWED_ORIGINS=${LUMI_API_CORS_ALLOWED_ORIGINS:-http://localhost,
 
 # AWS Variables for S3 Object Storage
 # Configure these for AWS access, or use defaults for local development with LocalStack.
-AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID:-test}
-AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY:-test}
+AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID:-lumigator}
+AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY:-lumigator}
 AWS_DEFAULT_REGION=${AWS_DEFAULT_REGION:-us-east-2}
-AWS_ENDPOINT_URL=${AWS_ENDPOINT_URL:-http://localhost:4566}
+AWS_ENDPOINT_URL=${AWS_ENDPOINT_URL:-http://localhost:9000}
```
veekaybee, line 18: let's leave a comment about why this is `9000`

javiermtorres:
> > Also, I didn't notice a clear indication that the .env would be generated from the .env.example and used within the docker-compose. If such indication is indeed present, could anyone please point me to it? @agpituk maybe? Thanks in advance.
> 
> Not sure if it is in the documentation, but you can find it in the Makefile:
> 
> ```
> .env:
>     @if [ ! -f .env ]; then cp .env.example .env; echo ".env created from .env.example"; fi
> ```

It won't hurt to add some comment in some README to warn that the `.env` is included in the Makefile. I'll check it out.

docker-compose.yaml
javiermtorres: Choosing RELEASE.2024-12-18T13-15-44Z as the latest non-new-feature release. Since semver is not used I cannot easily check sets of features or backward compatibility :-/ Let me know if anything breaks with this.

docker-compose.yaml
javiermtorres: > The minio image is already cached anyway, the load time decreases to ~35 sec, and we do not have to reinstall stuff at every run. WDYT?

If you tried it out and it worked, fine for me. I don't think the container size will be an issue. I didn't use it at first because one needs to install the minio client anyway (it's not bundled with minio itself afaict), but it definitely helps to use a cached image.

javiermtorres: > A follow up question: is there any way to persist the minio configuration so we can pick it up if available and run the following only if that is not already set?

Hm. Good question. If the commands are idempotent it doesn't matter (so buckets are not recreated anew if they already exist, etc etc). Was there something specific that failed in your setup?

lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
javiermtorres: Wops sorry :-S

docker-compose.yaml
javiermtorres: Resolving this issue. I followed your suggestion and reused the minio image, it does indeed contain the mc admin commands. Shame on me 👎

javiermtorres:
> > A follow up question: is there any way to persist the minio configuration so we can pick it up if available and run the following only if that is not already set?
> 
> Hm. Good question. If the commands are idempotent it doesn't matter (so buckets are not recreated anew if they already exist, etc etc). Was there something specific that failed in your setup?

Leaving the notes on minio persistence here for a future issue, since the original comment is resolved.

aittalam (approved):
LGTM, thank you!
I saw the branch got out-to-date wrt main, I pre-approve this so you'll be able to land this after merging from main.

javiermtorres:
Thanks a lot @aittalam :)

agpituk (approved):

PR #504, opened 2024-12-13
Fix job's results file name

### LINKED ISSUE #503: [BUG]: The generated results download URL always points to an `eval_results.json` regardless of the job type
### Description

When an inference (#339) or evaluation (#435) job completes successfully, it creates a results file and stores it in object storage. The file is named either `eval_results.json` or `inference_results.json`, depending on the job type.

However, when attempting to generate a download URL through `/api/v1/jobs/{job_id}/result/download`, the URL always points to a file named `eval_results.json`. For inference jobs, this file does not exist, resulting in a broken download process.

### Reproduction

Steps to Reproduce:
1. Run an inference job to completion.
2. Check the S3 URI that points to the results file. The file name is `inference_results.json`.
3. Try to download the file, by retrieving a download URL through the API. This step should break because the URL points to a non-existent  `eavl_results.json` file.

### Relevant log output

```shell
<Error>
  <Code>NoSuchKey</Code>
  <Message>The specified key does not exist.</Message>
  <RequestId>232b17c8-fefe-4826-a01e-6b5c39d41063</RequestId>
  <Key>
    jobs/results/bart-run/41413bf4-3c34-4c6a-9dc9-452043215b45/eval_results.json
  </Key>
</Error>
```

### Expected behavior

The generated URL should point to the correct file.

### System Info

- OS: OS-independent, but tested on macOS Sequoia `15.1.1`
- Version: commit `c83ec7de5cdf9d3e6552b560ccfcbc3358ce3318`

dpoulopoulos:
# What's changing

When an inference or evaluation job completes successfully, it creates a results file and stores it in object storage. The file is named either `eval_results.json` or `inference_results.json`, depending on the job type.

However, when attempting to generate a download URL through `/api/v1/jobs/{job_id}/result/download`, the URL always points to a file named eval_results.json. For inference jobs, this file does not exist, resulting in a broken download process.

Dynamically assign the file name based on the job type.

Also, change `eval_results` to `evaluation_results` for uniformity.

Closes #503

# How to test it

Steps to Reproduce:

1. Run an inference job to completion.
1. Check the S3 URI that points to the results file. The file name is `inference_results.json`.
1. Try to download the file, by retrieving a download URL through the API. It should point to a file called `inference_results.json`.

# I already...

- - [x] Added some tests for any new functionality
- [NA] Updated the documentation (both comments in code and [product documentation](https://mozilla-ai.github.io/lumigator) under `/docs`)

### DIFF
--- a/lumigator/python/mzai/backend/backend/services/jobs.py
+++ b/lumigator/python/mzai/backend/backend/services/jobs.py
@@ -220,11 +220,13 @@ class JobService:
             "env_vars": runtime_env_vars,
         }
 
+        metadata = {"job_type": job_type}
+
         loguru.logger.info("runtime env setup...")
         loguru.logger.info(f"{runtime_env}")
 
         entrypoint = RayJobEntrypoint(
-            config=ray_config, runtime_env=runtime_env, num_gpus=worker_gpus
+            config=ray_config, metadata=metadata, runtime_env=runtime_env, num_gpus=worker_gpus
         )
         loguru.logger.info("Submitting Ray job...")
         submit_ray_job(self.ray_client, entrypoint)
--- a/docs/source/get-started/quickstart.md
+++ b/docs/source/get-started/quickstart.md
@@ -240,7 +240,7 @@ user@host:~/lumigator$ curl -s http://localhost:8000/api/v1/jobs/$SUBMISSION_ID/
   -H 'accept: application/json' | jq
 {
   "id": "5195c9a5-938d-475e-b0fc-cf866492909d",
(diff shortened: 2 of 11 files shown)


lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -284,7 +288,17 @@ def get_job_result(self, job_id: UUID) -> JobResultResponse:
     def get_job_result_download(self, job_id: UUID) -> JobResultDownloadResponse:
         """Return job results file URL for downloading."""
         # Generate presigned download URL for the object
-        result_key = self._get_results_s3_key(job_id)
+        job_info = self.ray_client.get_job_info(job_id)
```
aittalam, line 291: A heads up that one might want to download a job result even after a system restart.
Data about experiments _should_ be persisted in the backend (DB + S3) but it would likely not persist in ray (especially in the local setup).

I think the main problem right now is that we are changing the output results filename depending on the job type. This is not really necessary as we are not planning a single job to return more than one output file.

Do you think it would make sense to just call the file `results.json` regardless of the job type? This would allow download to be agnostic of the job type (as it should be). If we want to introduce any more flexibility, I think storing S3 path together with job metadata would help, but I think it is out of scope for the current PR.

dpoulopoulos: That's even simpler. We can revisit it at some point, if we need a job to create more than one result files. I'll keep the commit which passes the job type through the metadata structure, and simplify the rest.

aittalam: Thanks for keeping the metadata commit, I think it will come handy sooner or later! 🙏

aittalam (approved):
LGTM, thanks Dimitris!

Issue #474, opened 2024-12-05
Refactor /api/v1/health/jobs/{job_id} endpoint


agpituk:
### Motivation

Currently under our health/jobs/{job_id} endpoint returns information that shouldn't be part of it.
`{
  "type": "string",
  "job_id": "string",
  "submission_id": "string",
  "driver_info": "string",
  "status": "string",
  "entrypoint": "string",
  "message": "string",
  "error_type": "string",
  "start_time": "2024-12-05T14:14:49.363Z",
  "end_time": "2024-12-05T14:14:49.363Z",
  "metadata": {},
  "runtime_env": {},
  "driver_agent_http_address": "string",
  "driver_node_id": "string",
  "driver_exit_code": 0
}`

The jobs/{job_id}. As a response, it returns:

`{
  "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "name": "string",
  "description": "string",
  "status": "created",
  "created_at": "2024-12-05T14:15:14.064Z",
  "updated_at": "2024-12-05T14:15:14.064Z"
}`

I believe we should merge both endpoints under `jobs/{job_id}` and health would only be related to api health

### Have you searched for similar issues before submitting this one?

- [X] Yes, I have searched for similar issues

PR #514, opened 2024-12-13
API: Refactor jobs endpoints from `health/` to `jobs/`

peteski22:
# What's changing

* New API schema type `Job` that composes (merges) `JobResponse` and `JobSubmissionResponse`.
* Job related endpoints have been removed from `/heath`.
* Updated `/jobs` endpoint to support logs at `/jobs/{id}/logs`.
* List all jobs returns results from the repository where Ray metadata also exists, it also updates the repository (job status) in the process.
* Get job ID returns composed/merged result from repository and Ray.
* UI updated to use new endpoints.

The rationale for merging the two types was:

* Reduce any need to make DB changes.
* There aren't really any conflicts in merging the data (only status 'clashes' but it's the same status anyway).
* We're at the stage where we're looking at how we track experiments which will require larger breaking changes to the API in the near future.

Fixes: #474 

* Schemas needs to be republished with an updated RC version
* The SDK needs to be updated to:
  * Remove health/jobs related code
  * Add job logs
  * Update schemas

Adding reviewers for components, e.g. UI, backend etc.

# I already...


lumigator/python/mzai/schemas/lumigator_schemas/jobs.py
```diff
@@ -105,3 +104,7 @@ class JobResultResponse(BaseModel, from_attributes=True):
 class JobResultDownloadResponse(BaseModel):
     id: UUID
     download_url: str
+
+
+class Job(JobResponse, JobSubmissionResponse):
```
aittalam, line 109: What is the rationale behind merging together all info? I am trying to understand whether it'd make sense to have a `ray` subsection here that might as well be optional if there's no job info on ray rather than a bunch of null fields.

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -58,8 +62,38 @@ def list_jobs(
     service: JobServiceDep,
     skip: int = 0,
     limit: int = 100,
-) -> ListingResponse[JobResponse]:
-    return service.list_jobs(skip, limit)
+) -> ListingResponse[Job]:
```
aittalam, line 35: Do we need to update `get_job(service: JobServiceDep, job_id: UUID) -> JobResponse` to `get_job(service: JobServiceDep, job_id: UUID) -> Job` too here?

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -58,8 +62,38 @@ def list_jobs(
...
+
+    results = list[Job]()
+
+    # Merge Ray jobs into the repositories jobs
+    for job in jobs.items:
+        found_job = next(
+            (x for x in filter(lambda x: x.submission_id == str(job.id), ray_jobs)), None
+        )
+        if found_job is None:
+            continue
```
aittalam, line 62: Have we decided to only show jobs which are currently logged on ray? Note that this way one might lose all info from previous runs if ray is restarted.

peteski22: The only results that are returned are when the ID matches something in the DB **and** also Ray., so you wouldn't see jobs that only exist in Ray.

This fixes a bug folks saw in some of the demo's, due to multiple users running Lumigator locally, but connecting to a 'hybrid' Ray (in the cloud).

The only downside is that if Ray has no information on the job that the repository believes should exist, it won't be included in the returned list of jobs. But (to me) that feels like a bit of an edge case for fault tolerance/integrity
(and something we could review and design what we'd like the intended behaviour to be in these situations).

lumigator/python/mzai/schemas/lumigator_schemas/jobs.py
peteski22: The rationale was mostly to reduce any changes required by consuming code as there weren't really any issues with conflicting fields (only status). Code that was calling `/health/jobs` related endpoints could change it's endpoint to `/jobs` (removing `/health`) and get the same data structure as before for the job (with a few additional ones like `name` and `description`).

This felt like a sensible middle ground until we have experiments in place and a more well defined set of models/boundaries etc for our API. So a step in the right direction, but not the last step. 😄

lumigator/python/mzai/backend/backend/api/routes/jobs.py
aittalam: Yeah this last use case is what I meant. In the local scenario we cannot expect ppl to have the system up the whole time and when they turn it off and on again they should expect to at least see the non-ray data available (we'll add persistence of older ray data too when we have experiment tracking in place).

This also works in a non-hybrid demo setting: suppose you need to restart ray for whatever reason, you'll immediately lose all the jobs you just ran.

This is the main reason why I raised this point and the fact that ray information could / should be optional

peteski22: Today the UI relies on both `jobs/` and `health/jobs` endpoints to function, so without Ray you wouldn't have anything useful such as the [list of experiments](https://github.com/mozilla-ai/lumigator/blob/main/lumigator/frontend/src/components/molecules/LExperimentTable.vue#L100), or updates to job status.

Even when you get experiment details, the UI might use the `jobs/` endpoint, but internally in the job service we somtimes make a call to Ray to get the up-to-date [job status](https://github.com/mozilla-ai/lumigator/blob/7c25de25ffe9e2d13d9bbcc957a32585f6e605d1/lumigator/python/mzai/backend/backend/services/jobs.py#L244) if it wasn't already 'finished' (succeed, failed). If the job no longer exists in Ray, then we'd [see errors](https://docs.ray.io/en/latest/cluster/running-applications/job-submission/doc/ray.job_submission.JobSubmissionClient.get_job_status.html).

It's a good point to raise, I do appreciate you calling it out. But since it doesn't work well today because we never planned for it, I think it makes sense to look at dealing with it separately from this PR. That way we can scope it properly and make sure we understand any other edge cases that might pop up, and loop in Product for feedback on how they'd expect things to work for the user.

For example, what kind of degraded performance should users expect based on which of the required components of Lumigator are down/missing/cleared out? Should we always show jobs from our respository (in whatever potentially stale state they're in), and what parts of the UI might be missing data, or unable to function. What if Ray's up and full of data but S3 storage went down? It feels like the current expectation for usage is that all the components need to be up and running, and not turned off/on. Totally agree that as the product matures this isn't going to work well, but soon we will have moved away from the jobs endpoints (hopefully) anyway.

I'm happy to create a GitHub issue for us to track this, as you mentioned it might also change as we look at introducing experiments.

aittalam: Ok, let us track this with an issue! My concern is that we are changing the semantics of the API (ie. `jobs/` currently returns the list of all jobs stored in the DB, you propose to only return those which are present in ray) because the UI uses both the DB to get jobs and ray to get info about them.

As additional info for the issue (you can create it fromt this convo), note that jobs' status can be retrieved lazily via [get_job](https://github.com/mozilla-ai/lumigator/blob/main/lumigator/python/mzai/backend/backend/services/jobs.py#L237) so if a job is completed you can get info about it without necessarily hitting ray.

peteski22: Yeah, I get what you're saying, it does change things a bit, and in a mature product/API I'd never suggest the same thing, but we've made a series of pre-MVP breaking changes to the API recently so I didn't feel as concerned.

I think I was saying somewhere in the comments or summary of the PR this is a step along the way/middle ground/temporary solution.

The two drivers for the change were:

1. Try to make an incremental improvement to the endpoint design to group job related 'stuff' together.
2.  Resolve the issue that was seen in demos where experiments (jobs) were shown in the UI that the user had no way to interact with because they didn't exist in local storage but did exist in Ray. 

RE: 1. We have similar but different information from each job endpoint, that now needs to live together. The UI needs both to function at all, and I didn't want to break the data structure of the objects themselves, and also not bleed out 'ray' where possible. 

I totally understand that it feels 'off'/sub-optimal/wrong to not list what we have in the DB, regardless of whether it's accurate or not, but we don't have enough info to make everything work well in the UI without Ray, so it seemed to make most sense to only show jobs for which we had 'full integrity', and under the expectation that Lumigator components are 'always on' as they need to be currently. 

> As additional info for the issue (you can create it from this convo), note that jobs' status can be retrieved lazily via [get_job](https://github.com/mozilla-ai/lumigator/blob/main/lumigator/python/mzai/backend/backend/services/jobs.py#L237) so if a job is completed you can get info about it without necessarily hitting ray.

Yep 👍🏼 , I agree, it's what I meant with this earlier comment:

> internally in the job service we somtimes make a call to Ray to get the up-to-date [job status](https://github.com/mozilla-ai/lumigator/blob/7c25de25ffe9e2d13d9bbcc957a32585f6e605d1/lumigator/python/mzai/backend/backend/services/jobs.py#L244) if it wasn't already 'finished' (succeed, failed)

I've [created the issue - 531](https://github.com/mozilla-ai/lumigator/issues/531), please feel free to add to it 😄

javiermtorres: > so you wouldn't see jobs that only exist in Ray.

And at the same time,

> currently returns the list of all jobs stored in the DB, you propose to only return those which are present in ray

I guess that if we show all jobs in the DB, whether or not they're in Ray, then we're fulfilling both constraints????
Of course, we'd need to signal somehow that the ray information is missing (that's why I usually prefer aggregation, it's easier to just omit a `ray_info` key), but in this case the ray provided keys are already expected by the UI, so that's fine for me.

IMHO we just need to go through this again once the jobs/experiment APIs are stable again. I'd accept any interim solution right now.

aittalam: Many thanks @peteski22 , I think the [issue](https://github.com/mozilla-ai/lumigator/issues/531) is very complete!

aittalam (approved):
Thank you @peteski22! I think the PR is ok after our conversation, thanks again for opening [this issue](https://github.com/mozilla-ai/lumigator/issues/531) as a follow-up.

As I am not very confident with the UI section, I'd suggest to wait for one more approval from either @george-mzai or @agpituk so you can make sure they are aligned on this too.

george-mzai (approved):

javiermtorres (approved):

agpituk (approved):

Issue #485, opened 2024-12-10
Workflow using independent inference and evaluation, without annotation service


ividal:

javiermtorres:
Closed by https://github.com/mozilla-ai/lumigator/pull/540

Issue #406, opened 2024-11-20
Workflow that includes annotation-generation without human input inbetween


ividal:
**Is your feature request related to a problem? Please describe.**
We temporarily stopped supporting experiments that require the generation of annotations with an LLM. Once annotation generation (`/groundtruth`) is back online (#337), it would make sense to build workflows (experiments) with datasets that lack human-provided groundtruth again.

**Describe the solution you'd like**
The ability for users to still use a dataset without ground truth to evaluate multiple models without stopping to ask for approval. The steps should be:

1. loading a dataset
1. generating annotations with a reference model (in this we should be opinionated on which one)
1. launch predictions with a user-defined model
1. launch and evaluation on the outcome of 2 and 3.

**Describe alternatives you've considered**
Staying within a more transparent flow as in #481 , where the user decides whether to complement a dataset with annotations and then the usual flow continues, with a now-annotated dataset.

Going into introducing full-blown DAGs for this. Up for discussion, but we could start with the same 2 workflows that already existed (with/without GT).

**Additional context**
This could have repercussions on how we track results.

ividal:
We've discussed with @hasangzl that the user flow involves:
* user uploading dataset - optional trigger of a stand-alone annotation job to add pseudo-gt to dataset
* launching an experiment the usual way, with the dataset that now has the annotations

This means we don't actually need a composite task that calls annotation as one of its steps yet.

When we include llm-as-judge metrics, again the automated workflow would still be the one in #485, so annotations is always a stand-alone task that the user consciously triggers.

PR #519, opened 2024-12-17
337 annotation/ground truth generation uses new inference job

### LINKED ISSUE #337: Annotation/ground-truth generation uses new inference job
Prior to #285 we had a service to generate annotations from a trusted model (the /groundtruth endpoint) when ground-truth was missing. 

Since we now have a separate inference job (jobs/inference, from #309), the annotation generation job can call it directly with specific parameters to generate labels with a predetermined model.

This issue re-introduces the endpoint with different wiring based on #309 (LUMI-115).

Note: The response needs to include what model was used for this job, since a separate check within an experiment should ensure that different models are used for annotations and predictions.

AC:

* SDK and UI are able to request annotations for a dataset via API.
* The metadata for the resulting dataset has the input from the original dataset, plus a field called "AI-generated" or similar set to True.
* The original dataset is not destroyed or modified.
* Response includes the model that was used for this inference job.

ividal:
# Context
Lumigator used to offer a service, /groundtruth, to annotate a dataset when it did not have human-provided ground-truth to compare against in evaluation. This PR does not address whether this is a good practice; inference marks its results as AI-generated (with the field "model: ai_model_that_generated_them" as an output attribute). The user can use this information to trust accordingly.

# What's changing

* Adds annotation using the new, independent inference job with a model selected by us (bart).
* Exposes the service as /jobs/annotate
* Introduces JobAnnotationCreate, which does not allow setting model or temperature. It hands over to a JobInferenceCreate which sets the model and temperature we decided. 

Closes #337 

# How to test it

Steps to test the changes:

1. Start lumigator (make local-up or start-lumigator build)
2. Navigate to localhost:8000/docs
3. Update a dataset, notice the dataset_id
4. Go to /jobs/annotate, set the dataset_id you obtained, a low number of samples, a random name and description, launch
5. Go to the ray dashboard (http://localhost:8265)
6. The new job should succeed and the results should show annotations.

I've considered:
* /annotate, to keep in line with the old /groundtruth
* Waiting until we have a ~~workflow~~ composite task from #485 and introduce this as a silent and optional step there (but that would be, rather, #406)
* Creating a composite task of 1 and expose through experiments

The two latter feel overkill, since we are considering introducing actual orchestration. Using /jobs/annotate acknowledges (hopefully signals to the user) that this is just another variant of the existing jobs).

This job will be called from the UI separately, after the user gets asked whether they are OK with generating some ground-truth as second best option for their dataset.

# I already...

### DIFF
--- a/docs/source/get-started/suggested-models.md
+++ b/docs/source/get-started/suggested-models.md
@@ -185,3 +185,35 @@ weights for a given open LLM, as well as everything needed to actually run that 
 computer. There's nothing to install or configure.
 
 There are no summarization-specific parameters for this model.
+
+
+# Reference models
+
+Before you jump into evaluating datasets, you should consider the following importance of quality ground-truth and annotations.
+
+Ground-truth would be the actual, expected output or correct answer in for a given task (such as summarization), serving as a reference to compare the model's predictions. Typically, a human with enough expertise in the task will annotate or label a dataset with those references for each sample (for example, an acceptable summary of the input text).
+
+To evaluate a model as reliably as possible, we encourage using human-provided ground-truth to compare against. Failing that, Lumigator enables the user to do automatic annotation with a [well tested model](https://blog.mozilla.ai/on-model-selection-for-text-summarization/) to get you started.
+
+You can do this through the API, using one of Lumigator jobs: `/jobs/annotate`.
+
+::::{tab-set}
+
+:::{tab-item} cURL
+:sync: tab1
+
+```console
+user@host:~/lumigator$
+curl -X 'POST' \
+  'http://localhost:8000/api/v1/jobs/annotate/' \
+  -H 'accept: application/json' \
+  -H 'Content-Type: application/json' \
+  -d '{
+  "name": "test_annotate",
+  "description": "annotate",
+  "dataset": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
+  "max_samples": -1,
+  "task": "summarization"
+}'
+```
+Under the hood, this will launch an inference job with the reference model for summarization (BART, described above).
--- a/lumigator/python/mzai/backend/backend/api/routes/jobs.py
+++ b/lumigator/python/mzai/backend/backend/api/routes/jobs.py
@@ -50,7 +50,7 @@ def create_annotation_job(
     request: Request,
     response: Response,
 ) -> JobResponse:
-    """This uses a hardcoded model, that is, Lumigator's opinion on what
+    """This uses a hardcoded model, that is, Lumigator's opinion on the what
     reference model should be used to generate annotations.
     See more: https://blog.mozilla.ai/lets-build-an-app-for-evaluating-llms/
     """
--- a/lumigator/python/mzai/backend/backend/api/routes/jobs.py
+++ b/lumigator/python/mzai/backend/backend/api/routes/jobs.py
@@ -50,7 +50,7 @@ def create_annotation_job(
     request: Request,
     response: Response,
 ) -> JobResponse:
-    """This uses a hardcoded model, that is, Lumigator's opinion on the what
+    """This uses a hardcoded model, that is, Lumigator's opinion on what
     reference model should be used to generate annotations.
     See more: https://blog.mozilla.ai/lets-build-an-app-for-evaluating-llms/
     """
--- a/lumigator/python/mzai/backend/backend/api/routes/jobs.py
+++ b/lumigator/python/mzai/backend/backend/api/routes/jobs.py
@@ -9,6 +9,7 @@ from fastapi import APIRouter, HTTPException, status
 from lumigator_schemas.extras import ListingResponse
 from lumigator_schemas.jobs import (
     Job,
+    JobAnnotateCreate,
     JobEvalCreate,
     JobInferenceCreate,
     JobLogsResponse,
@@ -35,6 +36,30 @@ def create_inference_job(
     response: Response,
 ) -> JobResponse:
     job_response = service.create_job(job_create_request)
+
+    url = request.url_for(get_job.__name__, job_id=job_response.id)
+    response.headers[HttpHeaders.LOCATION] = f"{url}"
+
+    return job_response
+
+
+@router.post("/annotate/", status_code=status.HTTP_201_CREATED)
+def create_annotation_job(
+    service: JobServiceDep,
+    job_create_request: JobAnnotateCreate,
+    request: Request,
+    response: Response,
+) -> JobResponse:
+    """This uses a hardcoded model, that is, Lumigator's opinion on what
+    reference model should be used to generate annotations.
+    See more: https://blog.mozilla.ai/lets-build-an-app-for-evaluating-llms/
+    """
+    inference_job_create_request = JobInferenceCreate(
+        **job_create_request.dict(),
+        model="hf://facebook/bart-large-cnn",
+        output_field="ground_truth",
+    )
+    job_response = service.create_job(inference_job_create_request)
 
     url = request.url_for(get_job.__name__, job_id=job_response.id)
     response.headers[HttpHeaders.LOCATION] = f"{url}"
--- a/lumigator/python/mzai/schemas/lumigator_schemas/jobs.py
+++ b/lumigator/python/mzai/schemas/lumigator_schemas/jobs.py
@@ -88,6 +88,14 @@ class JobInferenceCreate(BaseModel):
     config_template: str | None = None
 
 
+class JobAnnotateCreate(BaseModel):
+    name: str
+    description: str = ""
+    dataset: UUID
+    max_samples: int = -1  # set to all samples by default
+    task: str | None = "summarization"
+
+
 class JobResponse(BaseModel, from_attributes=True):
     id: UUID
     name: str
@@ -122,4 +130,5 @@ class Job(JobResponse, JobSubmissionResponse):
     consumers. Tt was not conceived as a type that will be around for long, as
     the API needs to be refactored to better support experiments.
     """
+
     pass
--- a/Makefile
+++ b/Makefile
@@ -123,7 +123,8 @@ test-backend-unit:
 
 test-backend-integration-target:
 	cd lumigator/python/mzai/backend/backend/tests;	\
-	SQLALCHEMY_DATABASE_URL=sqlite:///local.db uv run pytest -o python_files="backend/tests/integration/*/test_*.py"
+	RAY_WORKER_GPUS="0.0" RAY_WORKER_GPUS_FRACTION="0.0" \
+	SQLALCHEMY_DATABASE_URL=sqlite:///local.db uv run pytest -s -o python_files="backend/tests/integration/*/test_*.py"
 
 test-backend-integration:
 ifeq ($(CONTAINERS_RUNNING),)
--- a/lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
+++ b/lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
@@ -100,3 +100,27 @@ def test_job_non_existing(local_client: TestClient):
     response = local_client.get(f"/jobs/{non_existing_id}")
     assert response.status_code == 404
     assert response.json()["detail"] == f"Job {non_existing_id} not found."
+
+
+def test_annotation_job(
+    local_client: TestClient,
+):
+    payload = {
+        "name": "test_annotate",
+        "description": "Test run for Huggingface model",
+        "dataset": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
+        "max_samples": 2,
+        "task": "summarization",
+    }
+
+    post_response = local_client.post(
+        "/jobs/annotate/",
+        json=payload,
+    )
+
+    assert post_response.status_code == 201
+
+    job_id = post_response.json()["id"]
+    response = local_client.get(f"/jobs/{job_id}")
+
+    assert response.status_code == 200


docs/source/get-started/suggested-models.md
```diff
@@ -185,3 +185,12 @@ weights for a given open LLM, as well as everything needed to actually run that
 computer. There's nothing to install or configure.
 
 There are no summarization-specific parameters for this model.
+
+
+# Reference models
+
+We encourage using human-annotated dataset in your evaluations. Failing that, Lumigator enables the user to do automatic annotation with a [well tested model](https://blog.mozilla.ai/on-model-selection-for-text-summarization/) to get you started.
+
+You can do this through the API, using one of Lumigator jobs: `/jobs/annotate`.
```
veekaybee, line 194: would recommend leaving the full cURL command here for reference

docs/source/get-started/suggested-models.md
```diff
@@ -185,3 +185,12 @@ weights for a given open LLM, as well as everything needed to actually run that
 computer. There's nothing to install or configure.
 
 There are no summarization-specific parameters for this model.
+
+
+# Reference models
+
+We encourage using human-annotated dataset in your evaluations. Failing that, Lumigator enables the user to do automatic annotation with a [well tested model](https://blog.mozilla.ai/on-model-selection-for-text-summarization/) to get you started.
```
veekaybee, line 192:
```suggestion
We encourage using human-annotated dataset in your evaluations. However,  Lumigator also enables the user to do automatic annotation with a selection of [well-tested models](https://blog.mozilla.ai/on-model-selection-for-text-summarization/) to get you started.
```

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -33,12 +34,34 @@ def create_inference_job(
...
+@router.post("/annotate/", status_code=status.HTTP_201_CREATED)
+def create_annotation_job(
     service: JobServiceDep,
-    job_create_request: JobEvalCreate,
+    job_create_request: JobAnnotateCreate,
     request: Request,
-    response: Response
+    response: Response,
+) -> JobResponse:
+    # Lumigator's opinion on the best summarization model
```
veekaybee, line 44: Let's use docstrings here for multiline comments https://peps.python.org/pep-0257/

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -33,12 +34,34 @@ def create_inference_job(
...
+    # and the one that should generate annotations.
+    # In the future, we could expose this via a config file
```
veekaybee, line 46: I'd leave out what we plan to do in the future as this could be confusing in the code, relegating planning to issues or RFCS, or instead might leave this as a TODO

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -33,12 +34,34 @@ def create_inference_job(
...
+    # so that, for a supported task, there is a default
+    # "reference" model. For now, we keep the current functionality:
+    # Lumigator decides who annotates.
+
+    inference_job_create_request = JobInferenceCreate(
```
veekaybee, line 51: this should be `JobAnnotateCreate` if I'm understanding the intent correctly, otherwise if there's not enough of a distinction between Annotate and create to create its own class, we should forego JobIAnnotateCreate and stick with JobInferenceCreate, wdyt?

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -33,12 +34,34 @@ def create_inference_job(
...
+        **job_create_request.dict(), model="hf://facebook/bart-large-cnn"
+    )
+    job_response = service.create_job(inference_job_create_request)
+
+    url = request.url_for(get_job.__name__, job_id=job_response.id)
+    response.headers[HttpHeaders.LOCATION] = f"{url}"
+
+    return job_response
```
veekaybee, line 59: will the current generic job response include all the additional fields we need for inference as well, if there are any?

lumigator/python/mzai/backend/backend/tests/unit/api/routes/test_jobs.py
```diff
@@ -75,3 +75,29 @@ def test_get_job_results(
...
+    )
+
+    assert post_response.status_code == 201
+
+    job_id = post_response.json()["id"]
+    response = app_client.get(f"/jobs/{job_id}")
+    assert response.status_code == status.HTTP_200_OK
+
+    data = response.json()
+    assert data["status"].lower() == "pending"
```
veekaybee, line 103: if we have different assert points here, I'd break them up into separate unit tests so it's easier to see where a test fails, and inject the payload as a shared dependency from conftest as we do in other tests

lumigator/python/mzai/schemas/lumigator_schemas/jobs.py
```diff
@@ -88,6 +88,14 @@ class JobInferenceCreate(BaseModel):
     config_template: str | None = None
 
 
+class JobAnnotateCreate(BaseModel):
+    name: str
+    description: str = ""
+    dataset: UUID
+    max_samples: int = -1  # set to all samples by default
+    task: str | None = "summarization"
```
veekaybee, line 96: do we pass the JobAnnotateCreate into the generic job response?

ividal: We pass all it's contents and add the task-dependant model on top, to complete the JobInferenceCreate config. 
Currently we just have the one task, so there isn't much choice, but soon enough we'll want to distinguish between annotating for a translation job and a summarization job.

lumigator/python/mzai/backend/backend/tests/unit/api/routes/test_jobs.py
ividal: This little test helped uncover a few problems partially addressed in #453 and #543 .

So the pending check did not make sense (removed) and instead the following is necessary:
* specify no need for GPUs (that's why it stayed as pending, which in hindsight it shouldn't have);
* moved the test to integration tests, since we do want it to try to launch a job.

lumigator/python/mzai/backend/backend/api/routes/jobs.py
ividal: Just in case: the intent is to take the fields from the annotate job (= a subset of an inference job), complete with the "opinionated" fields, so to speak, and launch that opinionated inference job.

At first I considered sticking to JobInferenceCreate. The reason I didn't in the end was it meant exposing a field that we would later overwrite, which is kind of lying to the user (the user asks for mistral in their config file and then gets annotations by bart... that's not nice) and a silent failure risk.

So what we do here is create that opinionated InferenceCreate config with the AnnotateCreate fields plus the ones Lumigator mandates.

Does this make more sense?

lumigator/python/mzai/schemas/lumigator_schemas/jobs.py
veekaybee: If JobAnnotate is a special case of a an Inference Job, then we should have `JobAnnotateCreate` subclass `JobInferenceCreate` and pass additional parameters: 

```python
class JobAnnotateCreate(JobInferenceCreate):
    task: str | None = "summarization"
````

as far as I understand, this is only the extra parameter we pass for an AnnotationJob; we likely will also want the model, like so:

```
```python
class JobAnnotateCreate(JobInferenceCreate):
    task: str | None = "summarization"
    annotation_model: str = "hf://facebook/bart-large-cnn"
````

and pass that into the inference job via logic, wdyt? In thiat case we could call JobAnnotateCreate above https://github.com/mozilla-ai/lumigator/pull/519/files#diff-ae5c766d8034a73f4543261d29d9048bae65c707a6a0091ce61892963f8c64a1R64

File: lumigator/python/mzai/backend/backend/api/routes/jobs.py (hunk 6 above)
Comment (veekaybee) [reply lumigator/python/mzai/backend/backend/api/routes/jobs.py:72]: see comment here: https://github.com/mozilla-ai/lumigator/pull/519/files#r1901020255

File: lumigator/python/mzai/backend/backend/api/routes/jobs.py (hunk 5 above)
Comment (veekaybee) [reply lumigator/python/mzai/backend/backend/api/routes/jobs.py:64]: see comment here, we could change this to annotate: https://github.com/mozilla-ai/lumigator/pull/519/files#r1901020255

File: lumigator/python/mzai/schemas/lumigator_schemas/jobs.py (hunk 8 above)
Comment (ividal) [reply lumigator/python/mzai/schemas/lumigator_schemas/jobs.py:95]: My first instinct was also to inherit from JobInferenceCreate, but [that would expose model to the user](https://github.com/mozilla-ai/lumigator/pull/519/files#r1895931139), which is something I very much want to avoid for now. 

Notice that with the second option you list, JobAnnotateCreate will have 2 model fields (`annotation_model` and `model`, which it inherits from the parent class). So, if anything, we could go for the first option. Even better, if we want to give more agency to the user, we could:
1. stick to `JobInferenceCreate` (forget about `JobAnnotateCreate`),
2. ensure the GET /models by task returns which model is the preferred annotator, 
3. have the FE pre-fill the model with the preferred annotator and allow the user to modify it. 

This feels like going into the direction of MLE support (which I prefer, but wasn't the prio persona).

Do you have a different idea to enforce an annotator but let the user know what it will be?

Comment (veekaybee) [reply lumigator/python/mzai/schemas/lumigator_schemas/jobs.py:95]: I think for now as-is works, but it does bring up some thoughts for discussion for the future of where we'd want to make it easier, leaving them here for now in case we want to discuss later: 

1. If the user is someone **accessing the app via UI**, it doesn't matter much to them what the implementation is under the hood, and we can have a pre-filled model and allow the user to then modify. Their main access would be through the experiment creation process, starting here: 
Screenshot 2025-01-03 at 8 14 21 AM

2. If we envision users always **creating jobs via the API** directly via the SDK or API by hit `/jobs/annotate` with a set of parameters, we want to keep those minimal and inject all the logic afterwards at the service call site, like in the original code

3. If we envision the user to be the **platform developer** right now, for the sake of legibility, we need some place to keep track of all the parameters that are passed with any given job instantiation <-- this is what I was going with for the suggestion but it does create friction for user 2 and user 1, as you mentioned.

Comment (ividal) [reply lumigator/python/mzai/schemas/lumigator_schemas/jobs.py:95]: > we need some place to keep track of all the parameters that are passed with any given job instantiation

💯 We should be tracing + giving the user more information here. Sure, mlflow will, but we can be helpful with little effort by returning the trace in the response, for example. --> #557 created.

File: lumigator/python/mzai/backend/backend/api/routes/jobs.py (hunk 5 above)
Comment (ividal) [reply lumigator/python/mzai/backend/backend/api/routes/jobs.py:64]: Discussed [here](https://github.com/mozilla-ai/lumigator/pull/519/files#r1901742413). Resolving this one to make review easier.

File: lumigator/python/mzai/backend/backend/api/routes/jobs.py (hunk 6 above)
Comment (ividal) [reply lumigator/python/mzai/backend/backend/api/routes/jobs.py:72]: Same as above: discussed [here](https://github.com/mozilla-ai/lumigator/pull/519/files#r1901742413), resolving this one.

File: lumigator/python/mzai/backend/backend/api/routes/jobs.py [hunk 9]
```diff
@@ -48,6 +49,29 @@ def create_inference_job(
...
 
+@router.post("/annotate/", status_code=status.HTTP_201_CREATED)
+def create_annotation_job(
+    service: JobServiceDep,
+    job_create_request: JobAnnotateCreate,
+    request: Request,
+    response: Response,
+    background_tasks: BackgroundTasks,
+) -> JobResponse:
+    """This uses a hardcoded model, that is, Lumigator's opinion on the what
```
aittalam, line 60:
```suggestion
    """This uses a hardcoded model, that is, Lumigator's opinion on what
```

aittalam: I am concerned about hardcoding a single model for annotations: even if a model is good, it might not always be the best one depending on the use case.
Also, the choice of the model biases evaluation results - if we hardcode bart what can we expect in terms of LLM eval results?

I think this all boils down to how hard we want to make it for users to choose another model for annotation. What if we provided an opinionated default and good documentation for ppl to try and use different models if they choose to do so?

lumigator/python/mzai/schemas/lumigator_schemas/jobs.py
aittalam: +1 on a lot of things here :-)

- I like considering annotation as a particular case of inference, with a few fields which are filled by default (e.g. the default model we suggest and the output field name - or will it still be `predictions`?)
- I appreciate the different logical levels: if we want to be opinionated wrt the models we suggest for annotation that's fine, but for which audience is this? If for UI users, then let us not enforce stuff in the API and make things more difficult for MLEs / developers. My 2 cents here are: let us make things super easy for those who use the UI, but allow our APIs to be as general as possible so developers (or even SDK users) can easily get the best out of them

lumigator/python/mzai/backend/backend/api/routes/jobs.py
ividal: I completely agree with that concern (for anyone else reading, see discussion [here](https://github.com/mozilla-ai/lumigator/pull/519/files#r1901742413)). This PR is about re-introducing the annotation service, changing as little as possible conceptually, so it's a base for improvements.

aittalam (approved):
Thanks for the PR Irina!
I just read your reply to my main concern and I think we are aligned. Just two more things:
- a [minor fix to the docs](https://github.com/mozilla-ai/lumigator/pull/519/files#r1904361288): 
- merging from main

I pre-approve the PR as these won't require any further review on my side, you can merge after the two points above are addressed. Thanks again!

veekaybee (approved):
Looks good after tests pass

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -42,6 +43,29 @@ def create_inference_job(
     return job_response
 
 
+@router.post("/annotate/", status_code=status.HTTP_201_CREATED)
+def create_annotation_job(
+    service: JobServiceDep,
+    job_create_request: JobAnnotateCreate,
+    request: Request,
+    response: Response,
+    background_tasks: BackgroundTasks,
```
ividal, line 52: Not needed here anymore (see [this](https://github.com/mozilla-ai/lumigator/commit/60fa6d7de9a410f383188481af9c5091b26d1784#diff-ae5c766d8034a73f4543261d29d9048bae65c707a6a0091ce61892963f8c64a1R37)).

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -42,6 +43,29 @@ def create_inference_job(
...
+) -> JobResponse:
+    """This uses a hardcoded model, that is, Lumigator's opinion on what
+    reference model should be used to generate annotations.
+    See more: https://blog.mozilla.ai/lets-build-an-app-for-evaluating-llms/
+    """
+    inference_job_create_request = JobInferenceCreate(
+        **job_create_request.dict(), model="hf://facebook/bart-large-cnn"
+    )
+    job_response = service.create_job(inference_job_create_request, background_tasks)
```
ividal, line 61: [Same](https://github.com/mozilla-ai/lumigator/pull/519/files#r1907845447) here.

PR #521, opened 2024-12-17
Davide/485 inference eval

aittalam:
# What's changing

Draft PR - will add details before marking it ready for review

* Updated section...
* Added new...
* Fixed a typo in...
* ...

Refs #...
Closes #...

# How to test it

Steps to test the changes:

1.
2.
3.

# I already...


lumigator/python/mzai/backend/backend/services/experiments.py
```diff
@@ -0,0 +1,48 @@
+import json
+import time
+
+import loguru
+import requests
+from fastapi import BackgroundTasks
```
javiermtorres, line 6: IMHO we should get used to handling async and the event loop directly. For example, it seems the source for BackgroundTasks handles a threadpool for sync funs and asyncio for async funs. In our case, we may find clogging a threadpool just by running sleeps, which are other supported in a single thread when using asyncs. Inside the backend, we're not yet running cpu intensive tasks that require using separate threads.

lumigator/python/mzai/backend/backend/services/experiments.py
```diff
@@ -0,0 +1,48 @@
...
+        self._job_service = job_service
+
+    def experiment_specific_background_task(self, job_id: str):
+        status = "PENDING"
+        loguru.logger.info(f"Job id: {job_id}")
+        while status != "SUCCEEDED" and status != "FAILED":
+            res = requests.get(f"http://localhost:8000/api/v1/health/jobs/{job_id}")
+            status = json.loads(res.text)["status"]
+            loguru.logger.info(f"Job id: {job_id}, status: {status}")
+            time.sleep(5)
```
javiermtorres, line 26: I'd suggest making this an async fun and using asyncio.sleep, otherwise the rest should be the same 👍

javiermtorres: (This is independent of using or not BackgroundTasks, since it seems to use the asyncio event loop anyway under the hood for async funs)

aittalam: Thanks for the suggestion!

Requests does not work well with asyncio, which is why I opted for fastapi managing this with threads. In my next commit I will introduce proper code to get a job's status which relies on the ray client. This is not just way cleaner, but it also seems (just tested!) to be working properly within async so I'll be able to switch to async functions.

One of the advantages of using FastAPI built-in background task management is that it automatically deals with both sync and async functions, so I could first test something quickly with requests and then move to ray client + async.

lumigator/python/mzai/backend/backend/services/experiments.py
aittalam: Thank you! I think my answer below addresses this comment too re: both BackgroundTasks vs manual implementation and threads vs async, but please lmk if there's still something not clear!

I appreciate your comment and will make sure this is addressed in the inference job PR where ray client is already used. 🙏

lumigator/python/mzai/backend/backend/services/experiments.py
javiermtorres: Ouch, I see :-/ Let's tackle this later on, then. Thanks a lot for the info!!! :)

PR #522, opened 2024-12-17
Store the dataset generated by an inference job

### LINKED ISSUE #405: [FEATURE]: Save inference output as a dataset which can be downloaded
# Motivation

When the user runs an inference job (#339), an extra "field" is calculated containing a model’s predictions. The semantics of this field might be different depending on the purpose inference was used for (e.g., prediction or ground truth generation), but in all cases it is a prediction of a model. Only the user should keep track of what the data was used for.

Inference should create a new dataset, with one extra column containing the model's predictions. This dataset should be stored on object storage and the database should be updated.

# Alternatives

* An idea is to allow the prediction field to be named differently based on whether the job's goal was to generate "ground-truth" or another form of prediction. However, this could lead to errors. To maintain simplicity, we could standardize the output of an inference job to always represent "predictions", leaving it to the user to handle any further differentiation as needed.
* Having MLflow (#438) handle this entirely. Ideally this work would be independent, though: whether there is a tracking service or not, it is reasonable that the inference job stores the output somewhere and returns the URI. Whether this URI is tracked with one service or another is independent.

dpoulopoulos:
# What's changing

When the user runs an inference job, an extra "field" is calculated containing a model’s predictions.

This PR creates a new dataset, with one extra column containing the model's predictions. This dataset is stored on object storage and the (local) database.

Closes #405

# How to test it

Steps to test the changes:

1. Apply the database migration: `alembic upgrade head`
1. Add a new dataset and run an inference job.
1. Check the logs of the inference job; on successful completion a new dataset should be generated.
1. Get the dataset using the Lumigator API (either get all datasets or use the ID provided in the logs).

# I already...

- - [ ] Added some tests for any new functionality
- - [ ] Updated the documentation (both comments in code and [product documentation](https://mozilla-ai.github.io/lumigator) under `/docs`)
- - [x] Checked if a (backend) DB migration step was required and included it if required


aittalam (approved):
Thank you @dpoulopoulos! I ran the PR and the tests myself and everything works as expected.
I agree with your choice of mocking the background task. One further reason to avoid it is that as the experiment takes some time to run, we might not want to wait for it to end.  One thing we could do is mock the results data and just test the callback manually, but let's leave it for a follow-up PR.

Issue #435, opened 2024-11-25
Evaluation as a stand-alone job over data, not the trigger for multiple jobs


ividal:
**Is your feature request related to a problem? Please describe.**
So far, Lumigator's evaluation was actually a combination of these tasks/jobs:
* predictions from a model under test --> "predictions"
* [optionally, if true human-sourced ground-truth was missing] predictions from a reference model  --> "annotations"
* comparison of predictions and annotations to obtain a quantitative measure of how close the model under test performed w.r.t. to either the human annotator or the reference model.

This meant launching at least 1 inference job to get predictions per evaluation. This started to change with #328. This issue would wrap that up by having evaluation only ingest data (the actual annotations+predictions), rather than triggering jobs to generate the data.

**Describe the solution you'd like**
An evaluation job should only require data (a prediction from a model under test and some annotations, regardless of what the source was).

**Describe alternatives you've considered**
Part of the discussion is whether to include here a check that annotations and predictions did not come from the same model. Should this be done at experiment level?

**Additional context**
Draft work to collect feedback can be found in #333 .

ividal:
The scope of this is as discussed originally with @veekaybee, prior to any decisions in #437.

Should something need to change after that, it will be captured [here](https://github.com/orgs/mozilla-ai/projects/6/views/1?pane=issue&itemId=89419116).

veekaybee:
Per our discussion, acknowledging that for the scope of this I'll be:

1. Picking up the draft PR I had created in https://github.com/mozilla-ai/lumigator/pull/333 - I'll likely be closing out this PR since the codebase has shifted significantly since then
2. Splitting out evaluation into its own job
3. This work is independent of other ongoing work on evaluation.
4. @peteski22 and @javiermtorres will be buddies on this task, reviewing a draft of the PR.

veekaybee:
As a heads-up in the last iteration of this work, we ended up creating an eval-lite which shadowed evaluation in anticipation of breaking APIs for upcoming demos. Given that inference is now fully built out and we have more time to test, I'll be replacing evaluation entirely instead of duplicating flows to reduce code duplication and errors.

veekaybee:
Adding information about inputs and outputs of job:

![image](https://github.com/user-attachments/assets/b6f6e1da-7feb-492e-852b-f597ada00603)

ividal:
Thanks for this start @veekaybee 
Something to start the discussion and also help start designing a workflow:

[...]

Issue #478, opened 2024-12-09
[FEATURE] Dataset upload and validation


ividal:
Users should be able to add a custom dataset for a given task and know if it is not correctly formatted. This means the system should make a few preliminary checks and return a warning or error in case of trouble.

Checks during the MVP are limited to:
* format (pydantic model)
* [tbd] size

Open work:
* UI: the system confirms that the dataset is ready for use in experiments
* UI: Users can upload a dataset in the supported format (CSV)
* UI: the user gets feedback if they upload an invalid file

PR #523, opened 2024-12-17
Backend: Replacing evaluation module that currently does eval+inference with just evaluation (Issue 435)

veekaybee:
# What's changing

We previously ran inference and evaluation together in an `evaluation` job. This breaks apart these jobs, removes all inference from evaluation, and leaves an evaluation job that takes a HuggingFace dataset as input to run evaluation and returns results as JSON outputs.  

See complete requirements here https://github.com/mozilla-ai/lumigator/issues/435#issuecomment-2532571026

and discussion here:
https://github.com/mozilla-ai/lumigator/issues/435

# How to test it

Steps to test the changes:

1. `make local-up`
2. Upload a test dataset 
3. Hit the run evaluation job endpoint

```bash
curl -X 'POST' \
  'http://localhost:8000/api/v1/jobs/evaluate/' \
  -d '{
  "name": "hf://facebook/bart-large-cnn",
  "description": "",
  "dataset": "4dacfe8f-3d97-4da7-b0d0-37950585f595",
  "max_samples": -1,
  "model_url": "hf://facebook/bart-large-cnn",
  "system_prompt": "string",
  "config_template": ""
}'
``` 

Key changes include: 

1. Removal of all vestigal old evaluation module in favor of template that follows `inference` pattern for submitting jobs to Ray
2. Additional tests specifically for inputs passed between Lumigator and Ray
3. Removal of all inference patterns from current evaluator code

# I already...


.gitignore
```diff
@@ -177,4 +177,3 @@ cython_debug/
 .ijwb
 *.db
 *.db-journal
-.DS_Store
```
veekaybee, line 180: remove redundant copy of this, we have two in gitignore currently https://github.com/mozilla-ai/lumigator/blob/0ccd41f40d8fe85b1af580c546893da95b2c9cb8/.gitignore#L92

lumigator/python/mzai/backend/backend/ray_submit/submission.py
```diff
@@ -17,7 +17,7 @@ class RayJobEntrypoint(ABC):
     memory: int | float | None = None
 
     @property
-    def command_with_params(self) -> str:
+    def get_command_with_params(self) -> str:
```
veekaybee, line 20: Changing noun to verb for method

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -261,12 +263,16 @@ def create_job(
         # (request) and on the job type
         config_params = self._get_job_params(job_type, record, request)
 
+        loguru.logger.info(f"sending config_params...{config_params}")
+
         # load a config template and fill it up with config_params
-        if request.config_template is not None:
+        if request.config_template is not None and request.config_template != "":
```
veekaybee, line 269: this failed if you input a config template that's an empty string, putting this here so it saves the next person time, will add a unit test around it as well.

lumigator/python/mzai/backend/backend/settings.py
```diff
@@ -48,6 +48,8 @@ class BackendSettings(BaseSettings):
     # Eval job details
     EVALUATOR_WORK_DIR: str | None = None
     EVALUATOR_PIP_REQS: str | None = None
+    # TODO: change once we remove old eval
+    NEW_EVALUATOR_COMMAND: str = "python evaluator/evaluator.py"
```
veekaybee, line 52: Here's the change point for evaluation

ruff.toml
```diff
@@ -77,6 +77,7 @@ ignore = [
   "D415", # First line should end with a period, question mark, or exclamation point
   "D205",  # 1 blank line required between summary line and description
   "B905", # `zip()` without an explicit `strict=` parameter
+  "ISC001" # cause conflicts when used with the formatter
```
veekaybee, line 80: got a formatting issue during linting so adding this check

lumigator/python/mzai/jobs/evaluator_lite/README.md
```diff
@@ -0,0 +1 @@
+# Lite Eval
```
veekaybee, line 1: leaving this empty for now as we'll migrate it to the main eval docs

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/tests/conftest.py
```diff
@@ -0,0 +1,23 @@
...
+    with Path.open(path) as file:
+        return json.load(file)
+
+
+def resources_dir() -> Path:
+    return Path(__file__).parent / "data"
+
+
+@pytest.fixture(scope="session")
+def json_config_minimal() -> dict:
```
ividal, line 17: Are these used in the tests? (Maybe the corresponding test_* was not committed?)

ividal (commented):
Thanks for this @veekaybee !

Minor comments now, mostly thinking about how to test the new job.

ividal:
The PR description mentions hitting the evaluator endpoint to test, however, what I'm seeing in this PR is what we discussed (= no change to the route for now, just introduce the new job without a switch).

Did I miss anything or is it just that the PR description is just missing a "change the command..."?
Else, how can one test the job?

Thanks!

lumigator/python/mzai/backend/backend/settings.py
veekaybee: In this case we don't need it since it's not a module anymore. Historically, `-m` was used to execute Python modules at the package level. We could argue that we still have a package since inference contains inference and the Pydantic types, but the implication when we run -m is that we have a complex bundle of code that has an entrypoint. That's no longer the case, so we can delegate to the simpler Python command and keep in sync with the `inference` command.

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/tests/conftest.py
veekaybee: Good catch, I'll leave these changes out for now.

veekaybee:
> The PR description mentions hitting the evaluator endpoint to test, however, what I'm seeing in this PR is what we discussed (= no change to the route for now, just introduce the new job without a switch).
> 
> Did I miss anything or is it just that the PR description is just missing a "change the command..."? Else, how can one test the job?
> 
> Thanks!

I'd left a comment in the "Additional notes for reviewers" that I'll be adding routes in a follow-up PR - happy to do it in this one but since the PR is already fairly large, for the sake of legibility and we'd like to merge it and continue, I left it out for now. LMK what you think.

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/tests/conftest.py
veekaybee: Resolved these changes and added additional unit tests

Makefile
```diff
@@ -132,11 +132,15 @@ else
 	$(call run_with_existing_containers, test-backend-integration-target)
 endif
 
+test-jobs-unit:
+	cd lumigator/python/mzai/jobs/evaluator_lite; \
+	uv run pytest -o python_files="evaluator_lite/tests/test_*.py"
```
veekaybee, line 137: @ividal @aittalam @agpituk FYI, I'm adding additional unit test checks for the new jobs we're adding because they're not currently covered by the back-end checks. Right now they only cover evaluator_lite, since, if we expand the scope they will also test evaluator, which we are deprecating soon, and testing evaluator requires ray, which we don't want. So excluding to keep it simple.

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/eval_lite.py
```diff
@@ -0,0 +1,98 @@
...
+import click
+import s3fs
+from datasets import load_from_disk
+from eval_metrics import EvaluationMetrics
+from evaluate_config import EvalJobConfig
+from loguru import logger
+
+
+def save_to_disk(local_path: Path, data_dict: dict):
+    logger.info(f"Storing into {local_path}...")
```
dpoulopoulos, line 13: Make logs a bit more specific. For example:

```diff
logger.info(f"Storing evaluation results into {local_path}...")
```

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/eval_lite.py
```diff
@@ -0,0 +1,98 @@
...
+        json.dump(data_dict, f)
+
+
+def save_to_s3(config: dict, local_path: Path, storage_path: str):
+    s3 = s3fs.S3FileSystem()
+    if storage_path.endswith("/"):
+        storage_path = "s3://" + str(
+            Path(storage_path[5:]) / config.get("name") / "eval_results.json"
+        )
+    logger.info(f"Storing into {storage_path}...")
```
dpoulopoulos, line 25: Same:

```
logger.info(f"Storing evaluation results into {local_path}...")
```

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/eval_lite.py
```diff
@@ -0,0 +1,98 @@
...
+    logger.info(f"Retrieving {config.dataset.path} for evaluation")
+    if max_samples:
+        logger.info(f"max_samples ({max_samples}) resized to dataset size ({len(dataset)})")
+        # select data between the minimum and total length of dataset
+        num_samples = range(min(max_samples, len(dataset)))
+        dataset = dataset.select(num_samples)
+
+    # run evaluation and append to results dict
+    predictions = dataset["predictions"]
+    ground_truth = dataset["ground_truth"]
```
dpoulopoulos, line 71: At this point, should we check if there is a `ground_truth` column? Is this something we want to do here or in #478?

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/eval_lite.py
```diff
@@ -0,0 +1,98 @@
...
+
+    logger.info(f"Predictions: {predictions}")
+    logger.info(f"Ground truth: {ground_truth}")
```
dpoulopoulos, line 74: Do we need these logs? If the dataset is huge, they will just pollute stdout.

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/eval_lite.py
```diff
@@ -0,0 +1,98 @@
...
+    # add input data to results dict
+    if config.evaluation.return_input_data:
+        evaluation_results["predictions"] = dataset["predictions"]
+        evaluation_results["ground_truth"] = dataset["ground_truth"]
+
+    output_path = save_outputs(config, evaluation_results)
+    return output_path
+
+
+@click.command()
```
dpoulopoulos, line 87: Do we still want to use `click` and treat this as a CLI? In `inference` jobs we're just treating this as a Python script.

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/eval_metrics.py
```diff
@@ -0,0 +1,82 @@
...
+        # unsupported metrics are the difference between the provided and the supporterd ones
+        self._unsupported_metrics = set(metrics) - set(self._supported_metrics.keys())
+
+        if len(self._chosen_metrics) == 0:
+            logger.info("No valid metrics selected")
+        else:
+            logger.info(f"Chosen metrics: {self._chosen_metrics}")
+
+        if len(self._unsupported_metrics) > 0:
+            logger.info(f"Unsupported metrics: {self._unsupported_metrics}")
```
dpoulopoulos, line 25: I suggest turning this into a `warning`. This is something that the user should pay attention to: "You have selected to run an evaluation job with a few unsupported metrics. I'm removing these and continue only with the supported ones."

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/requirements.txt
```diff
@@ -0,0 +1,24 @@
...
+pydantic-yaml>=1.2.0
+pydantic>=2.6.4
+ray[default]==2.30.0
+ruff==0.5.5
+s3fs==2024.5.0
+scipy==1.13.1
+sentencepiece==0.2.0
+transformers==4.43.4
+urllib3>=1.26.18,<2
+wandb==0.16.3
```
dpoulopoulos, line 24: I think we're not using the following deps anymore:

```
bitsandbytes
einops
lm-eval
nltk
protobuf
```

Can you check if we still need any of these?

Also, do we need to install `ruff`? And what about `wandb`? I see we're importing it in a few Python files. Maybe, we should note that this is something to remove later.

dpoulopoulos (changes requested):

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/eval_lite.py
ividal: If in #478 : the user needs to say what task they want the dataset for when they upload the dataset. Or we can add checks for multiple tasks (summarization, Q&A, translation...) and the response of data validation can say which it found gt for, so the UI can show that. 

If here, an evaluation task could fail due to lack of suitable gt. Since eval jobs are meant to be async and the user may just come by some time later to see what the results were, it may be too late to warn them.

I lean towards #478 ...

veekaybee: Agree, let's group it with 478

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/eval_lite.py
veekaybee: The script itself is a Python script but it's passed to Ray using CLI args which then are processed in the script itself, so we can't move away from this paradigm entirely unfortunately in the Ray universe.   In this case, Click is an alternative to argparse, which we currently use and have discussed moving away from: https://github.com/mozilla-ai/lumigator/pull/333#discussion_r1825948491.

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/requirements.txt
veekaybee: `Einops` is a transformers/torch dependency that we had issues without at one point, `lm-eval`, we don't currently use it but we do have an entrypoint for it, so we should either remove that or completely develop it, `nltk` we are using, `protobuf` we need on the ray side, and `bitsandbytes` is also needed on the ray side for PyTorch optimization. `wandb` is a carryover from the old evaluator, so we don't need it. 

Thanks for checking!

lumigator/python/mzai/jobs/evaluator_lite/eval_config.py
```diff
@@ -0,0 +1,30 @@
...
+class DatasetConfig(BaseModel):
+    path: str
+
+
+class ModelConfig(BaseModel):
+    path: str
+
+
+class EvaluationConfig(BaseModel):
+    metrics: list[str] = Field(default=["rouge", "meteor", "bertscore"])
```
aittalam, line 13: I was checking the evaluator config for comparison and I like having these defaults instead of limiting the length to 1 value.
Looking for pydantic types I think this could perhaps even be a set/conset, as the same eval metric never appears more than once. WDYT?

lumigator/python/mzai/jobs/evaluator_lite/eval_config.py
```diff
@@ -0,0 +1,30 @@
...
+    use_pipeline: bool = False
+    max_samples: int = 0
+    return_input_data: bool = False
+    return_predictions: bool = False
+    storage_path: str = ""
+
+
+class TokenizerConfig(BaseModel):
```
aittalam, line 21: I think this can be removed, as there's no inference stuff in EvalJobConfig anymore

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/eval_lite.py
aittalam: Good point!
While I think it would be nice if the job returned an easy-to-interpret error if one required field is not found, this would be too late if only trapped here. +1 for https://github.com/mozilla-ai/lumigator/issues/478!

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/requirements.txt
aittalam, line 1: I think a few libraries (at a glance accelerate, bitsandbytes, mistralai, openai, sentencepiece) here will not be necessary as we are not running inference anymore. Transformers might be a dependency for bert_score, wondering whether we want to keep it explicit here or just pull the one that works best for whatever version of bertscore we'll use (even if I'd lean more towards pinning the full set of libraries we know work well together)

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/eval_lite.py
aittalam: My bad for not having done it yet!
After seeing how the two jobs (inference+lite_eval) look like and how simple their parameters are, I think there are probably no /strong/ reasons to choose one approach over another. I am happy to move to click for consistency after the PR is merged though!

lumigator/python/mzai/backend/backend/settings.py
aittalam, line 52: Two questions:
- why not `evaluator_lite` for now? I am missing how the job is run from the service otherwise
- why the nested `evaluator_lite` directory? As it is called as a simple python script and not a module, we should need neither this nor the `__init__.py` file (similarly to what we do in inference)

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/eval_metrics.py
```diff
@@ -0,0 +1,82 @@
...
+    def __init__(self, metrics):
+        self._supported_metrics = {
+            "rouge": self._rouge,
+            "meteor": self._meteor,
+            "bertscore": self._bertscore,
+        }
+
+        # chosen metrics are the intersection between the provided and the supported ones
+        self._chosen_metrics = set(metrics) & set(self._supported_metrics.keys())
+        # unsupported metrics are the difference between the provided and the supporterd ones
```
aittalam, line 16:
```suggestion
        # unsupported metrics are the difference between the provided and the supported ones
```

Found my own typo :D

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/eval_metrics.py
```diff
@@ -0,0 +1,82 @@
...
+        for p, r in zip(pred, ref):
+            evals["meteor"].append(ev.compute(predictions=[p], references=[r])["meteor"])
+
+        # calculate mean
+        evals["meteor_mean"] = np.mean(evals["meteor"])
+
+        return evals
+
+    def _bertscore(self, pred: list, ref: list):
+        """https://huggingface.co/spaces/evaluate-metric/bertscore
```
aittalam, line 55: TY for adding typing and comments! 🙏

lumigator/python/mzai/backend/backend/settings.py
veekaybee: There's no service implemented yet, as I mentioned here https://github.com/mozilla-ai/lumigator/pull/523#issuecomment-2568062486 and here https://github.com/mozilla-ai/lumigator/pull/523#issuecomment-2554891457.  We had previously discussed creating evaluator and lite in parallel and then switching over to the new command. 

I'm happy to add a new lite service call in this PR but excluded it for now as I felt it was getting large to review already and we wanted to merge what exists so far, not break functionality and keep going. 

For the second piece, great point - I'll remove the nesting. I included `__init__.py` for consistency with the rest of the application and because I've seen enough cases across Python codebases where it was excluded and led to issues at import time. You're right though that since we consider it separate in this case we could exclude in this case to make consistent with inference.

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/eval_lite.py
veekaybee: Sounds good, I added that as an issue and can take it, or alternatively it would make a small nice onboarding task for our new joiners. https://github.com/mozilla-ai/lumigator/issues/560

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/eval_lite.py
```diff
@@ -0,0 +1,95 @@
...
+            Path(storage_path[5:]) / config.get("name") / "eval_results.json"
+        )
+    logger.info(f"Storing evaluation results into {local_path}...")
+    s3.put_file(local_path, storage_path)
+
+
+def save_outputs(config: dict, eval_results: dict) -> Path:
+    storage_path = config.get("evaluation").get("storage_path")
+
+    local_path = Path("eval_results.json")
```
aittalam, line 32:
```suggestion
    local_path = Path("results.json")
```

If we want to change the default local path where results are stored, it'd be useful to raise a flag (eg an issue) so we can make this consistent with the only other job, i.e. inference.

lumigator/python/mzai/jobs/evaluator_lite/eval_config.py
veekaybee: Are you thinking something like this? 

`metrics: Set[str] = {"rouge", "meteor", "bertscore"}`

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/requirements.txt
veekaybee: I lean towards pinning and in fact might make a separate issue to test some of these assumptions I made here by running the jobs with the reduced set of dependencies https://github.com/mozilla-ai/lumigator/pull/523#discussion_r1901764171

Makefile
veekaybee: Addressed for now with https://github.com/mozilla-ai/lumigator/issues/559

lumigator/python/mzai/backend/backend/settings.py
veekaybee: Second part addressed with https://github.com/mozilla-ai/lumigator/pull/523/commits/bab4fd6701337172990ffedfbf56e79b9d1b4bce

lumigator/python/mzai/jobs/evaluator_lite/eval_config.py
veekaybee: we won't need this for bertscore right?

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/eval_lite.py
aittalam: That's great, thanks!

lumigator/python/mzai/jobs/evaluator_lite/eval_config.py
aittalam: Yep but it was mostly ranting about what further automatic checks we could put in there :D
I realise that changing type might break things around though (I am thinking about all config templates, tests, etc) so it might make sense to keep it a list and think if and how we want to improve this later. WDYT?

lumigator/python/mzai/backend/backend/settings.py
aittalam: TY! And sorry I missed the previous comments, I agree on merging as-is without changing / adding routes (if we merge the new experiments endpoint we'll be able to test it as a service without needing an ad-hoc route, this might allow us to parallelize some work)

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/eval_lite.py
veekaybee: Addressed with https://github.com/mozilla-ai/lumigator/pull/523/commits/1d81e87220820a3f2ee886f319810f992ef8ec49

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/eval_lite.py
```diff
@@ -0,0 +1,95 @@
...
+    local_path.parent.mkdir(exist_ok=True, parents=True)
+    with local_path.open("w") as f:
+        json.dump(data_dict, f)
+
+
+def save_to_s3(config: dict, local_path: Path, storage_path: str):
+    s3 = s3fs.S3FileSystem()
+    if storage_path.endswith("/"):
+        storage_path = "s3://" + str(
+            Path(storage_path[5:]) / config.get("name") / "eval_results.json"
```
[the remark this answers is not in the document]
veekaybee: Addressed with https://github.com/mozilla-ai/lumigator/pull/523/commits/1d81e87220820a3f2ee886f319810f992ef8ec49

lumigator/python/mzai/jobs/evaluator_lite/evaluator_lite/eval_metrics.py
veekaybee: Addressed with https://github.com/mozilla-ai/lumigator/pull/523/commits/1d81e87220820a3f2ee886f319810f992ef8ec49

aittalam (approved):
TY Vicki for this PR!
I tested the PR with `make test-all` and it worked fine. I understand there are not many other ways to test the code now as it is not wired to the API but I am happy to merge this as it does not break code and iterate on that, testing it as a standalone job and wiring it to the new experiments service as soon as it is merged.

This should allow us to verify everything works without tampering with the existing code, then move names and deprecate the older evaluator.

dpoulopoulos (approved):

PR #524, opened 2024-12-17
Add installation file for ubuntu and osx

davidmanzanoai:
# What's changing

The script allows in a easy way to set up the local environment for starting the development or local usage of the tool. It secures that docker, docker-compose and even git is installed or otherwise do it for you. Once downloaded the script, it requires to add chmod +x install-lumigator.sh. After, you decide how to execute it:
- l local files: your lumigator code is within the current folder or in the folder set with -d option.
- m means installation method, in case you don't have the code or it is a fresh installation it will download the code from github (using your credentials) or with a zip file containing the whole repo. It takes care of the folder in case this already exits. If it is, it is possible to force overwritting the folder with a fresh installation using "-o" 

* Updated section...
* Added new...
* Fixed a typo in...
* ...

Refs #...
Closes #...

# How to test it

Steps to test the changes:

1. Download the script and run chmod +x 
2. ./instal-lumigator.sh -m zip -d FOLDER_NAME OR .install-lumigator.sh -m git -d FODER_NAME
3. Check it in the website if lumigator UI is running

# I already...

### DIFF
--- a/install_lumigator.sh
+++ b/install_lumigator.sh
@@ -3,7 +3,7 @@
 # This script supports the initial setup of Lumigator for developing and using all functionalities locally.
 # It requires Docker and Docker Compose to run. If they are not present on your machine, the script will install and activate them for you.
 # If the local option (`-l`) is selected, Lumigator code is expected to be located in the current folder or in the provided folder.
-# If a method is selected, it will install Lumigator using Git or a Zip file, depending on the case, and will place the files in selected_folder/lumigator.
+# If the installation method (`-m`) is selected, it will install Lumigator using Git or a Zip file, depending on the case, and will place the files in selected_folder/lumigator.
 
 # Help
 show_help() {
--- a/install_lumigator.sh
+++ b/install_lumigator.sh
@@ -1,0 +1,274 @@
+#!/bin/bash
+
+# This script supports the initial setup of Lumigator for developing and using all functionalities locally.
+# It requires Docker and Docker Compose to run. If they are not present on your machine, the script will install and activate them for you.
+# If the local option is selected, Lumigator code is expected to be located in the current folder or in the provided folder.
+# If a method is selected, it will install Lumigator using Git or a Zip file, depending on the case, and will place the files in selected_folder/lumigator.
+
+# Help
+show_help() {
+    echo "Usage: $0 [OPTIONS]"
+    echo "Starts Lumigator by checking your setup or installing it."
+    echo ""
+    echo "Options:"
+    echo "  -d, --directory DIR   Specify root directory (default: current directory)"
+    echo "  -m, --method METHOD   Download method: 'git' or 'zip' (default: zip)"
+    echo "  -o, --overwrite       Overwrite existing directory (lumigator)"
+    echo "  -l, --local           Use local files in the same directory as the script (execute everything from the current directory without downloading anything)"
+    echo "  -h, --help            Display this help message"
+    exit 0
+}
+
+# Install Git on Linux
+install_git_linux() {
+    if ! command -v git &> /dev/null
+    then
+        echo "Git not found. Installing Git..."
+        sudo apt-get update
+        sudo apt-get install -y git
+        echo "Git installed successfully."
+    else
+        echo "Git is already installed."
+    fi
+}
+
+# Install Git on macOS
+install_git_macos() {
+    if ! command -v git &> /dev/null
+    then
+        echo "Git not found. Installing Git..."
+        # Check if Homebrew is installed
+        if ! command -v brew &> /dev/null
+        then
+            /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+        fi
+        brew install git
+        echo "Git installed successfully."
+    else
+        echo "Git is already installed."
+    fi
+}
+
+# Install Docker on Linux
+install_docker_linux() {
+    if ! command -v docker &> /dev/null
+    then
+        echo "Docker not found. Installing Docker..."
+        sudo apt-get update
+        sudo apt-get install -y apt-transport-https ca-certificates curl software-properties-common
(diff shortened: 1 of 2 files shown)


install_lumigator.sh
```diff
@@ -0,0 +1,274 @@
+#!/bin/bash
+
+# This script supports the initial setup of Lumigator for developing and using all functionalities locally.
+# It requires Docker and Docker Compose to run. If they are not present on your machine, the script will install and activate them for you.
+# If the local option is selected, Lumigator code is expected to be located in the current folder or in the provided folder.
```
Kostis-S-Z, line 5:
```suggestion
# If the local option (`-l`) is selected, Lumigator code is expected to be located in the current folder or in the provided folder.
```

install_lumigator.sh
```diff
@@ -0,0 +1,274 @@
...
+# If a method is selected, it will install Lumigator using Git or a Zip file, depending on the case, and will place the files in selected_folder/lumigator.
```
Kostis-S-Z, line 6:
```suggestion
# If the installation method (`-m`) is selected, it will install Lumigator using Git or a Zip file, depending on the case, and will place the files in selected_folder/lumigator.
```

davidmanzanoai:
> Thanks for this @davidmanzanoai ! I've got some general questions about the approach, more than the technical implementation. If we're trying to have a way for people to install lumigator easily, I think this tool shouldn't clone the repository at all (as it's not needed to run lumigator!) My favored approach here would be to have a script like this one, that pulls the docker-compose & run the start-lumigator command, we probably would like to have a specific docker-compose as well for that one with pre-initiated variables. What are your thoughts?

@agpituk What we can do is to set up the script in a new folder (e.g. install) and we can use the link to the script to be used for installing everything. The default choice will be to take the zip file and use the package from the package repositories instead of building locally (that option will remain accesible through out the Makefile).

javiermtorres (approved):
LGTM. I have tried it on WSL and it works. I'd propose enhancing this to just retrieve the docker compose files and let it download the necessary images from the image repos.

agpituk:
> > Thanks for this @davidmanzanoai ! I've got some general questions about the approach, more than the technical implementation. If we're trying to have a way for people to install lumigator easily, I think this tool shouldn't clone the repository at all (as it's not needed to run lumigator!) My favored approach here would be to have a script like this one, that pulls the docker-compose & run the start-lumigator command, we probably would like to have a specific docker-compose as well for that one with pre-initiated variables. What are your thoughts?
> 
> @agpituk What we can do is to set up the script in a new folder (e.g. install) and we can use the link to the script to be used for installing everything. The default choice will be to take the zip file and use the package from the package repositories instead of building locally (that option will remain accesible through out the Makefile).

Ok, happy with that path. We'd need to update the docs as well

davidmanzanoai:
@agpituk  @javiermtorres  @ividal  Could you please take a look. In OS X, if docker isn't installed, it stoped and points the user to a manual installation. Otherwise it sets up everything for the user and start the browser.  We might add the file into a folder if you prefer

start-lumigator.sh
```diff
@@ -0,0 +1,242 @@
...
+        fi
+    else
+        # Installation directory created, didn't exist
+        mkdir -p "$TARGET_DIR"
+    fi
+
+    # Download based on method
+ 
+        echo "Downloading ZIP file..."
+        curl -L "${REPO_URL}/archive/main.zip" -o lumigator.zip
```
javiermtorres, line 179: This needs to be a release link at some point. Maybe just after MVP, when we do have a tag and a release.

start-lumigator.sh
```diff
@@ -0,0 +1,242 @@
...
+LUMIGATOR_URL="http://localhost:80"
+
+
+# Command line arguments
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        -d|--directory)
+            ROOT_DIR="$2"
+            shift ;;
+        -m|--method)
```
javiermtorres, line 137: Is this needed anymore?

start-lumigator.sh
```diff
@@ -0,0 +1,242 @@
...
+            if [[ "$2" != "git" && "$2" != "zip" ]]; then
+                echo "Invalid method. Use 'git' or 'zip'."
+                exit 1
+            fi
+            download_method="$2";
+            shift ;;
+        -o|--overwrite) OVERWRITE=true ;;
+        -l|--local) local_option=true ;;
```
javiermtorres, line 145: Is this needed anymore?

javiermtorres (approved):
I didn't check it this time, but left some comments. Just check that it really works :)

start-lumigator.sh
```diff
@@ -0,0 +1,234 @@
...
+}
+
+
+
+# Install Docker on Linux
+install_docker_linux() {
+    if ! command -v docker &> /dev/null
+    then
+        echo "Docker not found. Installing Docker..."
+        sudo apt-get update
```
ividal, line 25: Argh. Needing sudo rights. Many devs at companies will have docker available to them through their IT but very likely not have sudo rights to the machine they're using.

Not necessary to merge, but we could have a top level script that asks/checks and then calls:
* (optional) one that requires sudo and takes care of pre-requirements like docker.
* a second one with everything else, which assumes you have docker correctly set up for you.

start-lumigator.sh
```diff
@@ -0,0 +1,234 @@
...
+        echo "Docker installed. Please complete the installation in the Docker Desktop app."
+    else
+        echo "Docker is already installed."
+    fi
+}
+
+
+# Install Docker Compose
+install_docker_compose() {
+    if ! command -v "$BREW_PATH/docker-compose" &> /dev/null
```
ividal, line 71: Although I did have docker compose installed this still didn't detect it.

ividal (approved):
Pretty handy and worked well on both OS and Ubuntu, except for the docker-compose detection which did not find it on either.

PR #540, opened 2024-12-19
#485 Build workflow calling inference + evaluation jobs sequentially

aittalam:
# What's changing

This is the first PR tackling https://github.com/mozilla-ai/lumigator/issues/485.
Here we introduce a new `/experiments_new` endpoint that has the same semantics of `/experiments`, but instead of running a single job using evaluator it runs

- an inference job
- code to save the inference job results into a dataset
- a "lite evaluation" job to run evaluation given an input dataset with GT and predictions

We rely on fastapi's [BackgroundTasks](https://fastapi.tiangolo.com/tutorial/background-tasks/) to run the job pipeline asynchronously. As an image is worth 2^10 words:

image

Refs https://github.com/mozilla-ai/lumigator/issues/485

# How to test it

- Upload a dataset (with samples + GT)
- Hit the `/evaluation_new` endpoint with the same syntax previously used for `/evaluation`

Scripts used for testing are available [here](https://github.com/user-attachments/files/18214031/scripts.zip)

Unit / integration tests: run `make test-all`

- lite eval is mocked by adding an extra parameter to evaluator (`skip_inference`), ideally the inputs/outputs should be the same as lite eval but I appreciate (and will address) any flags raised if there's any incompatibility
- WHAT IS NOT HERE: in the spirit of keeping this PR small, I only provided implementation for POSTing a new experiment. Other functionalities (i.e. get/list experiment runs, check exp status, etc) will be developed in a follow-up PR and tested all together.

# I already...

- [ ] Added some tests for any new functionality => mostly adapted existing tests, see "what is not here" above for new functionalities
- [ ] Updated the documentation (both comments in code and [product documentation](https://mozilla-ai.github.io/lumigator) under `/docs`) => will update it (if necessary) when we switch from old to new `/experiments` endpoint
- [x] Checked if a (backend) DB migration step was required and included it if required => no changes to the DB schema


lumigator/python/mzai/backend/backend/api/router.py
```diff
@@ -12,3 +20,6 @@
 api_router.include_router(experiments.router, prefix="/experiments", tags=[Tags.EXPERIMENTS])
 api_router.include_router(completions.router, prefix="/completions", tags=[Tags.COMPLETIONS])
 api_router.include_router(models.router, prefix="/models", tags=[Tags.MODELS])
+api_router.include_router(
+    experiments_new.router, prefix="/experiments_new", tags=[Tags.EXPERIMENTS_NEW]
```
ividal, line 24: This is where it could make sense to introduce API versioning, since we´re touching known endpoints, but I would leave it out of the scope of this PR and focus on the logic/job stitching first.

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -311,10 +320,15 @@ def create_job(
...
-        loguru.logger.info("Submitting Ray job...")
+        loguru.logger.info("Submitting {job_type} Ray job...")
         submit_ray_job(self.ray_client, entrypoint)
 
-        background_tasks.add_task(self._watch_job, record.id, request)
+        if job_type == JobType.INFERENCE:
+            # Inference jobs produce a new dataset
+            # Add the dataset to the (local) database
+            background_tasks.add_task(
+                self._run_after_job, record.id, self._add_dataset_to_db, record.id, request
```
ividal, line 330: This is the only bit I'm not too comfortable with. We now have a job modifying a queue created at experiment (i.e. orchestrator) level.

On the one side, I understand it can only add jobs, never amend or remove, and it removes the need for the orchestrator to know details (e.g. output data fields) of the job.

On the other, which I feel is stronger, it smears the responsibility division between orchestrator and job, forcing the job creator to know how lumigator task queues and, say, data storage in the DB work. If/when anything in the flow fails, it would also be a bit harder to debug...

Seeing the need to start testing this workflow/composite task and the fact we're discussing full orchestration elsewhere, perhaps what belongs in this PR is the outline of what problem keeping all modifications of the background_tasks at experiment level has.

WDYT?

lumigator/python/mzai/jobs/evaluator/evaluator/jobs/evaluation/hf_evaluate.py
```diff
@@ -123,8 +123,12 @@ def run_eval(config: HuggingFaceEvalJobConfig) -> Path:
             logger.info(f"Using direct HF model invocation. Model: {model_name}")
             model_client = HuggingFaceModelClient(model_name, config)
 
-    # run inference
-    predictions, summarization_time = predict(dataset_iterable, model_client)
+    if not config.evaluation.skip_inference:
```
ividal, line 126: Nit, as discussed... 
if skip: take from dataset
else: predict(...) 
?

lumigator/python/mzai/backend/backend/config_templates.py
```diff
@@ -11,7 +11,8 @@
         "max_samples": {max_samples},
         "return_input_data": true,
         "return_predictions": true,
-        "storage_path": "{storage_path}"
+        "storage_path": "{storage_path}",
+        "skip_inference": "{skip_inference}"
```
ividal, line 15: I want to say something against this, but... won't 😆 

Also, now seriously, this might be handy for #435 to compare against, since essentially they need to return the same thing.

ividal (commented):
Thanks a lot @aittalam - this is much needed!
The only bit that worries me is the queue modification at job level thing and worry it might be one of those temporary situations that stays.

aittalam:
> Thanks a lot @aittalam - this is much needed! The only bit that worries me is the queue modification at job level thing and worry it might be one of those temporary situations that stays.

Thanks for the comment @ividal ! I see your point and I pushed a commit that moves all the logic into the EvaluationService, adding some thoughts about pros/cons.

lumigator/python/mzai/backend/backend/services/jobs.py
aittalam: I see your point and I agree we can start using it at experiment level, then verify if and how much it limits us. I have added a few pros/cons to the related commit to summarize my experience with it

lumigator/python/mzai/jobs/evaluator/evaluator/jobs/evaluation/hf_evaluate.py
aittalam: TY, fixed in [here](https://github.com/mozilla-ai/lumigator/pull/540/commits/6ea76eed2049e471f2dc36d0a83b60f75bbbbf2e)

lumigator/python/mzai/backend/backend/config_templates.py
aittalam: :D yeah, this was really hacky! It is also why I tried to keep everything into a single commit, we might revert just that if for some reason we want to keep the original evaluator after we have already introduced the light eval

lumigator/python/mzai/backend/backend/services/datasets.py
```diff
@@ -176,6 +179,13 @@ def get_dataset(self, dataset_id: UUID) -> DatasetResponse | None:
 
         return DatasetResponse.model_validate(record)
 
+    def get_dataset_by_job_id(self, job_id: UUID) -> DatasetResponse | None:
```
ividal, line 182: How do you plan on using this?

lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
```diff
@@ -45,20 +42,15 @@ def test_upload_data_launch_job(local_client: TestClient, dialog_dataset):
...
 
-    # Mocking the background task because the test hangs indefinitely,
-    # waiting for the background task to finish.
-    with patch("fastapi.BackgroundTasks.add_task") as mock:
-        create_evaluation_job_response = local_client.post(
-            "/jobs/evaluate/", headers=headers, json=payload
-        )
-        assert create_evaluation_job_response.status_code == 201
-        assert mock.called
+    create_evaluation_job_response = local_client.post(
```
ividal, line 45: Not sure I'm understanding this integration test: it runs an evaluation before inference?
Wouldn't we want a test where the evaluation takes the output of that inference? Not necessarily the quality of the predictions (I agree that wouldn't be the point), but that the inference job inserted the new dataset and the evaluation job can successfully retrieve it.

lumigator/python/mzai/backend/backend/services/datasets.py
aittalam: Good point! I am not using it in this PR, but as we have datasets generated by inference jobs I wanted to expose, at service level, the possibility to get the dataset ID given the job ID. 

To surface this to the user we still need to add an endpoint in the API and I was planning to introduce this in the following PR where I'll make sure the `/experiments` API is complete and compatible with the previous version of it.

lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
aittalam: +100, I agree this is how we should test the new endpoint.
The reason why you see those calls is that they are still testing `/jobs`, and the two jobs are independent of each other. My changes to these tests only affected the check for a background task (which are not needed anymore as we create a bg task in experiments).
My intention was to keep these tests and add a new one specifically for `/experiments` when the API is complete, but I can add something now to make everything self-contained in this PR. WDYT?

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -216,11 +217,15 @@ def _add_dataset_to_db(self, job_id: UUID, request: JobInferenceCreate):
             f"Dataset '{dataset_filename}' with ID '{dataset_record.id}' added to the database."
         )
 
-    def _watch_job(self, job_id: UUID, request: JobEvalCreate | JobInferenceCreate):
+    async def _run_after_job(self, job_id: UUID, task: Callable, *args):
```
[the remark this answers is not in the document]
aittalam: good point, will do! 🙏

lumigator/python/mzai/backend/backend/api/router.py
```diff
@@ -12,3 +20,6 @@
 api_router.include_router(experiments.router, prefix="/experiments", tags=[Tags.EXPERIMENTS])
 api_router.include_router(completions.router, prefix="/completions", tags=[Tags.COMPLETIONS])
 api_router.include_router(models.router, prefix="/models", tags=[Tags.MODELS])
+api_router.include_router(
```
veekaybee, line 23: suggestion - we might want to add #TODO everywhere we see removing the old experiments/renaming for the next steps throughout the files in this PR

lumigator/python/mzai/backend/backend/repositories/datasets.py
```diff
@@ -7,3 +10,7 @@
 class DatasetRepository(BaseRepository[DatasetRecord]):
     def __init__(self, session: Session):
         super().__init__(DatasetRecord, session)
+
+    def get_by_job_id(self, job_id: UUID) -> DatasetRecord | None:
+        statement = select(DatasetRecord).filter_by(run_id=job_id)
+        return self.session.scalar(statement)
```
veekaybee, line 16: just checking that we want a scalar here since scalars usually return a single result, i.e. 

```
count = session.query(func.count(User.id)).scalar()  # int
price = session.query(Product.price).scalar()  # float/decimal
```

and it looks like we want an entire DatasetRecord returned (https://github.com/mozilla-ai/lumigator/blob/47b5dc2b12d390e84ecf7c735b3e61ea697d37c0/lumigator/python/mzai/backend/backend/records/datasets.py#L10)

lumigator/python/mzai/backend/backend/services/experiments.py
```diff
@@ -0,0 +1,168 @@
...
+    def _add_dataset_to_db(self, job_id: UUID, request: JobInferenceCreate):
+        loguru.logger.info("Adding a new dataset entry to the database...")
+        s3 = S3FileSystem()
+
+        # Get the dataset from the S3 bucket
+        result_key = self._job_service._get_results_s3_key(job_id)
+        with s3.open(f"{settings.S3_BUCKET}/{result_key}", "r") as f:
+            results = json.loads(f.read())
+
+        dataset = {
```
veekaybee, line 39: non-blocking, suggestion to make this slightly more legible by pulling out the fields for emphasis:

```python
fields = ["examples", "ground_truth", request.output_field]
dataset = {k: v for k, v in results.items() if k in fields}
```

lumigator/python/mzai/backend/backend/services/experiments.py
```diff
@@ -0,0 +1,168 @@
...
+            k: v
+            for k, v in results.items()
+            if k in ["examples", "ground_truth", request.output_field]
+        }
+
+        # Create a CSV in memory
+        csv_buffer = StringIO()
```
veekaybee, line 46: I might pull this piece out into a separate method for legibility/easier testing

lumigator/python/mzai/backend/backend/api/router.py
aittalam: Good point! I'll do it consistently in the next PR where I'll wire experiments_new with eval_lite so I can tag both at the same time

lumigator/python/mzai/backend/backend/repositories/datasets.py
aittalam: I was also bothered by that scalar(), also because documentation is very clear about the fact that it returns "the first column of the first row". In this case, however, we are dealing with [ORM entities](https://docs.sqlalchemy.org/en/20/orm/queryguide/select.html#selecting-orm-entities) and _"when selecting from ORM entities, the entity itself is returned in the result as a row with a single element, as opposed to a series of individual columns"_, so what we are returned is the full entity, ie. the whole DatasetRecord

veekaybee: Makes sense! Not a blocker for this PR, but what do you think about adding an issue to match implementations here then (if they return the same level of results, as I understand them to be) , either make both of these scalar or both use `first` https://github.com/mozilla-ai/lumigator/blob/46168b617f6fe9e9b7fd80e41bf29744c220835f/lumigator/python/mzai/backend/backend/repositories/jobs.py#L18 ?

veekaybee (approved):
Looks good on my side, left a few notes we might want to consider in the future. Thanks for this!

PR #541, opened 2024-12-19
UI: Ensure (all) experiments which can have their status updated, get their status updated.

peteski22:
# What's changing

This PR fixes a UI bug where the status of an experiment is updated each time the page/route is navigated away and back to.

The experiments are loaded once to populate the table, then we mainly deal with updating the running jobs. 

The change here just aligns the experiment and running job status when their IDs match. It's a short term/easy fix to keep the change small.

# How to test it

Steps to test the changes:

1. Start Lumigator and the UI
2.Upload a dataset and create an experiment, then fairly quicky...
3. Navigate to datasets, then back to experiments - as the experiments table renders it shouldn't change from 'created' to 'pending' it should already show 'pending'

A cleaner potential solution would probably look at how the experiments are updated in a single place.

# I already...

### DIFF
--- a/lumigator/frontend/src/stores/experiments/store.js
+++ b/lumigator/frontend/src/stores/experiments/store.js
@@ -130,6 +130,10 @@ export const useExperimentStore = defineStore('experiment', () => {
   async function updateJobStatus(jobId) {
     try {
       const status = await experimentService.fetchJobStatus(jobId);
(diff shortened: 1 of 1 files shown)


george-mzai (approved):

peteski22:
Closing in favour of https://github.com/mozilla-ai/lumigator/pull/542

PR #543, opened 2024-12-20
Take env vars from .env or shell vars

javiermtorres:
# What's changing

The Makefile will now incorporate the `RAY_WORKER_GPUS`,  `RAY_WORKER_GPUS_FRACTION` and `INFERENCE_PIP_REQS` environment variables from the shell or the `.env` file (if it exists). This allows selective use of GPU and prevents tasks from being stuck in pending for cpu-only tests.

# How to test it

From the base directory:

1. export RAY_WORKER_GPUS="0.0"
2. export RAY_WORKER_GPUS_FRACTION="0.0"
3. make test-backend-integration-target

Tests shall pass, but the Ray dashboard will show failed jobs (instead of pending).

N/A

# I already...

  - N/A
  - N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,13 @@ UNAME:= $(shell uname -o)
 PROJECT_ROOT := $(shell git rev-parse --show-toplevel)
 CONTAINERS_RUNNING := $(shell docker ps -q --filter "name=lumigator-")
 
+-include .env
+# GPU related settings
+#
+RAY_WORKER_GPUS ?= "0.0"
+RAY_WORKER_GPUS_FRACTION ?= "0.0"
+INFERENCE_PIP_REQS ?= ../jobs/inference/requirements_cpu.txt
+
 #used in docker-compose to choose the right Ray image
 ARCH := $(shell uname -m)
 RAY_ARCH_SUFFIX :=
--- a/lumigator/python/mzai/jobs/inference/requirements_cpu.txt
+++ b/lumigator/python/mzai/jobs/inference/requirements_cpu.txt
@@ -1,0 +1,11 @@
+--extra-index-url https://download.pytorch.org/whl/cpu
+accelerate==1.1.1
+datasets==2.20.0
+loguru==0.7.2
+mistralai==0.4.2
+openai==1.55.3
+python-box==7.2.0
+s3fs==2024.5.0
+sentencepiece==0.2.0
+torch==2.5.1
+transformers==4.46.3
--- a/lumigator/python/mzai/backend/backend/tests/README.md
+++ b/lumigator/python/mzai/backend/backend/tests/README.md
@@ -30,6 +30,13 @@ class. This class inherits from the
 class and reads its values from the environment when instantiated. These settings are then used for
 instantiating various clients/controlling business logic throughout the application.
 
+The Makefile used to run tests will take the following environment variables from the environment
+or the `.env` file (with this file taking precedence) if they exist:
+
+* RAY_WORKER_GPUS (default value `0.0`)
+* RAY_WORKER_GPUS_FRACTION (default value `0.0`)
+* INFERENCE_PIP_REQS (default value `../jobs/inference/requirements_cpu.txt`)
+
 In cases where a test requires a different dependency than those defined by the `BackendSettings`
 class (e.g., for creating a test client for a mock service), a test dependency override can be
 pecified in the `conftest.py` fixture.


Makefile
```diff
@@ -5,6 +5,13 @@ UNAME:= $(shell uname -o)
 PROJECT_ROOT := $(shell git rev-parse --show-toplevel)
 CONTAINERS_RUNNING := $(shell docker ps -q --filter "name=lumigator-")
 
+-include .env
+# GPU related settings
+#
+RAY_WORKER_GPUS ?= "0.0"
+RAY_WORKER_GPUS_FRACTION ?= "0.0"
+INFERENCE_PIP_REQS ?= ../jobs/inference/requirements_cpu.txt
+
```
agpituk, line 14: I think including the .env here is perfect , but, shouldn't we put this vars that you just added in the .env as well? you can put defaults in the .env.example file

ividal: How would this affect cloud/hybrid settings? On those, the user should be able to use a GPU.

Makefile
```diff
@@ -5,6 +5,13 @@ UNAME:= $(shell uname -o)
 PROJECT_ROOT := $(shell git rev-parse --show-toplevel)
 CONTAINERS_RUNNING := $(shell docker ps -q --filter "name=lumigator-")
 
+-include .env
+# GPU related settings
+#
+RAY_WORKER_GPUS ?= "0.0"
```
ividal, line 11: Suggestion to avoid issues with users that do have a GPU: check if "nvidia-smi" is present, and if it isn't force cpu mode.

This would not work with AMD GPU users, but my assumption is the README request to set the vars via .env file, coupled with your "?=" takes care of them.

lumigator/python/mzai/jobs/inference/requirements_cpu.txt
```diff
@@ -0,0 +1,11 @@
+--extra-index-url https://download.pytorch.org/whl/cpu
```
ividal, line 1: Are we sure this is only used in tests? What gets installed for regular usage and what happens to people who do not have a GPU in that case?

ividal (commented):
Thanks! FYI, explicitly setting RAY_WORKER_GPUS and RAY_WORKER_GPUS_FRACTION solved the integration test problems on #519 .

My only concern is how test/regular use differ from each other and how that affects GPU users.

Makefile
javiermtorres: This relates to https://github.com/mozilla-ai/lumigator/pull/453#discussion_r1903963041 from @aittalam. If we keep the PR as is, then I'll include the `.env` entries. We may use some other, more flexible, per-job way of specifying non-gpu requirements at a later stage.

@ividal this is resolved at lumigator startup time currently. Whoever creates the cloud service would be responsible to set this value appropriately (e.g. the helm charts provided with lumigator). AFAICT these Makefiles would not be used in those cases.

Makefile
javiermtorres: > Suggestion to avoid issues with users that do have a GPU: check if "nvidia-smi" is present, and if it isn't force cpu mode.

My impression is that we have a few settings (like the Ray GPU env vars and the `requirements.txt` repo url prefix lines) that we need to coordinate. I don't really like auto-detection since imho the tools themselves (vendor tools like this, python settings, ray itself...) don't seem so well coordinated, but issuing a warning or some logs seems ok to me ("GPU selected, but no nvidia-smi present" or the like).

lumigator/python/mzai/jobs/inference/requirements_cpu.txt
javiermtorres: This is somehow addressed in https://github.com/mozilla-ai/lumigator/blob/c0350f840b81fa20d023a538fd960108a225f296/lumigator/python/mzai/backend/backend/tests/README.md?plain=1#L38, although a bit more verbose description would probably be desirable.

Makefile
ividal: So...

* is the user responsible for setting GPUs? [Y/n]
* is the default behaviour is CPU? [Y/n]
* how exactly does a user force GPU usage? [On the .env via RAY_WORKER_GPU*/...]
we just need to be explicit in our guide as to how the user can run lumigator with GPU support.
* TODO: 
    * here: add log "setting GPU/CPU" so the user knows what's going on
    * separate issue: document GPU/CPU usage in our user docs.

Does this sound about right?

javiermtorres:
@veekaybee @aittalam @dpoulopoulos @ividal This PR is possibly outdated, with the latest changes in env vars handling :-/ Should I just close it?

dpoulopoulos:
> @veekaybee @aittalam @dpoulopoulos @ividal This PR is possibly outdated, with the latest changes in env vars handling :-/ Should I just close it?

Yes, I would say just close it. We can always re-open it, if we need something.

PR #548, opened 2024-12-26
Fix frontend build

agpituk:
# What's changing

We found out currently our frontend bulid wasn't working properly. The root cause is that the .env file wasn't being passed properly to the npm command, which meant the VUE_APP_BASE_URL wasn't used inside the Dockerfile.

In order to fixed that I've modified several files to ensure consistency accross the project. One of the key things is that the .env.example file has changed to avoid inline comments , which means everyone will need to ensure their .env file is aligned with that once this is merged.

# How to test it

Steps to test the changes:

1. Spin up the Frontend locally (Run make local-up or start-lumigator-build)
2. Check that the frontend can connect to the backend)

# I already...

- [N/A ] Updated the documentation (both comments in code and [product documentation](https://mozilla-ai.github.io/lumigator) under `/docs`)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,4 @@
 # .env.example
-
 # Lumigator API configuration
 # LUMI_API_CORS_ALLOWED_ORIGINS:
 # Comma separated list of origins (See: https://developer.mozilla.org/en-US/docs/Glossary/Origin)
@@ -9,7 +8,6 @@
 # To allow CORS requests from anywhere specify "*" as any, or the only value.
 # e.g. "*"
 LUMI_API_CORS_ALLOWED_ORIGINS=${LUMI_API_CORS_ALLOWED_ORIGINS:-http://localhost,http://localhost:3000}
-
 # AWS Variables for S3 Object Storage
 # Configure these for AWS access, or use defaults for local development with LocalStack.
 AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID:-test}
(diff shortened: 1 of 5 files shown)


ividal:
If it's any help, I've been testing this without the .env under frontend and launching an experiment, seeing results and downloading all works as expected (`local-up` and `start-lumigator`).

I was wondering why the regex was necessary - but will wait until you're done with this, @agpituk

macaab26 (approved):

agpituk:
> I was wondering why the regex was necessary - but will wait until you're done with this, @agpituk
Thanks for that @ividal . I definitely think we should improve this in a follow up, probably using [dotenv](https://www.npmjs.com/package/dotenv) or something like that. The regex is definitely something I don't like (if you mess with the `.env` it won't pull the envs properly for the frontend). Should we merge it as it is and have a follow up after MVP?

agpituk:
> > I was wondering why the regex was necessary - but will wait until you're done with this, @agpituk
> 
> Thanks for that @ividal . I definitely think we should improve this in a follow up, probably using [dotenv](https://www.npmjs.com/package/dotenv) or something like that. The regex is definitely something I don't like (if you mess with the `.env` it won't pull the envs properly for the frontend). Should we merge it as it is and have a follow up after MVP?


I actually wasn't happy with this after reading it and I decided to do another approach. After reading [vite doc ](https://v2.vitejs.dev/config/#async-config)I just removed the regex and changed vite configuration to point to our env file. Please have another look into it @ividal @george-mzai @macaab26

macaab26:
@agpituk Reviewed, looks and works great 😄

macaab26 (approved):

george-mzai (approved):

PR #576, opened 2025-01-07
List jobs per job type

### LINKED ISSUE #380: Return jobs of specific type
**Is your feature request related to a problem? Please describe.**
A user is likely to want to see what evaluation jobs they've run. They don't need to see any stand-alone inference jobs, etc.

**Describe the solution you'd like**
list_jobs could filter by job type (just like create_jobs) does.

javiermtorres:
# What's changing

A GET request can be issued to the `/jobs/{JOB_TYPE}/` endpoint, where currently `JOB_TYPE` is either `inference` or `evaluate`, to select all jobs of the desired type. A GET request can still be issued to `/jobs/{UUID}/` where `UUID` is a v4 UUID to retrieve the details of a single job.

Closes #380 

# How to test it

As tested in the integration tests, two jobs, one inference and one evaluation, can be started, and only one would be selected with the new API.

N/A

# I already...

### DIFF
--- a/lumigator/backend/backend/alembic/versions/3be4994d87d5_add_job_type_to_job_table.py
+++ b/lumigator/backend/backend/alembic/versions/3be4994d87d5_add_job_type_to_job_table.py
@@ -1,8 +1,8 @@
 """Add job type to job table
 
-Revision ID: ef5ee5662ce3
-Revises: e9679cbc3c36
-Create Date: 2025-01-07 13:24:40.657958
+Revision ID: 3be4994d87d5
+Revises: 4c0345b3d525
+Create Date: 2025-02-03 17:38:54.564918
 
 """
 
@@ -12,8 +12,8 @@ import sqlalchemy as sa
 from alembic import op
 
 # revision identifiers, used by Alembic.
-revision: str = "ef5ee5662ce3"  # pragma: allowlist secret
-down_revision: str | None = "e9679cbc3c36"  # pragma: allowlist secret
+revision: str = "3be4994d87d5"  # pragma: allowlist secret
+down_revision: str | None = "4c0345b3d525"  # pragma: allowlist secret
 branch_labels: str | Sequence[str] | None = None
 depends_on: str | Sequence[str] | None = None
 
--- a/lumigator/backend/backend/alembic/versions/ef5ee5662ce3_add_job_type_to_job_table.py
+++ b/lumigator/backend/backend/alembic/versions/ef5ee5662ce3_add_job_type_to_job_table.py
@@ -1,0 +1,25 @@
+"""Add job type to job table
+
+Revision ID: ef5ee5662ce3
+Revises: e9679cbc3c36
+Create Date: 2025-01-07 13:24:40.657958
+
+"""
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = 'ef5ee5662ce3' # pragma: allowlist secret
+down_revision: str | None = 'e9679cbc3c36' # pragma: allowlist secret
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.add_column('jobs', sa.Column('type', sa.String(), nullable=True))
+
+
+def downgrade() -> None:
+    op.remove_column('jobs', sa.Column('type', sa.String(), nullable=True))
--- a/lumigator/backend/backend/alembic/versions/ef5ee5662ce3_add_job_type_to_job_table.py
+++ b/lumigator/backend/backend/alembic/versions/ef5ee5662ce3_add_job_type_to_job_table.py
@@ -5,21 +5,22 @@ Revises: e9679cbc3c36
 Create Date: 2025-01-07 13:24:40.657958
 
 """
+
 from collections.abc import Sequence
 
 import sqlalchemy as sa
 from alembic import op
 
 # revision identifiers, used by Alembic.
-revision: str = 'ef5ee5662ce3' # pragma: allowlist secret
-down_revision: str | None = 'e9679cbc3c36' # pragma: allowlist secret
+revision: str = "ef5ee5662ce3"  # pragma: allowlist secret
+down_revision: str | None = "e9679cbc3c36"  # pragma: allowlist secret
 branch_labels: str | Sequence[str] | None = None
 depends_on: str | Sequence[str] | None = None
 
 
 def upgrade() -> None:
-    op.add_column('jobs', sa.Column('type', sa.String(), nullable=True))
+    op.add_column("jobs", sa.Column("job_type", sa.String(), nullable=True))
 
 
 def downgrade() -> None:
-    op.remove_column('jobs', sa.Column('type', sa.String(), nullable=True))
+    op.remove_column("jobs", sa.Column("job_type", sa.String(), nullable=True))
--- a/lumigator/backend/backend/api/routes/jobs.py
+++ b/lumigator/backend/backend/api/routes/jobs.py
@@ -2,10 +2,11 @@ import json
 from http import HTTPStatus
 from urllib.parse import urljoin
 from uuid import UUID
+from typing import Annotated
 
 import loguru
 import requests
-from fastapi import APIRouter, BackgroundTasks, HTTPException, status
+from fastapi import APIRouter, BackgroundTasks, HTTPException, status, Query
 from lumigator_schemas.datasets import DatasetResponse
 from lumigator_schemas.extras import ListingResponse
 from lumigator_schemas.jobs import (
@@ -134,6 +135,7 @@ def list_jobs(
     service: JobServiceDep,
     skip: int = 0,
     limit: int = 100,
+    job_types: Annotated[list[str] | None, Query()] = None
 ) -> ListingResponse[Job]:
     """Retrieves job data from the Lumigator repository where Ray
     metadata is also available.
@@ -142,6 +144,13 @@ def list_jobs(
 
     NOTE: Lumigator repository data takes precedence over Ray metadata.
     """
+
+    # if job_type not in ["inference", "annotate", "evaluate"]:
+    #     raise ValueError(f"Unknown job type {job_type}") from None
+    # jobs = service.list_jobs_per_type(job_type, skip, limit)
+
+
+
     jobs = service.list_jobs(skip, limit)
     if not jobs or jobs.total == 0 or len(jobs.items) == 0:
         return jobs
--- a/lumigator/backend/backend/api/routes/jobs.py
+++ b/lumigator/backend/backend/api/routes/jobs.py
@@ -1,12 +1,12 @@
 import json
 from http import HTTPStatus
+from typing import Annotated
 from urllib.parse import urljoin
 from uuid import UUID
-from typing import Annotated
 
 import loguru
 import requests
-from fastapi import APIRouter, BackgroundTasks, HTTPException, status, Query
+from fastapi import APIRouter, BackgroundTasks, HTTPException, Query, status
 from lumigator_schemas.datasets import DatasetResponse
 from lumigator_schemas.extras import ListingResponse
 from lumigator_schemas.jobs import (
@@ -135,7 +135,7 @@ def list_jobs(
     service: JobServiceDep,
     skip: int = 0,
     limit: int = 100,
-    job_types: Annotated[list[str] | None, Query()] = None
+    job_types: Annotated[list[str], Query()] = (),
 ) -> ListingResponse[Job]:
     """Retrieves job data from the Lumigator repository where Ray
     metadata is also available.
@@ -144,14 +144,13 @@ def list_jobs(
 
     NOTE: Lumigator repository data takes precedence over Ray metadata.
     """
-
-    # if job_type not in ["inference", "annotate", "evaluate"]:
+    # if job_type not in ["inference", "annotate", "evaluate"]:
     #     raise ValueError(f"Unknown job type {job_type}") from None
     # jobs = service.list_jobs_per_type(job_type, skip, limit)
 
+    loguru.logger.info(f"Listing jobs, job_types={job_types}")
 
-
-    jobs = service.list_jobs(skip, limit)
+    jobs = service.list_jobs(skip, limit, job_types)
     if not jobs or jobs.total == 0 or len(jobs.items) == 0:
         return jobs
 
@@ -205,48 +204,6 @@ def get_job(
     lm_info = job.model_dump()
     merged = {**ray_info, **lm_info}
     return Job(**merged)
-
-
-@router.get("/{job_type}/")
-def get_job_per_type(
-    service: JobServiceDep,
-    job_type: str,
-    skip: int = 0,
-    limit: int = 100,
-) -> ListingResponse[Job]:
-    """Retrieves merged job data from the Lumigator repository and Ray
-    for a valid job type (currently `inference` or `evaluation`).
-
-    Results are a merged representation which form an augmented view of a 'job'.
-
-    NOTE: Lumigator repository data takes precedence over Ray metadata.
-    """
-    if job_type not in ["inference", "annotate", "evaluate"]:
-        raise ValueError(f"Unknown job type {job_type}") from None
-    jobs = service.list_jobs_per_type(job_type, skip, limit)
-
-    # Get all jobs Ray knows about.
-    ray_jobs = _get_all_ray_jobs()
-
-    results = list[Job]()
-
-    # Merge Ray jobs into the repositories jobs
-    job: JobResponse
-    for job in jobs.items:
-        found_job: RayJobDetails
-        found_job = next(
-            (job for job in filter(lambda x: x.submission_id == str(job.id), ray_jobs)), None
-        )
-        if found_job is None:
-            continue
-
-        # Combine both types of response.
-        ray_info = found_job.dict()
-        lm_info = job.model_dump()
-        merged = {**ray_info, **lm_info}
-        results.append(Job(**merged))
-
-    return ListingResponse[Job](total=jobs.total, items=results)
 
 
 @router.get("/{job_id}/logs")
--- a/lumigator/backend/backend/api/routes/jobs.py
+++ b/lumigator/backend/backend/api/routes/jobs.py
@@ -158,15 +158,15 @@ def list_jobs(
     # Merge Ray jobs into the repositories jobs
     for job in jobs.items:
         found_job = next(
-            (x for x in filter(lambda x: x.submission_id == str(job.id), ray_jobs)), None
+            (job for job in filter(lambda x: x.submission_id == str(job.id), ray_jobs)), None
         )
         if found_job is None:
             continue
 
         # Combine both types of response.
-        x = found_job.dict()  # JobSubmissionResponse
-        y = job.model_dump()  # JobResponse
-        merged = {**x, **y}
+        ray_info = found_job.dict()
+        lm_info = job.model_dump()
+        merged = {**ray_info, **lm_info}
         results.append(Job(**merged))
 
     return ListingResponse[Job](
@@ -179,7 +179,7 @@ def list_jobs(
 def get_job(
     service: JobServiceDep,
     job_id: UUID,
-) -> Job | ListingResponse[Job]:
+) -> Job:
     """Attempts to retrieve merged job data from the Lumigator repository and Ray
     for a valid UUID.
 
@@ -191,9 +191,9 @@ def get_job(
     ray_job = _get_ray_job(job_id)
 
     # Combine both types of response.
-    x = ray_job.dict()  # JobSubmissionResponse
-    y = job.model_dump()  # JobResponse
-    merged = {**x, **y}
+    ray_info = ray_job.dict()
+    lm_info = job.model_dump()
+    merged = {**ray_info, **lm_info}
     return Job(**merged)
 
 
@@ -203,7 +203,7 @@ def get_job_per_type(
     job_type: str,
     skip: int = 0,
     limit: int = 100,
-) -> Job | ListingResponse[Job]:
+) -> ListingResponse[Job]:
     """Attempts to retrieve merged job data from the Lumigator repository and Ray
     for a valid job type (currently `inference` or `evaluation`).
 
@@ -223,15 +223,15 @@ def get_job_per_type(
     # Merge Ray jobs into the repositories jobs
     for job in jobs.items:
         found_job = next(
-            (x for x in filter(lambda x: x.submission_id == str(job.id), ray_jobs)), None
+            (job for job in filter(lambda x: x.submission_id == str(job.id), ray_jobs)), None
         )
         if found_job is None:
             continue
 
         # Combine both types of response.
-        x = found_job.dict()  # JobSubmissionResponse
-        y = job.model_dump()  # JobResponse
-        merged = {**x, **y}
+        ray_info = found_job.dict()
+        lm_info = job.model_dump()
+        merged = {**ray_info, **lm_info}
         results.append(Job(**merged))
 
     return ListingResponse[Job](total=jobs.total, items=results)
--- a/lumigator/backend/backend/api/routes/jobs.py
+++ b/lumigator/backend/backend/api/routes/jobs.py
@@ -18,9 +18,8 @@ from lumigator_schemas.jobs import (
     JobResponse,
     JobResultDownloadResponse,
     JobResultResponse,
-    JobSubmissionResponse,
-    JobType,
 )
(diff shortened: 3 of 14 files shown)


lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -16,18 +16,20 @@
...
 from backend.api.deps import JobServiceDep
 from backend.api.http_headers import HttpHeaders
+from backend.api.tags import SERVICES_PATHS
 from backend.settings import settings
 
 router = APIRouter()
 
 
-@router.post("/inference/", status_code=status.HTTP_201_CREATED)
+@router.post(f"/{SERVICES_PATHS[JobType.INFERENCE]}/", status_code=status.HTTP_201_CREATED)
```
veekaybee, line 32: I'd love to better understand what problem we're looking to solve here, are we envisioning that the path will change frequently?

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -103,20 +105,57 @@ def list_jobs(
     )
 
 
-@router.get("/{job_id}")
-def get_job(service: JobServiceDep, job_id: UUID) -> Job:
+@router.get("/{job_spec}")
```
veekaybee, line 108: what will a spec be in this case? we could consider a more descriptive variable name here

veekaybee: i.e. are we changing this from a UUID?

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -103,20 +105,57 @@ def list_jobs(
...
-    y = job.model_dump()  # JobResponse
-    merged = {**x, **y}
-    return Job(**merged)
+    try:
+        job_id = UUID(job_spec)
+        job = service.get_job(job_id)
+        ray_job = _get_ray_job(job_id)
+
+        # Combine both types of response.
+        x = ray_job.model_dump()  # JobSubmissionResponse
```
veekaybee, line 129: if we're looking to understand the type here, we can annotate it: 

```python
x: JobSubmissionResponse = ray_job.model_dump()
```

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -103,20 +105,57 @@ def list_jobs(
...
+        y = job.model_dump()  # JobResponse
```
veekaybee, line 130: same suggestion here, we can annotate the type, and would also suggest renaming x and y to more descriptive variables

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: Consistency. I'd actually like to give a closer look at all the consts we use across several places and enforce the proper derivation, so that if it's changed in one place those changes propagate to all other consts which should be equal but were somehow defined independently :-/

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: We are. It can be a uuid (returning a single job) or a job type name (`inference` or `evaluate`, returning a list of jobs). It follows the approach in the create_job service for consistency. I'd rather use a param (actually inheritance via anyOf plus a discriminator).

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: Hmmm, I'm not sure I wrote that, but type annotations are fine for me. IIRC this would need to be taken from ray (but we do have those dependencies and can import the types, right?).

javiermtorres: > I'm not sure I wrote that

I mean I just copied the existing approach of merging the Ray info together with our own.

lumigator/python/mzai/backend/backend/api/routes/jobs.py
aittalam: +1 to avoid defining things in different, possibly unconsistent, places. I see a couple of issues in doing this here and this way though:

- allowing for a mapping between jobtype and the actual service path adds one level of indirectness that I think we do not need at the moment. It adds no value right now and the price we pay is that someone trying to understand what service is actually hit via HTTP would need to check both `SERVICES_PATHS` and `JobType.INFERENCE`

- more importantly IMO, I feel this is outside of the scope of this PR. As you say, we should give a closer look at all the consts we use across several places and make sure we do things as consistently as possible: if anyone wanted to check when this happened, "list jobs per job type" does not look like the place where this should happen.

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -103,20 +105,57 @@ def list_jobs(
...
 
-@router.get("/{job_id}")
-def get_job(service: JobServiceDep, job_id: UUID) -> Job:
+@router.get("/{job_spec}")
+def get_job(
+    service: JobServiceDep,
+    job_spec: str,
+    skip: int = 0,
+    limit: int = 100,
+) -> Job | ListingResponse[Job]:
```
aittalam, line 114: We already have a method (`list_jobs`) returning a list of jobs. Wouldn't it be easier to create an extra path (e.g. `/filter/{job_spec}` or whatever you want to call it) and map it to a `list_jobs` that takes `job_spec` as an extra optional parameter and behaves like before if that is none?

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -254,6 +278,19 @@ def get_job(self, job_id: UUID) -> JobResponse:
 
         return JobResponse.model_validate(record)
 
+    def list_jobs_per_type(
```
aittalam, line 281: We already have a `list_jobs` [method](https://github.com/mozilla-ai/lumigator/blob/238cbb12bda6151b57db529201d6ab9cd65cef55/lumigator/python/mzai/backend/backend/services/jobs.py). Would it be possible to add an optional type parameter to it and just change the query you do (i.e. call `self._list_job_records_per_type` instead of `self.job_repo.list`) depending on its presence?
The advantage is that this avoid code replication and guarantees the general behavior is consistent (for instance, here you do not validate the returned records, while this happens in the other method)

lumigator/python/mzai/backend/backend/api/routes/jobs.py
veekaybee: I'd prefer to have one route return one type of object for clarity and consistency. If we have a spec that returns multiple objects, we should return a dictionary of elements.

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -103,20 +105,57 @@ def list_jobs(
...
+        job = service.get_job(job_id)
+        ray_job = _get_ray_job(job_id)
+
+        # Combine both types of response.
+        x = ray_job.model_dump()  # JobSubmissionResponse
+        y = job.model_dump()  # JobResponse
+        merged = {**x, **y}
+        return Job(**merged)
+    except ValueError:
+        # If job_spec is not a UUID, it's assumed to be a
```
veekaybee, line 134: missing comment?

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -103,20 +105,57 @@ def list_jobs(
...
+@router.get("/{job_spec}")
+def get_job(
+    service: JobServiceDep,
+    job_spec: str,
+    skip: int = 0,
+    limit: int = 100,
+) -> Job | ListingResponse[Job]:
     """Attempts to retrieve merged job data from the Lumigator repository and Ray.
+    If a valid UUID is provided, a single record is returned. If a valid
+    job type (currently `inference` or `evaluation`) is provided, then a list
```
veekaybee, line 117: I would split this functionality into two methods, one that handles UUIDs and another that handles list of jobs by type

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -103,20 +105,57 @@ def list_jobs(
...
+        jobs = service.list_jobs_per_type(job_spec, skip, limit)
+
+        # Get all jobs Ray knows about.
+        ray_jobs = _get_all_ray_jobs()
+
+        results = list[Job]()
+
+        # Merge Ray jobs into the repositories jobs
+        for job in jobs.items:
+            found_job = next(
```
veekaybee, line 146: Not sure I understand this piece of logic: why are we looking to combine Ray job and lumigator IDs here?

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: Yep, no problem. I'll set it up using the hardcoded names again, and later on we can have a specific issue to see how the identifiers for a job are generated and used across lumigator itself.

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: We have a couple of constraints: make it backwards compatible, and follow a similar interface to the existing calls.
Regarding the dictionary, the `GET /jobs` call already returns a list, so I'd see it natural to return a list as well.
Regarding the routes, I can use a more specific route like `GET /jobs/per_type/{job_type}` to avoid overloading the `/jobs/{job_id}` route. Wdyt?

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: I'm not sure I follow, but we need an extra select in the query since `list_jobs` retrieves everything. We could refactor the existing queries, but each one is 3 lines and I don't see much advantage in reusing them. I can research if setting `None` to things like selects avoids using them in the query, but we may end up with a single over-generic query with 8 out of 10 params set to `None`, which is not desirable either :-/

BTW I have the feeling that if the limit is used, and the amount of objects is larger than the limit, the wrong total is returned. But I need to make a separate test and PR to check it out.

lumigator/python/mzai/backend/backend/services/jobs.py
javiermtorres: As I mentioned, I don't think generalizing the queries is a good idea. I can check if we can do some generalization at service level (which I have already tried to do), but we can introduce the per-type selection first and then refactor it in a later PR.

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: Ouch, yep, it should go there :-/ But it'll go away if we make a separate endpoint, anyway.

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: Basically https://github.com/mozilla-ai/lumigator/pull/576#discussion_r1908935742, right?

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: This was done in a separate PR already, I'm just following it. The Ray info is bundled with the Lumigator job info so the UI has all the info in one place. Please check https://github.com/mozilla-ai/lumigator/pull/514

lumigator/python/mzai/backend/backend/api/routes/jobs.py
veekaybee: Yep

lumigator/python/mzai/backend/backend/alembic/versions/ef5ee5662ce3_add_job_type_to_job_table.py
```diff
@@ -0,0 +1,26 @@
...
+down_revision: str | None = "e9679cbc3c36"  # pragma: allowlist secret
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.add_column("jobs", sa.Column("job_type", sa.String(), nullable=True))
+
+
+def downgrade() -> None:
```
veekaybee, line 25: are we adding and removing the same column here?

javiermtorres: If I understand correctly (which may not be doing!) adding the column is part of the upgrade procedure _into_ this db revision from the previous, and removing the column is part of the downgrade procedure _from_ this db revision into the previous. But maybe I'm not getting it right :-/

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: We can actually use `/jobs/{uuid}` and `/jobs/inference/` / `/jobs/evaluate/` with trailing slashes :-/ It's a bit messy too but I kinda like it better.

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: As discussed, the queries will be kept separate, since queries are small and new ones can be composed easily.

The services will also be kept separate for the moment, as the benefit of overloading a single service with several queries is not yet clear, but we may rethink this over at a later stage. The params for the queries will be set as consts to ensure consistency across both types of job listings.

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: I agreed with @aittalam on using `/jobs/{job_type}/` to keep it consistent with the existing POST methods. Let me know if this would be ok for you.

lumigator/python/mzai/backend/backend/alembic/versions/ef5ee5662ce3_add_job_type_to_job_table.py
```diff
@@ -0,0 +1,26 @@
...
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "ef5ee5662ce3"  # pragma: allowlist secret
+down_revision: str | None = "e9679cbc3c36"  # pragma: allowlist secret
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
```
veekaybee, line 18: Let's make sure when we merge this it doesn't conflict with other alembic changes

lumigator/python/mzai/backend/backend/alembic/versions/ef5ee5662ce3_add_job_type_to_job_table.py
aittalam: I confirm! This is automatic code generated by alembic to upgrade / downgrade the database to a new / older version.
As Dimitri's PR has been merged, I think you don't need to have this anymore as you are not actively modifying the DB (IIRC)

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -111,15 +111,15 @@ def list_jobs(
...
         )
         if found_job is None:
             continue
 
         # Combine both types of response.
-        x = found_job.model_dump()  # JobSubmissionResponse
-        y = job.model_dump()  # JobResponse
-        merged = {**x, **y}
+        ray_info = found_job.dict()
+        lm_info = job.model_dump()
```
veekaybee, line 121: more descriptive variable name would be helpful here, `ray_job_info` perhaps?

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -129,21 +129,67 @@ def list_jobs(
...
-def get_job(service: JobServiceDep, job_id: UUID) -> Job:
-    """Attempts to retrieve merged job data from the Lumigator repository and Ray.
+def get_job(
+    service: JobServiceDep,
+    job_id: UUID,
+) -> Job:
+    """Attempts to retrieve merged job data from the Lumigator repository and Ray
+    for a valid UUID.
+
+    The result is a merged representation which forms an augmented view of a 'job'.
```
veekaybee, line 139: looks like the comment string is incorrectly formatted

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -129,21 +129,67 @@ def list_jobs(
...
 
 
+@router.get("/{job_type}/")
+def get_job_per_type(
+    service: JobServiceDep,
+    job_type: str,
+    skip: int = 0,
+    limit: int = 100,
+) -> ListingResponse[Job]:
+    """Attempts to retrieve merged job data from the Lumigator repository and Ray
```
veekaybee, line 160: Would change this to Retrieves instead of Attempts to retrieve

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -194,7 +240,7 @@ def get_job_result_download(
     return service.get_job_result_download(job_id)
 
 
-def _get_all_ray_jobs() -> list[JobSubmissionResponse]:
+def _get_all_ray_jobs() -> list[RayJobDetails]:
```
veekaybee, line 243: why are we changing this?

lumigator/python/mzai/backend/backend/repositories/jobs.py
```diff
@@ -10,14 +10,19 @@ class JobRepository(BaseRepository[JobRecord]):
...
     def __init__(self, session: Session):
         super().__init__(JobResultRecord, session)
 
     def get_by_job_id(self, job_id: UUID) -> JobResultRecord | None:
-        return (
-            self.session.query(JobResultRecord)
-            .where(JobResultRecord.job_id == job_id)
-            .first()
-        )
+        return self.session.query(JobResultRecord).where(JobResultRecord.job_id == job_id).first()
```
veekaybee, line 28: do we need to change this?

lumigator/python/mzai/backend/backend/alembic/versions/ef5ee5662ce3_add_job_type_to_job_table.py
javiermtorres: Hmmmmm, true :-/ I'll regenerate the alembic part just before merging just in case.

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -111,15 +111,15 @@ def list_jobs(
     # Merge Ray jobs into the repositories jobs
     for job in jobs.items:
         found_job = next(
```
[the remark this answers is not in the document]
javiermtorres: Handling a set/dict to keep track of the ids definitely helps finding out ray/job relations.
Ok, refactoring 👍

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: Ok

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -129,21 +129,67 @@ def list_jobs(
...
     NOTE: Lumigator repository data takes precedence over Ray metadata.
     """
     job = service.get_job(job_id)
     ray_job = _get_ray_job(job_id)
 
     # Combine both types of response.
-    x = ray_job.model_dump()  # JobSubmissionResponse
-    y = job.model_dump()  # JobResponse
-    merged = {**x, **y}
+    ray_info = ray_job.dict()
```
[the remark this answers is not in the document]
javiermtorres: We do. It's just an older version that Ray uses, afaict :-/ I tried using model_dump, but if I got it right it's their Pydantic version not implementing it.
https://github.com/mozilla-ai/lumigator/blob/1b5151e48501e66f5922193d71f48a7eb659ea48/lumigator/python/mzai/backend/backend/api/routes/jobs.py#L243
https://github.com/mozilla-ai/lumigator/blob/1b5151e48501e66f5922193d71f48a7eb659ea48/lumigator/python/mzai/backend/backend/api/routes/jobs.py#L20
https://github.com/ray-project/ray/blob/916f534e571278b26733812b24a7b3dee08f24e4/python/ray/dashboard/modules/job/pydantic_models.py#L38

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: Because the info we get is modelled by the Ray pydantic models. _Then_ it is merged with our own job info.

lumigator/python/mzai/backend/backend/repositories/jobs.py
javiermtorres: Hum, I'll check it out. It may have been the formatter :-S

lumigator/python/mzai/backend/backend/api/routes/jobs.py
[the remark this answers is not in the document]
javiermtorres: Did something similar:

```
    ray_jobs = {ray_job.submission_id: ray_job for ray_job in _get_all_ray_jobs()}

    results = list[Job]()

    # Merge Ray jobs into the repositories jobs
    job: JobResponse
    for job in jobs.items:
        found_job: RayJobDetails
        found_job = ray_jobs[str(job.id)]
        if found_job is None:
            continue
```

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: In what sense? I don't get the issue :(

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: The merged information is not modelled as `JointJobInfo(RayJobDetails, JobResponse)` because this would require the lumigator schemas to import the `ray` package, which is too heavy for the end user as there is no separate api or model package for Ray.

lumigator/python/mzai/backend/backend/repositories/jobs.py
javiermtorres: Dunno.... :-/ Reverted to the original state anyway.

lumigator/python/mzai/backend/backend/alembic/versions/ef5ee5662ce3_add_job_type_to_job_table.py
aittalam: I stand corrected, the previous PR was not dealing with the`job_type` field. You'll need to run alembic again then and generate a more recent migration if you have not done it already

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -129,21 +129,67 @@ def list_jobs(
...
+    lm_info = job.model_dump()
+    merged = {**ray_info, **lm_info}
     return Job(**merged)
 
 
+@router.get("/{job_type}/")
```
njbrake, line 153: This seems slightly strange since already have a get route for /{job_id}. Is there a better way to make it clear that one route is getting a job, and another list listing jobs per type? It seems like this works, but may confuse people who are exploring our swagger docs/etc, not sure.

njbrake (approved):
I don't see any major issues, just one question about the choice of route name for retrieving jobs by type. Thanks for the work!

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: Please take a read at https://github.com/mozilla-ai/lumigator/pull/576#discussion_r1907515466 for reference. TL;DR any choice seems bad 😅 This was chosen to keep in line with the current POSTs and avoid overloading the root jobs route. But maybe some additional comments in the openapi spec could do the trick; I'll inspect the generated spec and see how it looks.

khaledosman: IMO I don't think a separate endpoint is needed, how about a query param for filtering?

i.e GET /jobs <-- returns all jobs
GET /jobs?job_type=evaluate <-- returns only evaluate jobs

javiermtorres: LGTM!!!! @aittalam @veekaybee what do you think?
`GET /jobs?job_types=evaluate` would seem ok to me (and iirc this _may_ allow unifying some calls in the underlying service/records as a bonus)

njbrake: I love @khaledosman's suggestion! My vote is for that 💯

PR #578, opened 2025-01-08
SDK: Adjust behaviour for vendor caching

peteski22:
# What's changing

At present the SDK client attempts to reach out to the backend API when it is instantiated. It does this in order to get the list of supported vendors and caches them.

If completions are requested via the SDK then the cached vendors are used, but it means that we attempt to do this regardless of what the developer using the SDK intended. 

This PR adjusts the behaviour of the SDK so that we only attempt to get and cache vendors when developers explicitly ask for them, or if they request a completion and we can't initially find the vendor in the cache.

A side effect of this could be that we don't prematurely attempt to hit the API in things like SDK integration tests.

# How to test it

Run unit and integration tests.

Also ran `uvx ruff check --fix`

# I already...

### DIFF
--- a/lumigator/python/mzai/sdk/lumigator_sdk/client.py
+++ b/lumigator/python/mzai/sdk/lumigator_sdk/client.py
@@ -4,7 +4,6 @@ from typing import Any, Dict  # noqa: UP035
 
 import requests
 from loguru import logger
-from requests.exceptions import HTTPError
 
 
 def _make_request(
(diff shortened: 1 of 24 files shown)


lumigator/python/mzai/sdk/lumigator_sdk/client.py
peteski22, line 1: Changes from `uvx ruff check --fix`

lumigator/python/mzai/sdk/tests/integration/test_scenarios.py
peteski22, line 1: Changes from `uvx ruff check --fix`

javiermtorres (commented):
LGTM

javiermtorres (approved):

macaab26 (approved):

PR #596, opened 2025-01-10
Issue 556 retrieve models

### LINKED ISSUE #556: Retrieve models from the API
### Motivation

We currently have the models that you can use for summarization hardcoded in the Frontend. After the work done in #381 we can query the API to get the URL of these models instead of having them hardcoded.

### Alternatives

_No response_

### Contribution

_No response_

george-mzai:
# What's changing
We use the `/models/{task_name}` endpoint to populate the list of models inside the experiment form

Refs #...
Closes #556 

# How to test it
Try to run a new experiment 

⚠️ Some of the new model options require an API key ⚠️ 

If the Lumigator operator hasn't configured the specific API key during setup, Lumigator won't be able to contact the API for the model successfully.

Currently there is no warning about this as the UI has no way to know:

1. If the API requires an API key
1. If Lumigator has the key configured or not

As a result, the experiment/job is registered in the backend but soon after receives a "Failed" status. The user will likely only notice the misconfiguration by examining the logs. 

failed

> The link to the documentation website of each model is missing because it's not provided from the API. (previously it was hardcoded)

### DIFF
--- a/lumigator/frontend/src/components/molecules/LExperimentForm.vue
+++ b/lumigator/frontend/src/components/molecules/LExperimentForm.vue
@@ -90,7 +90,6 @@
     </div>
     <div class="l-experiment-form__models-container">
       <h3>Model Selection</h3>
-      <h4>Hugging Face Model Hub & Mistral</h4>
       <div class="l-experiment-form__models">
         <l-model-cards  ref="modelSelection"/>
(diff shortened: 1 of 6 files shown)


peteski22 (approved):
Thanks @george-mzai 😄

Issue #597, opened 2025-01-10
[FEATURE]: Improve API docs for dataset upload


peteski22:
### Motivation

It's unclear what kind of status codes should be returned from the API if there are problems with the dataset that is being uploaded. 

We should make it clear by highlighting the status codes that are returned for a file that's too big, and a file that isn't a CSV.

### Have you searched for similar issues before submitting this one?

- [x] Yes, I have searched for similar issues

PR #598, opened 2025-01-10
API: Documentation improvement for dataset upload

peteski22:
# What's changing

Improve the Swagger documentation for our API to show the status codes for errors that could occur when trying to upload a dataset.

Closes: https://github.com/mozilla-ai/lumigator/issues/597

# How to test it

Steps to test the changes:

1. `gh pr checkout 598`
1. `make local-up`
1. Visit [API docs for dataset upload](http://localhost:8000/docs#/datasets/upload_dataset_api_v1_datasets__post)
1. Check the responses section

Should now look like this:

image

# I already...


lumigator/python/mzai/backend/backend/settings.py
```diff
@@ -10,7 +11,8 @@
 class BackendSettings(BaseSettings):
     # Backend
     DEPLOYMENT_TYPE: DeploymentType = DeploymentType.LOCAL
-    MAX_DATASET_SIZE: ByteSize = "50MB"
+    MAX_DATASET_SIZE_HUMAN_READABLE: Final[str] = "50MB"
```
dpoulopoulos, line 14: Is there a reason for this 50MB limit? If yes, can we add a comment here explaining it?

dpoulopoulos (commented):
Thanks for this @peteski22. I added a minor comment.

lumigator/python/mzai/backend/backend/settings.py
peteski22: I'm not sure, from the git history, it looks like it has always been set at ~50MB.

I just wanted to pull it out into a string so we could include it in the docs and prevent user frustration.

dpoulopoulos: @peteski22 I edited my comment, adding more context. Sorry for the latency I caused!

dpoulopoulos:
It would be good to have a few unit tests for this. :)

peteski22:
> It would be good to have a few unit tests for this. :)

@dpoulopoulos do you mean unit tests that cover the max upload size via the dataset service? Please feel free to create an issue and we can look to address that. 😄 

If you mean unit tests for the changes in this PR specifically, I don't think it's worthwhile trying to test the output of the Swagger docs, or test a one line constant being split into two lines.

dpoulopoulos:
> > It would be good to have a few unit tests for this. :)
> 
> @dpoulopoulos do you mean unit tests that cover the max upload size via the dataset service? Please feel free to create an issue and we can look to address that. 😄
> 
> If you mean unit tests for the changes in this PR specifically, I don't think it's worthwhile trying to test the output of the Swagger docs, or test a one line constant being split into two lines.

I meant adding unit tests that test whether the backend returns the right HTTP status code for each use case. But we can address that later.

dpoulopoulos (approved):

Issue #593, opened 2025-01-10
[BUG] Inference jobs should support API-based models


ividal:
### Description

Currently we cannot request inference with models behind an external API, like Mistral or OpenAI.

Neither the [inference+evaluator combo](https://github.com/mozilla-ai/lumigator/blob/fde73e2084d5f00bca560eea675c55ee00e7c610/lumigator/python/mzai/jobs/evaluator/evaluator/jobs/model_clients.py#L103), nor the [stand-alone inference job](https://github.com/mozilla-ai/lumigator/blob/main/lumigator/python/mzai/jobs/inference/model_clients.py#L84) seem to support it.

NB: `/experiments` delegates to `/jobs/evaluate` (which is using the current inference+evaluator combo).

### Reproduction

1. Launch Lumigator and upload a test csv - note the dataset_id
2. On the Lumigator swagger (e.g. localhost:8000/docs ), trigger an experiment (POST /experiments ) with a payload including that dataset_id and an openai or mistral model.
3. See the connection error below.

[job-driver-29608e6e-9f7f-4694-a277-2c54cef3af60.log](https://github.com/user-attachments/files/18374216/job-driver-29608e6e-9f7f-4694-a277-2c54cef3af60.log)

Repeat this with /jobs/inference or /jobs/evaluate (always using that dataset_id and a mistral/openai model in the payload).

Additional logs for these to be added.

### Relevant log output

```shell
openai.APIConnectionError: Connection error.

See more in the log above.
```

### Expected behavior

Provided one has set the correct API key, the job should go through.
Without an API key, the connection should still go through and a different (authentication) error would appear.

### System Info

* Local mode, macOS
* Current main ([fde73e2](https://github.com/mozilla-ai/lumigator/commit/fde73e2084d5f00bca560eea675c55ee00e7c610)).

### Have you searched for similar issues before submitting this one?

- [x] Yes, I have searched for similar issues

aittalam:
Fixed with the merging of https://github.com/mozilla-ai/lumigator/pull/599

PR #599, opened 2025-01-10
Harden OpenAI and Mistral env var injection

veekaybee:
# What's changing

We had an issue where we weren't able to hit `completions` endpoints for inference. (https://github.com/mozilla-ai/lumigator/issues/593) 

There were two core issues: 

1. Correctly propogating local environment variables to the application
2. Reading dynamic config properties from backend settings

Injecting the environment variables directly via the docker-config file fixes this and propogates Mistral and OpenAI correctly: 

In the process of fixing, we're cleaning up several completions issues: 

* Making sure env vars are computed fields
* Fixed the mistral backend service dependency
* Changed typing on model listing from OpenAI
* Getting summarizer prompt from config now

# How to test it

Steps to test the changes:

1. `make local-up` from this branch

2.  run: 
```bash
curl -X 'POST' \                                                           
  'http://localhost:8000/api/v1/completions/openai' \
  -d '{
  "text": "string"
}'

{"text":"It seems you've entered the word \"string.\" If you have a specific topic or text related to \"string\" that you would like summarized, please provide more details."}%

curl -X 'POST' \                                                            
  'http://localhost:8000/api/v1/completions/mistral' \
  -d '{
  "text": "string"
}
```

# I already...

- [X ] Tested the changes in a working environment to ensure they work as expected

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -113,8 +113,8 @@ services:
       - INFERENCE_WORK_DIR=/mzai/lumigator/python/mzai/jobs/inference
       - RAY_DASHBOARD_PORT
       - RAY_HEAD_NODE_HOST
-      - MISTRAL_API_KEY
-      - OPENAI_API_KEY
+      - MISTRAL_API_KEY=${MISTRAL_API_KEY}
+      - OPENAI_API_KEY=${OPENAI_API_KEY}
       - RAY_WORKER_GPUS=0
       - RAY_WORKER_GPUS_FRACTION=0
       - LUMI_API_CORS_ALLOWED_ORIGINS
--- a/lumigator/python/mzai/backend/backend/services/completions.py
+++ b/lumigator/python/mzai/backend/backend/services/completions.py
@@ -16,6 +16,8 @@ class CompletionService(ABC):
 
 
 class MistralCompletionService(CompletionService):
+    pass
+
     def __init__(self):
         self.client = MistralClient(api_key=settings.MISTRAL_API_KEY)
         self.model = "open-mistral-7b"
@@ -51,8 +53,11 @@ class OpenAICompletionService(CompletionService):
         self.max_tokens = 256
         self.temperature = 1
         self.top_p = 1
+        self.prompt = """You are a helpful assistant, expert in text summarization.
+                For every prompt you receive,
+                provide a summary of its contents in at most two sentences."""
 
-    def get_models(self) -> mistralai.client.ModelList:
+    def get_models(self):
         response = self.client.list_models()
 
         return response
--- a/lumigator/python/mzai/backend/backend/services/completions.py
+++ b/lumigator/python/mzai/backend/backend/services/completions.py
@@ -1,6 +1,7 @@
 from abc import ABC, abstractmethod
 
 import mistralai.client
+from loguru import logger
 from lumigator_schemas.completions import CompletionRequest, CompletionResponse
 from mistralai.client import MistralClient
 from mistralai.models.chat_completion import ChatMessage
@@ -63,6 +64,7 @@ class OpenAICompletionService(CompletionService):
         return response
 
     def get_completions_response(self, request: CompletionRequest) -> CompletionResponse:
+        logger.info(f"oai key...{settings.OAI_API_KEY}")
         response = self.client.chat.completions.create(
             model=self.model,
             messages=[
--- a/lumigator/python/mzai/backend/backend/services/completions.py
+++ b/lumigator/python/mzai/backend/backend/services/completions.py
@@ -1,7 +1,6 @@
 from abc import ABC, abstractmethod
 
 import mistralai.client
-from loguru import logger
 from lumigator_schemas.completions import CompletionRequest, CompletionResponse
 from mistralai.client import MistralClient
 from mistralai.models.chat_completion import ChatMessage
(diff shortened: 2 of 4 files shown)


docker-compose.yaml
```diff
@@ -113,8 +113,8 @@ services:
       - INFERENCE_WORK_DIR=/mzai/lumigator/python/mzai/jobs/inference
       - RAY_DASHBOARD_PORT
       - RAY_HEAD_NODE_HOST
-      - MISTRAL_API_KEY
-      - OPENAI_API_KEY
+      - MISTRAL_API_KEY=${MISTRAL_API_KEY}
```
veekaybee, line 116: @agpituk fyi

veekaybee: @peteski22 related to your env injection issues from yesterday, this is helpful for these specific vars

veekaybee: They should, but they don't the way we have it currently set - changing this specifically fixes this error at the env var level

lumigator/python/mzai/backend/backend/settings.py
```diff
@@ -81,9 +81,15 @@ def inherit_ray_env(self, runtime_env_vars: Mapping[str, str]):
             if env_var is not None:
                 runtime_env_vars[env_var_name] = env_var
 
-    OAI_API_KEY: str = os.environ.get("OPENAI_API_KEY", "")
+    @computed_field
+    @property
+    def OAI_API_KEY(self) -> str:  # noqa: N802
+        return os.environ.get("OPENAI_API_KEY", "")
```
dpoulopoulos, line 87: Maybe we can add `info` or `warning` logs here to notify the user about the presence or absence of the environment variable. This would make it easier to track these kinds of issues in the future. Same for the Mistral key. WDYT?

dpoulopoulos (commented):
Thanks for this @veekaybee. I added a minor comment.

docker-compose.yaml
```diff
@@ -113,8 +113,8 @@ services:
       - INFERENCE_WORK_DIR=/mzai/lumigator/python/mzai/jobs/inference
       - RAY_DASHBOARD_PORT
       - RAY_HEAD_NODE_HOST
-      - MISTRAL_API_KEY
-      - OPENAI_API_KEY
```
ividal, line 117: The .env plays a role here. Specially if you generated (as advised) from .env.example.

Let's remove the expression from the .env.example + document that the user needs to add their API key if they want to use openai/mistral.

Context:
:white_check_mark: not setting *API_KEY in .env and just exporting separately in the terminal you launch lumigator from or via one's .zshrc/.bashrc) works
:white_check_mark: setting the explicit key inside .env works --> let's advise this to the user
:x: setting and expression (MISTRAL_API_KEY=${MISTRAL_API_KEY:-} ) in my .env does not seem to work

lumigator/python/mzai/backend/backend/settings.py
veekaybee: added here at the service level https://github.com/mozilla-ai/lumigator/pull/599/files#diff-35d12bc5fa5da1376ba28fa5fe007bfd8849aa4229dd39f85125969c2134d253R50

docker-compose.yaml
veekaybee: addressed with these https://github.com/mozilla-ai/lumigator/pull/599/commits/2fce3b37e8c941bfbfae2418a2297470987fd9bd

docs/source/get-started/installation.md
```diff
@@ -11,7 +11,7 @@ Before you start, make sure you have the following:
 - A working installation of [Docker](https://docs.docker.com/engine/install/)
     - On MAC, Docker Desktop >= 4.3, and `docker-compose` >= 1.28.
     - On Linux, please also complete the [post-installation steps](https://docs.docker.com/engine/install/linux-postinstall/).
-- The system Python; no version manager, such as `pyenv` or `uv`, should be active.
+- The system Python; no version manager, such as `uv`, should be active.
```
ividal, line 14: I'd leave that there - notice it's a "don't have pyenv". Last I checked (around the switch to github), pyenv (both on MacOS and Linux) would still conflict with uv, no?

docker-compose.yaml
ividal: Thanks! Then this closes #579 too.

README.md
```diff
@@ -59,6 +59,11 @@ services networked together to make up all the components of the Lumigator appli
 > [!NOTE]
 > Lumigator requires an SQL database to hold metadata for datasets and jobs. The local deployment
 > uses SQLite for this purpose.
+>
+> [!NOTE]
+> If you'd like to evaluate against LLM APIs like OpenAI and Mistral, you'll need to have your
+> environment variable set locally for Lumigator pick it up at runtime, or, alternately, inject
+> into the running `backend` docker container.
```
dpoulopoulos, line 66: I would like a link to `.env.examples` here, just to see the names of the env vars.

dpoulopoulos (commented):
Looks good to me! I left a minor comment.

lumigator/python/mzai/backend/backend/services/completions.py
```diff
@@ -46,13 +47,19 @@ def get_completions_response(self, request: CompletionRequest) -> CompletionResp
 
 class OpenAICompletionService(CompletionService):
     def __init__(self):
+        if settings.OPENAI_API_KEY is None:
```
javiermtorres, line 50: Shouldn't the key be OAI_API_KEY instead? Mistral works for me, but OpenAI doesn't, according to the test instructions
```suggestion
        if settings.OAI_API_KEY is None:
```

aittalam (approved):
Thanks for this PR Vicki!
I tested it both with your code and with the actual pipeline (running an inference job) and they all worked.
I also verified that the keys were propagated to the containers with the following:

```
docker exec `docker ps | grep "lumigator:latest" | awk '{print $1}'` sh -c "set | grep MISTRAL_API_KEY"

docker exec `docker ps | grep "rayproject/ray" | awk '{print $1}'` sh -c "set | grep MISTRAL_API_KEY"
```
... and they were in both containers. Good to go for me!

Issue #571, opened 2025-01-07
Add get experiment status + tests


aittalam:
The purpose of this issue is adding the equivalent of jobs/{job_id} status for experiments_new. This is done by taking an experiment_id as input and looking in the jobs table for the (two) jobs matching that id. Then, the returned status should be:

PENDING if the inference job status is PENDING
SUCCESSFUL if the evaluation job status is SUCCESSFUL
FAILED if any of the two jobs’ statuses is FAILED
RUNNING otherwise

Depends on https://github.com/mozilla-ai/lumigator/issues/570.

ividal:
I think that what you're doing on the integration tests for #590 is related to this @javiermtorres 
Untagging myself to free this!

PR #604, opened 2025-01-10
Add experiment record handling to experiment service

### LINKED ISSUE #590: Inference jobs should store results as a dataset
### Motivation

Currently inference jobs create results as a json and store them on S3. The limitation of this is they cannot be later used as input for experiments. For example, when a dataset without groundtruth is annotated, this is an inference job. The user then has to download the json file and convert that to a csv on their own.

Instead, it would be nice if the user saw the result appear as a new row in the dataset view, meaning a new dataset has been added to the DB.

### Alternatives

Allowing the upload of a dataset as a json file, which would allow users to download the results as they are provided by `/jobs/{job_id}/results/download`. This is still a huge inconvenience.

### Contribution

_No response_

javiermtorres:
# What's changing

The creation of datasets for a job is indicated through a new parameter in the JobInferCreate, called `store_to_dataset`. The experiment will set this when it creates the appropriate jobs.

The experiment service (used by the `experiments_new` endpoint) will now use the experiment repository and keep track of its contained jobs to provide a status.

Closes #590

# How to test it

An inference job that includes `store_to_dataset = True` in the job config will also add a dataset with the results. The integration tests already check this.

There are additional checks in the experiment launch about the datasets created by the experiment jobs, as well as about the experiment status. The integration test logs the experiment status for manual inspection; only the final `SUCCEEDED` or `FAILED` status are actually checked.

N/A

# I already...

  - No migration needed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,6 +163,8 @@ test-backend-integration:
 	INFERENCE_WORK_DIR=../jobs/inference \
 	EVALUATOR_PIP_REQS=../jobs/evaluator/requirements.txt \
 	EVALUATOR_WORK_DIR=../jobs/evaluator \
+	EVALUATOR_LITE_PIP_REQS=../jobs/evaluator_lite/requirements.txt \
+	EVALUATOR_LITE_WORK_DIR=../jobs/evaluator_lite \
 	uv run pytest -s -o python_files="backend/tests/integration/*/test_*.py"
 
 test-backend-integration-containers:
--- a/lumigator/python/mzai/backend/backend/config_templates.py
+++ b/lumigator/python/mzai/backend/backend/config_templates.py
@@ -99,7 +99,7 @@ default_infer_template = """{{
         "use_fast": "{use_fast}",
         "trust_remote_code": "{trust_remote_code}",
         "torch_dtype": "{torch_dtype}",
-        "max_new_tokens": 500
+        "max_length": 500
     }},
      "job": {{
         "max_samples": {max_samples},
--- a/lumigator/python/mzai/backend/backend/config_templates.py
+++ b/lumigator/python/mzai/backend/backend/config_templates.py
@@ -98,7 +98,8 @@ default_infer_template = """{{
         "revision": "{revision}",
         "use_fast": "{use_fast}",
         "trust_remote_code": "{trust_remote_code}",
-        "torch_dtype": "{torch_dtype}"
+        "torch_dtype": "{torch_dtype}",
+        "max_new_tokens": 500
     }},
      "job": {{
         "max_samples": {max_samples},
--- a/lumigator/python/mzai/backend/backend/services/experiments.py
+++ b/lumigator/python/mzai/backend/backend/services/experiments.py
@@ -1,14 +1,11 @@
 import asyncio
 import csv
-import json
 from collections.abc import Callable
 from io import BytesIO, StringIO
-from pathlib import Path
 from uuid import UUID
 
 import loguru
-from fastapi import BackgroundTasks, UploadFile
-from lumigator_schemas.datasets import DatasetFormat
+from fastapi import BackgroundTasks
 from lumigator_schemas.experiments import ExperimentCreate, ExperimentResponse
 from lumigator_schemas.extras import ListingResponse
 from lumigator_schemas.jobs import (
@@ -16,12 +13,10 @@ from lumigator_schemas.jobs import (
     JobInferenceCreate,
     JobStatus,
 )
-from s3fs import S3FileSystem
 
 from backend.repositories.experiments import ExperimentRepository
 from backend.services.datasets import DatasetService
 from backend.services.jobs import JobService
-from backend.settings import settings
 
 
 class ExperimentService:
@@ -53,44 +48,6 @@ class ExperimentService:
 
         return bin_data
 
-    def _add_dataset_to_db(self, job_id: UUID, request: JobInferenceCreate):
-        loguru.logger.info("Adding a new dataset entry to the database...")
-        s3 = S3FileSystem()
-
-        # Get the dataset from the S3 bucket
-        result_key = self._job_service._get_results_s3_key(job_id)
-        with s3.open(f"{settings.S3_BUCKET}/{result_key}", "r") as f:
-            results = json.loads(f.read())
-
-        # define the fields we want to keep from the results JSON
-        # and build a CSV file from it as a BytesIO object
-        fields = ["examples", "ground_truth", request.output_field]
-        bin_data = self._results_to_binary_file(results, fields)
-        bin_data_size = len(bin_data.getvalue())
-
-        # Figure out the dataset filename
-        dataset_filename = self._dataset_service.get_dataset(dataset_id=request.dataset).filename
-        dataset_filename = Path(dataset_filename).stem
-        dataset_filename = f"{dataset_filename}-annotated.csv"
-
-        upload_file = UploadFile(
-            file=bin_data,
-            size=bin_data_size,
-            filename=dataset_filename,
-            headers={"content-type": "text/csv"},
-        )
-        dataset_record = self._dataset_service.upload_dataset(
-            upload_file,
-            format=DatasetFormat.JOB,
-            run_id=job_id,
-            generated=True,
-            generated_by=results["model"],
-        )
-
-        loguru.logger.info(
-            f"Dataset '{dataset_filename}' with ID '{dataset_record.id}' added to the database."
-        )
-
     async def on_job_complete(self, job_id: UUID, task: Callable = None, *args):
         """Watches a submitted job and, when it terminates successfully, runs a given task.
 
@@ -101,6 +58,7 @@ class ExperimentService:
         """
         job_status = self._job_service.ray_client.get_job_status(job_id)
 
+        # TODO rely on https://github.com/ray-project/ray/blob/7c2a200ef84f17418666dad43017a82f782596a3/python/ray/dashboard/modules/job/common.py#L53
         valid_status = [
             JobStatus.CREATED.value.lower(),
             JobStatus.PENDING.value.lower(),
@@ -122,7 +80,11 @@ class ExperimentService:
                 task(*args)
 
     def _run_eval(
-        self, inference_job_id: UUID, request: ExperimentCreate, experiment_id: UUID = None
+        self,
+        inference_job_id: UUID,
+        request: ExperimentCreate,
+        background_tasks: BackgroundTasks,
+        experiment_id: UUID = None
     ):
         # use the inference job id to recover the dataset record
         dataset_record = self._dataset_service._get_dataset_record_by_job_id(inference_job_id)
@@ -138,7 +100,9 @@ class ExperimentService:
 
         # submit the job
         self._job_service.create_job(
-            JobEvalLiteCreate.model_validate(job_eval_dict), experiment_id=experiment_id
+            JobEvalLiteCreate.model_validate(job_eval_dict),
+            background_tasks,
+            experiment_id=experiment_id
         )
 
     def create_experiment(
@@ -170,17 +134,8 @@ class ExperimentService:
         # submit inference job first
         job_response = self._job_service.create_job(
             JobInferenceCreate.model_validate(job_inference_dict),
+            background_tasks,
             experiment_id=experiment_record.id,
-        )
-
-        # Inference jobs produce a new dataset
-        # Add the dataset to the (local) database
-        background_tasks.add_task(
-            self.on_job_complete,
-            job_response.id,
-            self._add_dataset_to_db,
-            job_response.id,
-            JobInferenceCreate.model_validate(job_inference_dict),
         )
 
         # run evaluation job afterwards
@@ -191,6 +146,7 @@ class ExperimentService:
             self._run_eval,
             job_response.id,
             request,
+            background_tasks,
             experiment_record.id,
         )
 
--- a/lumigator/python/mzai/backend/backend/services/experiments.py
+++ b/lumigator/python/mzai/backend/backend/services/experiments.py
@@ -12,7 +12,10 @@ from lumigator_schemas.jobs import (
     JobStatus,
 )
 
+from backend.records.experiments import ExperimentRecord
+from backend.records.jobs import JobRecord
 from backend.repositories.experiments import ExperimentRepository
+from backend.repositories.jobs import JobRepository
 from backend.services.datasets import DatasetService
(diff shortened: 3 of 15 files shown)


aittalam:
This PR addresses two issues at the same time. Would it be possible to split this in two separate PRs? This would make it easier to review and address their issues, which are very independent:

- we want to save datasets from annotations ASAP, as this is a requirement for the MVP
- we care less about getting an experiment status from ExperimentService, as the current functionality provided by JobService is equivalent

aittalam (changes requested):
This PR addresses two issues at the same time. Would it be possible to split this in two separate PRs? This would make it easier to review and address their issues, which are very independent:

- we want to save datasets from annotations ASAP, as this is a requirement for the MVP
- we care less about getting an experiment status from ExperimentService, as the current functionality provided by JobService is equivalent

javiermtorres:
> This PR addresses two issues at the same time. Would it be possible to split this in two separate PRs? 
> 

I think this would need more effort than it would yield. The experiment status is implemented in commit https://github.com/mozilla-ai/lumigator/pull/604/commits/327e37a7ed2572698fe94a520762bf796e39c1f4 entirely. I can definitely rearrange the PR into two commits (first 3 + this last one) for easier inspection. Implementing the status at this point also avoids complicating the integration tests unnecessarily, by just checking that the experiment finished instead of checking the individual jobs.

javiermtorres:
As agreed with @aittalam , the last commit with the status change will be addressed in a separate PR.

Makefile
```diff
@@ -149,7 +149,7 @@ test-backend-unit:
 test-backend-integration:
 	cd lumigator/python/mzai/backend/; \
 	docker container list --all; \
-	SQLALCHEMY_DATABASE_URL=sqlite:////tmp/local.db RAY_WORKER_GPUS="0.0" RAY_WORKER_GPUS_FRACTION="0.0" INFERENCE_PIP_REQS=../jobs/inference/requirements_cpu.txt INFERENCE_WORK_DIR=../jobs/inference EVALUATOR_PIP_REQS=../jobs/evaluator/requirements.txt EVALUATOR_WORK_DIR=../jobs/evaluator uv run pytest -s -o python_files="backend/tests/integration/*/test_*.py"
+	SQLALCHEMY_DATABASE_URL=sqlite:////tmp/local.db RAY_WORKER_GPUS="0.0" RAY_WORKER_GPUS_FRACTION="0.0" INFERENCE_PIP_REQS=../jobs/inference/requirements_cpu.txt INFERENCE_WORK_DIR=../jobs/inference EVALUATOR_LITE_PIP_REQS=../jobs/evaluator_lite/requirements.txt EVALUATOR_LITE_WORK_DIR=../jobs/evaluator_lite EVALUATOR_PIP_REQS=../jobs/evaluator/requirements.txt EVALUATOR_WORK_DIR=../jobs/evaluator uv run pytest -s -o python_files="backend/tests/integration/*/test_*.py"
```
aittalam, line 152: Out of curiosity: we have these settings already saved in the main docker compose, which is also used by tests AFAIK. Why do we need to explicitly pass them? Is there a change in the PYTHONPATH?
If not, the backend should still run from `/mzai/lumigator/python/mzai/backend` and `../jobs/..` should match what is provided inside the dockerfile

aittalam: The main reason for trying to understand this is: we should not have redundant ENV settings, otherwise it will be very hard to track where something has been set. Suppose we move evaluator_lite to evaluator (which is what will happen soon), this extra path has to be updated too otherwise things will break

lumigator/python/mzai/backend/backend/config_templates.py
```diff
@@ -98,7 +98,8 @@
         "revision": "{revision}",
         "use_fast": "{use_fast}",
         "trust_remote_code": "{trust_remote_code}",
-        "torch_dtype": "{torch_dtype}"
+        "torch_dtype": "{torch_dtype}",
+        "max_new_tokens": 500
```
aittalam, line 102: You are adding a new parameter here, could you please add some rationale about this for ppl who want to track why this happened?
Why `max_new_tokens` instead of `max_length` or `max_tokens`?
I see that you have added it in templates and job configuration, how does the job makes use of it?

Makefile
javiermtorres: > Why do we need to explicitly pass them? 

Because of the integration tests setting. First, there are some that we want to override for tests (GPUs, req with cpu only). Then the job directories need to be available. I'll dig the exact details but iirc these needed to be resolvable form the backend, which is not in a container during tests so mounts are not an option.

lumigator/python/mzai/backend/backend/services/experiments.py
```diff
@@ -191,11 +145,29 @@ def create_experiment(
...
 
         return ExperimentResponse.model_validate(experiment_record)
 
+    # TODO Move this into a "composite job" impl
+    def get_experiment(self, experiment_id: UUID) -> ExperimentResponse:
+        record = self._get_experiment_record(experiment_id)
+        loguru.logger.info(f"Obtaining info for experiment {experiment_id}: {record}")
+
+        all_succeeded = True
+        for job in self._get_all_owned_jobs(experiment_id):
```
aittalam, line 160: This relates to experiment status and should be logically part of a subsequent PR.
Let's keep it here to avoid further complications, but please be explicit about what you added in this PR in its description.

Makefile
javiermtorres: We don't have redundant ENV settings, we have different deployments and different ways to specify the env vars :) We can definitely try to extract them from the .env file, existing environment or similar. But there are some settings that need specific values like this because they depend on the testing process.

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -64,7 +71,7 @@ def __init__(
         self.job_repo = job_repo
         self.result_repo = result_repo
         self.ray_client = ray_client
-        self.data_service = data_service
+        self._dataset_service = data_service
```
aittalam, line 74:
```suggestion
        self._dataset_service = dataset_service
```
Please also update this in the method parameters for consistency

Makefile
javiermtorres: And this may mean that we need to expand our range of tests, e.g. running some on GPU enabled runners (possibly our own), in release-only CI workflows, maybe, to avoid extra costs.

lumigator/python/mzai/backend/backend/config_templates.py
javiermtorres: I found that it was needed to run the models using the pipeline. I can check it out again. IIRC it went along the lines of https://discuss.huggingface.co/t/confused-about-max-length-and-max-new-tokens/30892

lumigator/python/mzai/backend/backend/services/experiments.py
javiermtorres: Totally true. As I mentioned, the integration tests use this to check if the job is finished, but there's no detailed job status, just CREATED and SUCCEEDED/FAILED (so #571 is actually not addressed). I can still remove this and manually check the experiment's jobs status in the integration test if you prefer.

lumigator/python/mzai/jobs/inference/inference_config.py
```diff
@@ -127,6 +127,7 @@ class HfPipelineConfig(BaseModel, arbitrary_types_allowed=True):
     torch_dtype: TorchDtype = "auto"
     accelerator: Accelerator = Field(title="Accelerator", default=Accelerator.AUTO, exclude=True)
     model_config = ConfigDict(extra="forbid")
+    max_new_tokens: int | None = None
```
aittalam, line 130: Note to ourselves: it might be useful to document how this works wrt `max_tokens` in the SamplingParameters section (the answer is: `max_new_tokens` is used in pipelines, `max_tokens` with causal models in APIs).
Also, we have different parameters (`max_tokens` != `max_new_tokens`) and different default values, shall we keep them consistent?
If you are not planning to document this in this PR no worries, just please track it with an issue 🙏

lumigator/python/mzai/backend/backend/config_templates.py
javiermtorres: IIRC the job directly moves the HF pipeline in the job config to the HF pipeline function, so we're basically handling HF pipeline params.

lumigator/python/mzai/schemas/lumigator_schemas/jobs.py
```diff
@@ -65,6 +65,7 @@ class JobEvalCreate(BaseModel):
     system_prompt: str | None = None
     config_template: str | None = None
     skip_inference: bool = False
+    store_to_dataset: bool = False
```
aittalam, line 68: This is false in schemas and true in experiments, is this intentional?
I personally like this, i.e. the job itself just produces results but the way we deal with this in experiments is that we always want to generate a dataset from them

lumigator/python/mzai/jobs/inference/inference_config.py
javiermtorres: I'd make a new PR, yes, because I feel we'll have significant work tracking pipeline and API params. We could also add basic pipeline sanity unit tests to find these issues early, e.g. when we update library dependencies.

dpoulopoulos: From the Hugging Face docs:

> max_new_tokens (int, optional, defaults to None) — The maximum numbers of tokens to generate, ignore the current number of tokens. Use either max_new_tokens or max_length but not both, they serve the same purpose.

I see that most Hugging Face models that support summarization (e.g., BART, Pegasus, T5, etc.) use the `max_length` attribute in their configs to set a default value. So, would it be better to use `max_length` instead of `max_new_tokens`?

Makefile
aittalam: I agree on the overrides. Just to clarify, I would like to understand why those paths, which are required to run ray jobs, have to be different from the defaults in integration tests, and what is the different deployment that requires this change.

lumigator/python/mzai/backend/backend/config_templates.py
aittalam: Great, thanks!

lumigator/python/mzai/backend/backend/services/experiments.py
aittalam: nonono it's fine! Let's keep it in

lumigator/python/mzai/jobs/inference/inference_config.py
aittalam: That's a good point, TY @dpoulopoulos 🙏

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -265,6 +356,20 @@ def create_job(
         loguru.logger.info("Submitting {job_type} Ray job...")
         submit_ray_job(self.ray_client, entrypoint)
 
+        # Inference jobs produce a new dataset
```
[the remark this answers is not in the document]
aittalam: Agree. Saving to dataset should only work for predictions+annotations

Makefile
javiermtorres: They're not different, but they use different paths. In the container we use `INFERENCE_WORK_DIR=/mzai/lumigator/python/mzai/jobs/inference`because the `/mzai` dir is mounted. This is not the case for integration tests. Maybe we should structure both properly to ensure that they're derived from the same values, but imho that falls out of scope of this PR.

lumigator/python/mzai/jobs/inference/inference_config.py
javiermtorres: It may have been that I initially used a dummy causal model for summarization (to which HF objected but nonetheless ran it). On the other hand, the model we use now for testing (`tiny-random-T5ForConditionalGeneration`) will provide empty predictions, which get misinterpreted by the HF pipeline later on. Also, are these configs recommended by HF themselves? Because it seems they changed their pipeline params and maybe this is still not updated in other examples. Since this seems a simple issue of deriving the new tokens from the input and total tokens, maybe we can do some arithmetic instead of adding the param.

lumigator/python/mzai/schemas/lumigator_schemas/jobs.py
javiermtorres: Yes, it's intentional 👍

lumigator/python/mzai/backend/backend/services/jobs.py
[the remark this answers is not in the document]
javiermtorres: I didn't want to limit our options, nor introduce another "if job type is X" dependency to remove when we make a more generic job handling. What we _can_ do is remove the `store_to_dataset` option in the JobEval/LiteCreate so the eval jobs don't store anything (again, I added it to evals to keep our options open).

lumigator/python/mzai/jobs/inference/inference_config.py
javiermtorres: For reference: https://discuss.huggingface.co/t/confused-about-max-length-and-max-new-tokens/30892

lumigator/python/mzai/backend/backend/services/jobs.py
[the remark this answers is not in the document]
javiermtorres: Since the option to create a dataset has been explicitly removed from evals, an "if job_type == inference" has been added.

aittalam (approved):
I preapprove this to avoid blocking. @javiermtorres @agpituk per our conversation, you choose whether you prefer to tackle the params issue first or postpone it to a subsequent PR. TY @javiermtorres !

lumigator/python/mzai/jobs/inference/inference_config.py
javiermtorres: FYI we get the following error when not setting a max_length, which I can't find in the settings anyway:

```
ValueError: Input length of input_ids is 199, but `max_length` is set to 20. This can lead to unexpected behavior. You should consider increasing `max_length` or, better yet, setting `max_new_tokens`.
```

javiermtorres: Using `max_length` instead of `max_new_tokens`. Let's keep an eye on these params in case they get deprecated.

javiermtorres:
> I preapprove this to avoid blocking. @javiermtorres @agpituk per our conversation, you choose whether you prefer to tackle the params issue first or postpone it to a subsequent PR. TY @javiermtorres !

As mentioned, I'll open a PR to verify the HF pipeline params we use/should use. For now we're going with `max_length` which seems to be the approved method and works ok, although the library issues a warning.

PR #605, opened 2025-01-10
Add support for Debugpy when in development mode

njbrake:
I'm sorry for the adding of all the PR reviewers, not quite sure yet who are the appropriate people to request reviews from.

I generally like to use a debugger so that I can follow the logic flow and inspect variables. Since the dev environment runs in docker, this PR is to expose and run debugpy when in dev mode (`make local-up`) so that we can have enhanced debugging tools.

I'm not sure where this would be documented since it doesn't appear that we yet have a source doc for a "developer guide" or anything like it yet. Perhaps for the backlog.

# What's changing

* Adds connection for debugpy when running in dev mode ("make local-up")
# How to test it

Steps to test the changes:

1. Run `make local-up`
2. In vscode, run the debug config "Attach to backend", set and hit a breakpoint

I saw in https://github.com/mozilla-ai/lumigator/commit/30c5ba6d8bd76e98479713a266214e84904f5748 that @dpoulopoulos  gitignored the launch.json, but wasn't sure of the reason. We could continue to ignore the launch.json but it might be helpful for integrating with this effort? If that's an issue, happy to undo that edit to the gitignore.

# I already...

- [🤷🏼] Added some tests for any new functionality
- [🤷🏼] Updated the documentation (both comments in code and [product documentation](https://mozilla-ai.github.io/lumigator) under `/docs`)

### DIFF
--- a/.devcontainer/docker-compose.override.yaml
+++ b/.devcontainer/docker-compose.override.yaml
@@ -12,7 +12,7 @@ services:
     ports:
       - "5678:5678"
     environment:
-      - DEBUG=1
+      - ENABLE_DEBUGPY=1
     develop:
       watch:
         - path: lumigator/python/mzai/backend/
--- a/.devcontainer/docker-compose.override.yaml
+++ b/.devcontainer/docker-compose.override.yaml
@@ -9,6 +9,10 @@ services:
       target: "dev_image"
     volumes:
       -  database_volume:/mzai/backend/local.db
+    ports:
+      - "5678:5678"
+    environment:
+      - DEBUG=1
     develop:
       watch:
         - path: lumigator/python/mzai/backend/
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,23 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Attach to Backend",
+            "type": "debugpy",
+            "request": "attach",
+            "connect": {
+                "host": "localhost",
+                "port": 5678
+            },
+            "pathMappings": [
+                {
+                    "localRoot": "${workspaceFolder}",
+                    "remoteRoot": "/mzai"
+                }
+            ]
+        }
+    ]
+}
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,6 +8,7 @@
             "name": "Attach to Backend",
             "type": "debugpy",
             "request": "attach",
+            "justMyCode":  false,
             "connect": {
                 "host": "localhost",
                 "port": 5678
--- a/lumigator/python/mzai/backend/backend/main.py
+++ b/lumigator/python/mzai/backend/backend/main.py
@@ -67,7 +67,7 @@ def create_app() -> FastAPI:
 
     app.include_router(api_router)
 
-    if os.environ.get("DEBUG"):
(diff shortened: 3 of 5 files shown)


dpoulopoulos:
@njbrake `launch.json` is a VSCode-specific configuration file, and not everyone uses VSCode. That was the reasoning behind ignoring such files. However, I don’t have a strong opinion on this; we could just as easily commit the file to have it handy when needed.

lumigator/python/mzai/backend/backend/main.py
```diff
@@ -67,6 +67,13 @@ def create_app() -> FastAPI:
 
     app.include_router(api_router)
 
+    if os.environ.get("DEBUG"):
+        import debugpy  # type: ignore
+
+        debug_port = os.environ.get("DEBUG_PORT", 5678)
+        logger.info(f"Starting debugpy on port {debug_port}")
+        debugpy.listen(("0.0.0.0", debug_port))
```
dpoulopoulos, line 75: Let’s avoid hardcoding the host and port. Perhaps backend.settings is a better place to define sensible defaults and read them from there.

.devcontainer/docker-compose.override.yaml
```diff
@@ -9,6 +9,10 @@ services:
       target: "dev_image"
     volumes:
       -  database_volume:/mzai/backend/local.db
+    ports:
+      - "5678:5678"
+    environment:
+      - DEBUG=1
```
dpoulopoulos, line 15: Would you consider passing this as a flag in the Makefile? Typically, when building something, you indicate whether it’s a release or a debug build through the build tool. I know Python isn’t like C++, but what do you think?

.vscode/launch.json
```diff
@@ -0,0 +1,23 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Attach to Backend",
+            "type": "debugpy",
+            "request": "attach",
```
dpoulopoulos, line 10: Can we use the `justMyCode` attribute here? I usually set this to `false` so I can trace the path to other dependencies as well. For example, one of the dependencies here is `lumigator_schemas`, which is also *our* code.

dpoulopoulos (changes requested):

lumigator/python/mzai/backend/backend/main.py
```diff
@@ -67,6 +67,13 @@ def create_app() -> FastAPI:
 
     app.include_router(api_router)
 
+    if os.environ.get("DEBUG"):
```
njbrake, line 70:
```suggestion
    if os.environ.get("ENABLE_DEBUGPY"):
```

veekaybee:
Thanks for this! I'm not familiar with this package so I have a couple questions: 

- would this allow for attaching debuggers in IDEs other than VS code, and how might debugging work if you're using something like vi?
- does this cover the ability to debug remote code such as Ray jobs?
- It looks like we might not need to include code in the route path itself, but can pass it via the Makefile as `CMD ["python3", "-m", "debugpy", "--listen", "0.0.0.0:5678", "-m", "uvicorn", "main:app", "--reload", "--host", "0.0.0.0", "--port", "8000"]` via a local start-up. I'd prefer that for a an environment-based piece of code,  what do you think about this approach?  https://github.com/mozilla-ai/lumigator/blob/eeef145d904ee920be45cc977670721d1b222682/Dockerfile#L21

njbrake:
(Renamed the branch and created an issue to align to team standards)

PR #608, opened 2025-01-13
Pipeline refactor

macaab26:
# What's changing

* Updated section...
* Added new...
* Fixed a typo in...
* ...

Refs #...
Closes #...

# How to test it

Steps to test the changes:

1.
2.
3.

# I already...


.github/workflows/lumigator_pipeline.yaml
```diff
@@ -0,0 +1,287 @@
...
+        run: make test-sdk-unit
+
+  integration-tests:
+    name: Integration tests
+    runs-on: ubuntu-latest
+    needs: lint
+    if: ${{ needs.lint.result == 'success' }}
+    strategy:
+      matrix:
+        test: ["sdk", "backend" ]
```
javiermtorres, line 83: Can we factor unit tests in the same way?

macaab26: Good idea, probably yes, let me try

javiermtorres: Are we triggering integration tests on any kind of change? Would doc-only changes make it unnecessary to run tests? In this case, changes to files like the Makefile or the .env.example file should also trigger tests.

javiermtorres (approved):
LGTM. Some comments, but I think we can polish this in future PRs if needed.

.github/workflows/lumigator_pipeline.yaml
macaab26: Yeah, I discussed it with @agpituk, and we both agree that the direction is to be more "restrictive" with the pipeline success, we are going to run all tests with all changes. But yeah, I get your point and I agree, we should trigger it only with **application code** changes, not with changes in documentation (just an example)

macaab26: I just realized that we're going to require a successful execution of the pipeline to merge the PR, and as far as I know you can't make that requirement path dependent, so, yes, we will have to run everything, even with changes in the documentation.

PR #610, opened 2025-01-13
Use nginx to send Lumigator requests from the FE to any backend deployment

### LINKED ISSUE #553: Configure frontend container to send request to different backends
### Motivation

Right now our frontend container will always point to this [variable](https://github.com/mozilla-ai/lumigator/blob/0500d5e9ecd757d8aac91f7ea9a60ebe79ce0c9f/.env.example#L35) when connecting to the backend. This presents an issue when we try to deploy lumigator in a complex environment, meaning we would need specific builds per environment. Instead, we should use the nginx in the frontend container to use an environment variable that points to the right backend.

This means, the default of the vue_app_base_url could be /api/v1/ and the nginx can route that on runtime to a different backend. This works in local installations, hybrid, and cloud installations. 

### Alternatives

_No response_

### Contribution

_No response_

agpituk:
# What's changing

In this PR I'm changing how we use NGINX to ensure we have the ability to deploy Lumigator on different environments. As currentlue VUE gets our backend api host on build time, it was impossible for us to deploy this on kubernetes without creating specific images for each environment. 

In this PR I've changed how nginx container to make it work with env variables. Later, in our helm chart, we can use this to specify where the backend is.

Refs #553 
Closes #553 

# How to test it

Steps to test the changes:

1. Ensure you can still run lumigator locally with all different commands and the requests get routed to the API properly

We still haven't  removed the VUE_BASE_APP_URL from the .env in file in this PR, but it's only used now in development mode. 

# I already...


agpituk:
A comment for the reviewers. Nginx by default doesn't support environment variables, but you can inject them in a docker container if you put them in the right templats You can read more in the section `Using environment variables in nginx configuration` of [https://hub.docker.com/_/nginx](https://hub.docker.com/_/nginx)

chainlink (approved):

PR #612, opened 2025-01-13
[FEATURE] Support for Debugpy

njbrake:
Sorry this is a re-open of PR-605: Making this PR align to the team expectations of PRs.

I generally like to use a debugger so that I can follow the logic flow and inspect variables. Since the dev environment runs in docker, this PR is to expose and run debugpy when in dev mode (`make local-up`) so that we can have enhanced debugging tools.

I'm not sure where this would be documented since it doesn't appear that we yet have a source doc for a "developer guide" or anything like it yet. Perhaps for the backlog.

# What's changing

* Adds connection for debugpy when running in dev mode ("make local-up")
# How to test it

Steps to test the changes:

1. Run `make local-up`
2. In vscode, run the debug config "Attach to backend", set and hit a breakpoint

I saw in https://github.com/mozilla-ai/lumigator/commit/30c5ba6d8bd76e98479713a266214e84904f5748 that @dpoulopoulos  gitignored the launch.json, but wasn't sure of the reason. We could continue to ignore the launch.json but it might be helpful for integrating with this effort? If that's an issue, happy to undo that edit to the gitignore.

# I already...

- [🤷🏼] Added some tests for any new functionality
- [🤷🏼] Updated the documentation (both comments in code and [product documentation](https://mozilla-ai.github.io/lumigator) under `/docs`)


veekaybee:
Same comments as here, https://github.com/mozilla-ai/lumigator/pull/605#issuecomment-2587197627

Thanks for this! I'm not familiar with this package so I have a couple questions: 

- would this allow for attaching debuggers in IDEs other than VS code, and how might debugging work if you're using something like vi?
- does this cover the ability to debug remote code such as Ray jobs?
- It looks like we might not need to include code in the route path itself, but can pass it via the Makefile as `CMD ["python3", "-m", "debugpy", "--listen", "0.0.0.0:5678", "-m", "uvicorn", "main:app", "--reload", "--host", "0.0.0.0", "--port", "8000"]` via a local start-up. I'd prefer that for a an environment-based piece of code,  what do you think about this approach?  https://github.com/mozilla-ai/lumigator/blob/eeef145d904ee920be45cc977670721d1b222682/Dockerfile#L21

dpoulopoulos:
>It looks like we might not need to include code in the route path itself, but can pass it via the Makefile as CMD ["python3", "-m", "debugpy", "--listen", "0.0.0.0:5678", "-m", "uvicorn", "main:app", "--reload", "--host", "0.0.0.0", "--port", "8000"] via a local start-up. I'd prefer that for a an environment-based piece of code, what do you think about this approach? 

+100 to this!

njbrake:
@veekaybee Thank you for the review!

>     * would this allow for attaching debuggers in IDEs other than VS code, and how might debugging work if you're using something like vi?
It looks like PyCharm has yet to support the feature, but it's a commonly +1'd [feature request ](https://youtrack.jetbrains.com/issue/PY-63403/Support-debugpy).
As far as I can tell, debugpy is designed to work with VSCode, I don't think it has a shell interface (though I've never tried).

----------
>     * does this cover the ability to debug remote code such as Ray jobs?
It does not. I have some ideas about how we might do this (and I have a hacky implementation locally to do it), but I don't yet have my head wrapped around the motivation behind the way the Ray interface was designed. My previous experience with Ray used @ray.remote() and the Ray sdk to pass all of the logic to run on the ray cluster, while the current design of Lumigator is passing a script command and relying on the Ray cluster to already have the python code installed in specific locations, which makes the debugpy setup more convoluted. 

----------
>     * It looks like we might not need to include code in the route path itself, but can pass it via the Makefile as `CMD ["python3", "-m", "debugpy", "--listen", "0.0.0.0:5678", "-m", "uvicorn", "main:app", "--reload", "--host", "0.0.0.0", "--port", "8000"]` via a local start-up. I'd prefer that for a an environment-based piece of code,  what do you think about this approach?  https://github.com/mozilla-ai/lumigator/blob/eeef145d904ee920be45cc977670721d1b222682/Dockerfile#L21
Personally I prefer it to be in the python code, because I am thinking that we will also probably want to add some other dev/debug type logic to allow for toggling on debug logging for instance, so that way we can group more of the logic around debug setup in one place (and I like generally to put this in python as opposed to bash or a dockerfile). However, I don't have a strong opinion and this isn't an area that I have too much experience with so if you or @dpoulopoulos would like it shifted to the dockerfile, I'm happy to make that change?

veekaybee:
> Personally I prefer it to be in the python code, because I am thinking that we will also probably want to add some other dev/debug type logic to allow for toggling on debug logging for instance, so that way we can group more of the logic around debug setup in one place (and I like generally to put this in python as opposed to bash or a dockerfile). However, I don't have a strong opinion and this isn't an area that I have too much experience with so if you or @dpoulopoulos would like it shifted to the dockerfile, I'm happy to make that change?

Thanks for adding this context! Given that this covers only one IDE, a smaller part of the codebase, and only at dev time, I'd opt for keeping it at the dev-time install boundaries of the Dockerfile rather than the entire application itself for now. If we find a debugging tool that we can cover at least 50% of the app, it will be good to rethink this strategy.

veekaybee (approved):
IMO this looks really clean, thanks for taking this! LGTM.

Issue #472, opened 2024-12-04
[BUG]: `S3_BUCKET` config setting is ignored


peteski22:
### Description

When `S3_BUCKET` is configured in the `.env` file it should be used in Lumigator, however even the default setting will be ignored currently and `"lumigator-storage"` will always be used:

https://github.com/mozilla-ai/lumigator/blob/e42fd90f25aa232c00da8b821900741b4daed837/lumigator/python/mzai/backend/backend/settings.py#L27

This value is used in methods such as `def _get_s3_path` etc.

### Reproduction

Clone Lumigator and `cd` into the repository directory.

Export `S3_BUCKET`:

```
export S3_BUCKET="somewhere"
```

Set the S3 bucket value like this in the `.env` file:

```
S3_BUCKET=${S3_BUCKET:-somewhere-else}
```

```
make start-lumigator-build
```

Expect to see a bucket has been created named: `somewhere` but it will be named `lumigator-storage` (using [s5cmd ls](https://github.com/peak/s5cmd)).

```
make stop-lumigator
```

(Just to double check against the default in the `.env` file that is used when there's no environment variable set...

unset S3_BUCKET

```
make start-lumigator-build
```

Expect to see a bucket has been created named: `somewhere-else` but it will be named `lumigator-storage`.

### Relevant log output

```shell
NA
```

### Expected behavior

Lumigator should respect the setting in the `.env` file (supplied, or the default value) and not default to a hard coded string (`lumigator-storage`)

### System Info

NA

### Have you searched for similar issues before submitting this one?

- [X] Yes, I have searched for similar issues

peteski22:
Closed by https://github.com/mozilla-ai/lumigator/pull/614

https://github.com/mozilla-ai/lumigator/pull/614/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R156

PR #614, opened 2025-01-13
Minor consolidation of Environment Variables

njbrake:
# What's changing
This pull request involves several changes to update environment configuration files and improve the setup for local development and testing. The key changes include renaming the environment template file, updating references across the codebase, and modifying environment variable defaults.

### Environment Configuration Updates:

* `.env.example` renamed to `.env.template` and updated with additional comments and placeholders for various configuration sections. [[1]](diffhunk://#diff-749e06f64632f62a0c0dfbf4c4f3850e27e94ac109aa121fabd5c29469ae88deL1-R12) [[2]](diffhunk://#diff-749e06f64632f62a0c0dfbf4c4f3850e27e94ac109aa121fabd5c29469ae88deR22-R62)

### Makefile Adjustments:

* Updated `Makefile` to reference the new `.env.template` file instead of the old `.env.example` file.
* Added necessary environment variables for backend unit and integration tests in the `Makefile`.

### Documentation Update:

* Updated the reference to the environment file in `docs/source/user-guides/inference.md` from `.env.example` to `.env.template`.

### Backend Settings Update:

* Modified `lumigator/python/mzai/backend/backend/settings.py` to remove default values for `S3_BUCKET`, `RAY_HEAD_NODE_HOST`, and `RAY_DASHBOARD_PORT`, indicating that these should be specified in the `.env` file.
# How to test it

Steps to test the changes:

1. Run `make local-up`, check to see if .env file variables are correctly loaded.

The idea of this PR is that now the environment variable file is only used for overrides, not for regular settings.

# I already...


Makefile
```diff
@@ -65,7 +65,7 @@ LOCAL_DOCKERCOMPOSE_FILE:= docker-compose.yaml
 DEV_DOCKER_COMPOSE_FILE:= .devcontainer/docker-compose.override.yaml
 
 .env:
-	@if [ ! -f .env ]; then cp .env.example .env; echo ".env created from .env.example"; fi
+	@if [ ! -f .env ]; then cp .env.template .env; echo ".env created from .env.template"; fi
```
javiermtorres, line 68: Now that I read it... Would this target even fire if the .env file were present? So the `if` would be really redundant :-/ I'll try to check it out.

Makefile
```diff
@@ -123,12 +123,26 @@ endif
 
 test-backend-unit:
 	cd lumigator/python/mzai/backend/; \
-	SQLALCHEMY_DATABASE_URL=sqlite:////tmp/local.db uv run pytest -o python_files="backend/tests/unit/*/test_*.py"
+	S3_BUCKET=lumigator-storage \
+	RAY_HEAD_NODE_HOST=localhost \
+	RAY_DASHBOARD_PORT=8265 \
+	SQLALCHEMY_DATABASE_URL=sqlite:////tmp/local.db \
+	uv run pytest -o python_files="backend/tests/unit/*/test_*.py"
```
javiermtorres, line 130: @aittalam @veekaybee did we speak about moving the necessary config to the conftest instead for tests????

javiermtorres (commented):
LGTM. I'll wait for other opinions before approving, though.
I agree with keeping defaults in a single point, and adding configuration if needed.

Makefile
njbrake: this `.env` target is optionally run before a subset of tests. However, to your point it is a confusing target name. I will update the target name to make it a little clearer.

agpituk: What's the reasoning behind renaming `.env` to `check-dot-env`? As this target is actually the only one in the Makefile that needs to creat a file, I think it makes a lot of sense that to keep the .env name

agpituk:
I think this is a great consolidation PR @njbrake ! Just added a minor comment on the Makefile. Also, what do you think about also having the `SQLALCHEMY_DATABASE_URL` in the `.env `file instead of the `docker-compose.yaml`? in order to consolidate things in one place.

Makefile
njbrake: The rest of our makefile commands are verbs (check-blah, something-start, you-down) so this makes it follow that implicit naming protocol.

I also just generally like having target names be something that gives some indication about what it's doing. In this case, the command isn't loading the .env file (which you would probably guess if you didn't know the code), it's checking and creating the .env file if it doesn't exist.

njbrake:
> I think this is a great consolidation PR @njbrake ! Just added a minor comment on the Makefile. Also, what do you think about also having the `SQLALCHEMY_DATABASE_URL` in the `.env `file instead of the `docker-compose.yaml`? in order to consolidate things in one place.

@agpituk  would it be ok if we push that change off into a future change? Tbh I'm not totally sure how the database setup works yet, and that env var shows up in enough places that I'm not yet confident that I won't accidentally break something when trying to change it 😆 .

Makefile
agpituk: I get what you mean, but the Make targets are either files, or verbs that we add to the `.PHONY` target (1st line of the file). The way we use Make I think what you are suggesting could make sense, but if we do that, it definitely has to be in the `.PHONY` target

agpituk (changes requested):

agpituk (approved):

Makefile
njbrake: Deferring this conversation to a later cleanup effort :)

peteski22:
Closes: #472

PR #623, opened 2025-01-14
Fix runner FE builds

macaab26:
# What's changing

We are including the "VUE_APP_BASE_URL" var for the runner builds. 

# How to test it

Try to run make start-lumigator with the image tag generated with the check.

# I already...


.github/workflows/build_images_frontend.yaml
```diff
@@ -52,6 +52,8 @@ jobs:
           file: "lumigator/frontend/Dockerfile"
           target: "server"
           push: true
+          build-args: |
+            "VUE_APP_BASE_URL=http://localhost/api/v1/"
```
agpituk, line 56: Shouldn't it be: `VUE_APP_BASE_URL=http://localhost:8000/api/v1/` ? (as in the docker-compose). When my other PR is merged, this will definitely be the right one (VUE_APP_BASE_URL=http://localhost/api/v1/) so nginx will route it to where the BE is based on the env variable, so localhost without port will be correct, but now it's always going to be running on 8000

macaab26: Hey @agpituk yeah, you are right, as we discussed this morning, I expected your PR to be merged first. Based on what we discussed, that PR is going to be delayed for a couple of days, so I will change it

agpituk (approved):

PR #624, opened 2025-01-14
fix: crash if requested output field for inference doesn't exist in dataset

### LINKED ISSUE #585: [BUG]: Inference does not work when column ground_truth is missing
### Description

If one tries to launch an inference job via /jobs/inference on a dataset that does **not** have a ground_truth column with the request to call the output_field ground_truth, this service fails (see below).

### Reproduction

1. Upload file without gt
2. With a running lumigator, request an inference job for 2 samples, with bart, on that dataset. Ensure output_fiedl is set to "ground_truth".
3. See job fail

Attached:
* input file without gt column

[dialogsum_mini_no_gt.csv](https://github.com/user-attachments/files/18354140/dialogsum_mini_no_gt.csv)

### Relevant log output

```shell
Error when there is no gt field:


Job entrypoint command failed with exit code 1, last available logs (truncated to 20,000 chars):
    return self._getitem(key)
           ^^^^^^^^^^^^^^^^^^
  File "/tmp/ray/session_2025-01-08_14-52-23_572617_1/runtime_resources/pip/617697551215d8488f42ccab087d2364573ba842/virtualenv/lib/python3.11/site-packages/datasets/arrow_dataset.py", line 2850, in _getitem
    pa_subtable = query_table(self._data, key, indices=self._indices)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/ray/session_2025-01-08_14-52-23_572617_1/runtime_resources/pip/617697551215d8488f42ccab087d2364573ba842/virtualenv/lib/python3.11/site-packages/datasets/formatting/formatting.py", line 584, in query_table
    _check_valid_column_key(key, table.column_names)
  File "/tmp/ray/session_2025-01-08_14-52-23_572617_1/runtime_resources/pip/617697551215d8488f42ccab087d2364573ba842/virtualenv/lib/python3.11/site-packages/datasets/formatting/formatting.py", line 521, in _check_valid_column_key
    raise KeyError(f"Column {key} not in the dataset. Current columns in the dataset: {columns}")
KeyError: "Column ground_truth not in the dataset. Current columns in the dataset: ['examples', '__index_level_0__']"
```

### Expected behavior
No crash, but predictions correctly added to the file. 

Undefined (⚠ ) and out of scope: what should happen if the API user requests an output_field that already exists and is not empty: should we overwrite, since we're saving as a separate "dataset" anyway? Return an error?

### System Info

* macOS
* current main

ividal:
# What's changing

If a field is requested for outputs and it doesn't exist in the input dataset (which is to be expected), inference crashes (and, by extension, so does annotation).

This PR:
* copies every content from the original dataset to the new one that should be stored (so the user can always delete the previous one if they prefer)
* creates or **overwrites** the desired column. This changes the previous behaviour of keeping whatever was in the existing column (which meant we did inference for nothing if there were already predictions/annotations there).

Closes #585

# How to test it

Steps to test the changes:

1. Upload the csv listed in the issue - keep the dataset id
2. Launch via API an inference job with any model, using the dataset id
3. The inference job should succeed.

This changes the previous behaviour of keeping whatever was in the existing column (which meant we did inference for nothing if there were already predictions/annotations there).

# I already...

### DIFF
--- a/lumigator/python/mzai/jobs/inference/inference.py
+++ b/lumigator/python/mzai/jobs/inference/inference.py
@@ -119,7 +119,7 @@ def run_inference(config: InferenceJobConfig) -> Path:
     # they selected, we overwrite it with the values from our inference.
 
     if config.job.output_field in dataset.column_names:
-        logger.info(f"Overwriting {config.job.output_field}")
+        logger.warning(f"Overwriting {config.job.output_field}")
 
     output[config.job.output_field] = predict(dataset_iterable, model_client)
     output["model"] = output_model_name
--- a/lumigator/python/mzai/jobs/inference/inference.py
+++ b/lumigator/python/mzai/jobs/inference/inference.py
@@ -109,10 +109,19 @@ def run_inference(config: InferenceJobConfig) -> Path:
     else:
         raise NotImplementedError("Inference pipeline not supported.")
 
-    # run inference
+    # We keep any columns that were already there (not just the original input
+    # samples, but also past predictions under another column name)
+    for k in dataset.column_names:
+        logger.info(f"Keeping original dataset's {k}")
+        output[k] = dataset[k]
+
+    # We are trusting the user: if the dataset already had a column with the output_field
+    # they selected, we overwrite it with the values from our inference.
+
+    if config.job.output_field in dataset.column_names:
+        logger.info(f"Overwriting {config.job.output_field}")
+
     output[config.job.output_field] = predict(dataset_iterable, model_client)
-    output["examples"] = dataset["examples"]
-    output["ground_truth"] = dataset["ground_truth"]
     output["model"] = output_model_name
 
     output_path = save_outputs(config, output)
--- a/lumigator/python/mzai/jobs/inference/inference.py
+++ b/lumigator/python/mzai/jobs/inference/inference.py
@@ -111,9 +111,7 @@ def run_inference(config: InferenceJobConfig) -> Path:
 
     # We keep any columns that were already there (not just the original input
     # samples, but also past predictions under another column name)
-    for k in dataset.column_names:
-        logger.info(f"Keeping original dataset's {k}")
-        output[k] = dataset[k]
+    output.update(dataset.to_dict())
 
     # We are trusting the user: if the dataset already had a column with the output_field
     # they selected, we overwrite it with the values from our inference.
--- a/lumigator/python/mzai/backend/backend/tests/conftest.py
+++ b/lumigator/python/mzai/backend/backend/tests/conftest.py
@@ -94,8 +94,15 @@ def dialog_dataset():
 
 
 @pytest.fixture(scope="function")
-def dialog_no_gt_dataset():
+def dialog_empty_gt_dataset():
     filename = common_resources_dir() / "sample_data" / "dialogsum_mini_empty_gt.csv"
+    with Path(filename).open("rb") as f:
+        yield f
+
+
+@pytest.fixture(scope="function")
+def dialog_no_gt_dataset():
+    filename = common_resources_dir() / "sample_data" / "dialogsum_mini_no_gt.csv"
     with Path(filename).open("rb") as f:
         yield f
 
--- a/lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
+++ b/lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
@@ -135,17 +135,20 @@ def test_upload_data_launch_job(
     logger.info(f"#{logs_infer_job_response_model.logs}#")
 
 
+@pytest.mark.parametrize("unnanotated_dataset", ["dialog_empty_gt_dataset", "dialog_no_gt_dataset"])
 def test_upload_data_no_gt_launch_annotation(
+    request: pytest.FixtureRequest,
     local_client: TestClient,
-    dialog_no_gt_dataset,
+    unnanotated_dataset,
     simple_eval_template,
     simple_infer_template,
     dependency_overrides_services,
 ):
+    dataset = request.getfixturevalue(unnanotated_dataset)
     create_response = local_client.post(
         "/datasets/",
         data={},
-        files={"dataset": dialog_no_gt_dataset, "format": (None, DatasetFormat.JOB.value)},
+        files={"dataset": dataset, "format": (None, DatasetFormat.JOB.value)},
     )
 
     assert create_response.status_code == 201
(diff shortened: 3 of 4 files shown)


lumigator/python/mzai/jobs/inference/inference.py
```diff
@@ -109,10 +109,17 @@ def run_inference(config: InferenceJobConfig) -> Path:
     else:
         raise NotImplementedError("Inference pipeline not supported.")
 
+    # we'll keep data and any previous predictions/annotations
+    for k in dataset:
+        output[k] = dataset[k]
```
javiermtorres, line 114:
```suggestion
    for k in ['examples', 'ground_truth']:
        output[k] = dataset[k]
```

aittalam: @javiermtorres this fix brings the code back to the previous state, which was breaking because we assumed we had the `ground_truth` field (usually not the case for annotations). I'd suggest

```suggestion
    for k in dataset.column_names:
        output[k] = dataset[k]
```

I have not tested it but you should be able to do so without actually running lumigator. Just cd to the job directory, create e.g. a conda env, `pip install -r requirements` and run `python inference.py --config ...` (you can copy/paste a configuration from an inference/annotation job you ran previously).

ividal: Stumbled upon it too, thanks! I remembered late that we were dealing with a [hf dataset](https://huggingface.co/docs/datasets/v2.1.0/en/access).

@javiermtorres Thanks, but that suggestion still ignored we weren't dealing with a dict! Also, the reason I wanted to keep all columns is it seems we want to keep all annotations in a single file, so we'll need to be flexible and copy more than examples and ground_truth. (A separate topic I think we should discuss is dropping the notion of ground_truth... not just because it's LLM-generated, but because it's too generic: "ground_truth for what task exactly?")

ividal:
⚠️ Rebased.

lumigator/python/mzai/jobs/inference/inference.py
```diff
@@ -109,10 +109,19 @@ def run_inference(config: InferenceJobConfig) -> Path:
     else:
         raise NotImplementedError("Inference pipeline not supported.")
 
-    # run inference
+    # We keep any columns that were already there (not just the original input
+    # samples, but also past predictions under another column name)
+    for k in dataset.column_names:
```
dpoulopoulos, line 114: Can we avoid the for loop for this?

```python
logger.info(f"Keeping original dataset's columns: {', '.join(dataset.column_names)}")
output.update(dataset.to_dict())
```

I think it's doing the same thing, but I feel it's a bit cleaner.

lumigator/python/mzai/jobs/inference/inference.py
```diff
@@ -109,10 +109,19 @@ def run_inference(config: InferenceJobConfig) -> Path:
...
+        logger.info(f"Keeping original dataset's {k}")
+        output[k] = dataset[k]
+
+    # We are trusting the user: if the dataset already had a column with the output_field
+    # they selected, we overwrite it with the values from our inference.
+
+    if config.job.output_field in dataset.column_names:
+        logger.info(f"Overwriting {config.job.output_field}")
```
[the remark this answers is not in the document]
aittalam: Yes, as this is logged by the job this should appear in the Ray logs, not on the backend ones. You should be able to see it on the dashboard or in the UI (if not, you might want to restart the containers as until this morning the jobs dir was not watched 😅)

I think this is one of those cases where we'd like to break early: in the UI, if the dataset already has a `ground_truth` field, we should warn users and allow them to choose. If they choose to overwrite, then I think it should be fine (we'll still log it for the records). WDYT?

lumigator/python/mzai/jobs/inference/inference.py
aittalam: Oh yeah using the dict directly looks so much nicer! My bad for suggesting `column_names` instead.
I'd still do the overwrite thing though, if you all agree with my comment above. The main rationale is that otherwise we'd run a full inference job -which might take a while- and then decide to throw everything away

aittalam (approved):
TY Irina for this PR!

I left a couple of minor comments but once they are addressed (and code from main is merged again 😅) this LGTM 🙏

lumigator/python/mzai/jobs/inference/inference.py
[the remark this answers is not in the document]
dpoulopoulos: Sure! But do we have the logic on the backend to allow them to choose if they want to overwrite it? Or are we delegating this completely to the frontend?

lumigator/python/mzai/jobs/inference/inference.py
dpoulopoulos: Yes, I don't object to the overwrite thing. I left a comment.

ividal: Don't feel bad @aittalam  I had reached the same conclusion, not your fault :D 

@dpoulopoulos , I did try updating instead, but it breaks the job in ray - not sure if maybe the content is being exploded into a list of chars? This for was the only thing that was working.

And note to self: we need 1) a more convenient way of testing jobs without ray, 2) a more convenient way of retrieving logs to see what's breaking inside a job while developing...

In the meantime, I'll get the logs from the CI.

lumigator/python/mzai/jobs/inference/inference.py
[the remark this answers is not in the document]
ividal: We don't have that logic. We could add a "force" field (or override, whatever) post-MVP.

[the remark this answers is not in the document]
ividal: >  If they choose to overwrite, then I think it should be fine (we'll still log it for the records).

FWITW, I think it should be fine because (and as long as) we are saving the results as a separate dataset. The two have to go together: doing that, we trust the user and they can always choose to keep or delete the original, untouched, dataset. 

If we let them truly modify existing datasets (which I really prefer we don't do), then we need them to decide on the fly.

lumigator/python/mzai/jobs/inference/inference.py
ividal: ... and the issues observed are gone. I'll keep the update with to_dict if CI is green!

javiermtorres (approved):
LGTM

lumigator/python/mzai/jobs/inference/inference.py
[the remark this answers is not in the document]
ividal: To recap, this is now a warning anyhow, as @dpoulopoulos suggested. We will revisit how to deal with new datasets rather soon, since aggregating annotations in a dataset, plus deleting datasets and other changes to the flow is something that will come up already with translation.

PR #627, opened 2025-01-14
Frontend: Change to relative backend

chainlink:
# What's changing

This changes the front end from having a hard coded path to the backend to accessing the backend on a path relative to the backend

Before

```
baseURL: 'http://my-backend-domain:8000/api/v1'
```

After

```
baseURL: '/api/v1'
```

This worked before, since with docker compose, the front end build occurs just before runtime, and `VUE_APP_BASE_URL` gets passed in. This doesn't work in environments where the front end is built prior and shipped in a docker container (like kubernetes) **This change prevents us from having to do a custom front end build for each environment**

This will need to be used in conjunction with https://github.com/mozilla-ai/lumigator/pull/610 

Cases covered
- [x] local dev mode uses [vite proxy](https://github.com/mozilla-ai/lumigator/blob/5bd9e85c25139134221229b30dab03549864eb4d/lumigator/frontend/vite.config.js#L29)
- [x] Docker compose uses [nginx proxy pass](https://github.com/mozilla-ai/lumigator/pull/610)
- [x] Kube will use nginx proxy pass as well (forthcoming helm chart)

Refs #...
Closes #...

# How to test it

Steps to test the changes:

Run `make start-lumigator-build` locally

- [x] Try uploading a dataset
- [x] Try running an experiment

N/A

# I already...

### DIFF
--- a/lumigator/frontend/src/services/http/index.js
+++ b/lumigator/frontend/src/services/http/index.js
@@ -1,7 +1,7 @@
 import axios from 'axios';
 
 const http = axios.create({
-  baseURL: import.meta.env.VUE_APP_BASE_URL, // API baseURL as env variable
+  baseURL: '/api/v1/', // Set to relative path to the backend
   headers: {
     'Content-Type': 'application/json',
   },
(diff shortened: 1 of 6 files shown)


lumigator/frontend/src/services/http/index.js
```diff
@@ -1,7 +1,7 @@
 import axios from 'axios';
 
 const http = axios.create({
-  baseURL: import.meta.env.VUE_APP_BASE_URL, // API baseURL as env variable
+  baseURL: '/api/v1/', // API baseURL as env variable
```
javiermtorres, line 4: It would be good to update the comment to reflect the new approach

javiermtorres (approved):

PR #630, opened 2025-01-15
Use output-field field in inference requests 

### LINKED ISSUE #622: [BUG]: Inference job output field is ignored
### Description

When a csv with an (empty or populated) ground_truth column is uploaded (regardless of the name) and we launch an inference job on it, setting the desired output_field (i.e. the field we want predictions to be added to), the service succeeds but promptly ignores output_field and still calls the field `predictions`.

### Reproduction

1. Upload file with empty gt
2. Repeat step 2 above, but with this new dataset.
3. See job succeed. 
4. Use /job/job_id/results/download to get the uri to a file like the one attached.

See:

* input file with empty gt column: [dialogsum_mini_empty_gt.csv](https://github.com/user-attachments/files/18354141/dialogsum_mini_empty_gt.csv)
* output file for the input with empty gt column: [results_empty_gt_requested_gt_field.json](https://github.com/user-attachments/files/18354150/results_empty_gt_requested_gt_field.json)

### Relevant log output

```shell
See above:

* input file with empty gt column
* output file for the input with empty gt column
```

### Expected behavior

Obtaining a file where, if the output_field requested is "model_prediction", for example, there is a new column named "model_prediction".

### System Info

* OS and LInux
* Current main: [34873f6](https://github.com/mozilla-ai/lumigator/commit/34873f680f7d944610c65b077d6a44736607246c)

javiermtorres:
# What's changing

The default inference config template was not passing the `output_field` to the inference job, so the ground truth column was not being generated. Now this key is passed, with a default value of `predictions` to match the inference job expectations.

The path for the job logs checks in the integration tests is also fixed.

Closes #622 

# How to test it

A new test integration case `test_upload_data_no_gt_launch_annotation` has been added.

N/A

# I already...

  - No changes in the DB needed


lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
```diff
@@ -78,7 +83,7 @@ def test_upload_data_launch_job(
     assert succeeded
 
     logs_evaluation_job_response = local_client.get(
-        f"/health/jobs/{create_evaluation_job_response_model.id}/logs"
+        f"/jobs/{create_evaluation_job_response_model.id}/logs"
```
ividal, line 86: 🤦‍♀️ How was this not failing earlier?
/me worries

lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
```diff
@@ -123,13 +128,89 @@ def test_upload_data_launch_job(
...
+    }
+
+    annotation_payload = {
+        "name": "test_annotate",
+        "description": "Test run for Huggingface model",
+        "dataset": str(created_dataset.id),
+        "max_samples": 2,
+        "task": "summarization",
+    }
+    create_inference_job_response = local_client.post(
```
ividal, line 167: Technically create_annotate(...). I know under the hood it's inference, but it's just to mentally keep track of what schema we should be following. 

For example, if we set the model here, it should be rejected. If we set it in a real, full inference job, it should go through.

lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
javiermtorres: Because there's no failing call :-/ It will say "404, {}" and everything is fine :)

lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
javiermtorres: Sorry, it is changed to inference in further vars :( Fixing the name 👍 

> For example, if we set the model here, it should be rejected. If we set it in a real, full inference job, it should go through.

This is the basis for a new unit test....

ividal (approved):
Thanks 🙏 !
Pre-approved and just a small comment on naming.

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -201,6 +201,9 @@ def create_job(
             name=request.name, description=request.description, experiment_id=experiment_id
         )
 
+        if isinstance(request, JobInferenceCreate) and not request.output_field:
```
aittalam, line 204: if the request is of type `JobInferenceCreate` it should always have an output field: https://github.com/mozilla-ai/lumigator/blob/906ed4531d5cfe9d1d554694e957b33d0526ed58/lumigator/python/mzai/schemas/lumigator_schemas/jobs.py#L81

lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
```diff
@@ -123,13 +128,89 @@ def test_upload_data_launch_job(
...
+        get_job_response = local_client.get(f"/jobs/{create_annotation_job_response_model.id}")
+        assert get_job_response.status_code == 200
+        get_job_response_model = JobResponse.model_validate(get_job_response.json())
+        if get_job_response_model.status == JobStatus.SUCCEEDED.value:
+            succeeded = True
+            break
+        if get_job_response_model.status == JobStatus.FAILED.value:
+            succeeded = False
+            break
+        time.sleep(10)
```
aittalam, line 187: We'll now have the same for loop replicated 4 times in this file. Could you please factor out the common code into a function and call that instead?
Also, 200*10 = 2000 seconds for each of these tests is definitely more than anyone is willing to wait. We should cap this to a reasonable amount of time and fail the test otherwise, warning that it failed due to the time cap so we can improve the way we are testing.

lumigator/python/mzai/backend/backend/services/jobs.py
javiermtorres: AFAICT both `None`s and `""`s are (quite conveniently) falses:

```python
>>> somestr = None
>>> if somestr:
...   print("yay")
... else:
...   print("nope")
... 
nope
>>> somestr = ""
>>> if somestr:
...   print("yay")
... else:
...   print("nope")
... 
nope
>>> somestr = "blurb"
>>> if somestr:
...   print("yay")
... else:
...   print("nope")
... 
yay
``` 

If you can think of other invalid values we can check those explicitly.

lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
javiermtorres: Yes, now that it's working and we have a clear picture of how the tests should go, we can refactor the integration tests, also reducing the waiting time. Maybe we can make it a separate PR, but if you prefer I'll include it in this one.

Is there some specific annoying corner case related to this waiting time that you'd like to be addressed?

lumigator/python/mzai/backend/backend/services/jobs.py
aittalam: Whoops sorry, just gave for granted that "" was not false, should have checked 😅

lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
aittalam: I think it's ok to have an issue for that! I have opened [this](https://github.com/mozilla-ai/lumigator/issues/643) so you are not blocked on this. Feel free to merge after SDK tests are addressed!

lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
javiermtorres: Ok, clarification. There's no need for a new unit test since the `JobAnnotationCreate` model does not allow a `model` field anyway (it would be dropped). The var names have been updated.

Issue #572, opened 2025-01-07
Add  get experiment results + its tests


aittalam:
Implements the equivalent of /jobs/{id}/result/download for experiments. This is done by taking an experiment_id as input and looking in the jobs table for the (two) jobs matching that id. To be compatible with the results we currently provide to the UI, we need to build a JSON that holds information from both (metrics from eval and inference time from inference)

veekaybee:
Current method call: https://github.com/mozilla-ai/lumigator/blob/fde73e2084d5f00bca560eea675c55ee00e7c610/lumigator/python/mzai/backend/backend/services/jobs.py#L293

veekaybee:
High-level arch: <img width="1592" alt="Image" src="https://github.com/user-attachments/assets/62160ea2-05fe-4489-87d9-3683069407b0" />

veekaybee:
https://github.com/mozilla-ai/lumigator/compare/experiment_results?expand=1

PR #632, opened 2025-01-15
Implement job_result_download for experiment service

veekaybee:
# What's changing

We currently download results per job. We'd like to implement all data results downloaded per experiment. 

1. We look in the jobs table for all jobs that match a specific experiment id (inference + eval) and 
2. return the results as JSON. 

image

In order to make these changes, we need to change the jobs, experiments service, and related API calls. 

Provide a clear and concise description of the content changes you're proposing. List all the changes you are making to the content.

+ Changing jobs service
+ Changing experiment service
+ Changing experiment service API calls

See https://github.com/mozilla-ai/lumigator/issues/572

# How to test it

Steps to test the changes:
 
1. Upload dialogsum and run experiment: 
```
#!/bin/bash
if [ "$#" -gt 0 ]; then
    DATA_CSV_PATH="$1"
else
    DATA_CSV_PATH="$HOME/lumigator-datasets/dialogsum/dialogsum.csv"
fi

fi

DATASET_ID=$(curl -s $BACKEND_URL/api/v1/datasets/ \
  -F 'format=job' | jq -r '.id')

EVAL_NAME="test_experiment_bart"
EVAL_DESC="Test experiment (inference + eval) with BART"
EVAL_MODEL="hf://facebook/bart-large-cnn"
EVAL_MAX_SAMPLES="10"

JSON_STRING=$(jq -n \
                --arg name "$EVAL_NAME" \
                --arg desc "$EVAL_DESC" \
                --arg model "$EVAL_MODEL" \
                --arg dataset_id "$DATASET_ID" \
                '{name: $name, description: $desc, model: $model, dataset: $dataset_id, max_samples: $max_samples}')

echo Starting new experiment...

echo $JSON_STRING

curl -s $BACKEND_URL/api/v1/experiments_new/ \
  -d "$JSON_STRING"
```

3. Once experiment runs, hit the new experiments endpoint to get results: 

```
curl -X 'GET' \
  'http://localhost:8000/api/v1/experiments_new/{yourid}/result/download' \
  -H 'accept: application/json'
```
You should get URLs for two files: 

```
{
  "id": "65dfc643-b35a-4b0b-8491-3089d2fab944",
  "download_url": "http://localhost:4566/lumigator-storage/jobs/results/test_experiment_bart-evaluation/65dfc643-b35a-4b0b-8491-3089d2fab944/results.json?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=test%2F20250120%2Fus-east-2%2Fs3%2Faws4_request&X-Amz-Date=20250120T211805Z&X-Amz-Expires=3600&X-Amz-SignedHeaders=host&X-Amz-Signature=32b8a08273e6406534ffd7887813006f556b950c60270732a2cbe136bd1cdfa9"
}
```

# I already...

- [X ] Tested the changes in a working environment to ensure they work as expected
- [ ] Added some tests for any new functionality - would like to address in a follow-up PR

### DIFF
--- a/lumigator/python/mzai/backend/backend/repositories/jobs.py
+++ b/lumigator/python/mzai/backend/backend/repositories/jobs.py
@@ -1,9 +1,7 @@
 from uuid import UUID
 
-from sqlalchemy import desc
 from sqlalchemy.orm import Session
 
-from backend.records.experiments import ExperimentRecord
 from backend.records.jobs import JobRecord, JobResultRecord
 from backend.repositories.base import BaseRepository
 
@@ -22,12 +20,3 @@ class JobResultRepository(BaseRepository[JobResultRecord]):
 
     def get_by_job_id(self, job_id: UUID) -> JobResultRecord | None:
         return self.session.query(JobResultRecord).where(JobResultRecord.job_id == job_id).first()
-
-    def get_jobs_by_experiment_id(self, experiment_id: UUID) -> list[ExperimentRecord]:
-        return (
-            self.session.query(ExperimentRecord)
-            .where(ExperimentRecord.id == experiment_id)
-            .order_by(desc(ExperimentRecord.created_at))
-            .limit(2)
-            .all()
-        )
--- a/lumigator/python/mzai/backend/backend/repositories/jobs.py
+++ b/lumigator/python/mzai/backend/backend/repositories/jobs.py
@@ -10,7 +10,7 @@ class JobRepository(BaseRepository[JobRecord]):
     def __init__(self, session: Session):
         super().__init__(JobRecord, session)
 
-    def get_by_experiment_id(self, experiment_id: UUID) -> list[JobRecord]:
+    def get_by_experiment_id(self, experiment_id: UUID) -> list[JobRecord] | None:
         return self.session.query(JobRecord).where(JobRecord.experiment_id == experiment_id).all()
 
 
--- a/lumigator/python/mzai/backend/backend/repositories/jobs.py
+++ b/lumigator/python/mzai/backend/backend/repositories/jobs.py
@@ -1,7 +1,9 @@
 from uuid import UUID
 
+from sqlalchemy import desc
 from sqlalchemy.orm import Session
 
+from backend.records.experiments import ExperimentRecord
 from backend.records.jobs import JobRecord, JobResultRecord
 from backend.repositories.base import BaseRepository
 
@@ -21,8 +23,6 @@ class JobResultRepository(BaseRepository[JobResultRecord]):
     def get_by_job_id(self, job_id: UUID) -> JobResultRecord | None:
         return self.session.query(JobResultRecord).where(JobResultRecord.job_id == job_id).first()
 
-
-
     def get_jobs_by_experiment_id(self, experiment_id: UUID) -> list[ExperimentRecord]:
         return (
             self.session.query(ExperimentRecord)
@@ -31,5 +31,3 @@ class JobResultRepository(BaseRepository[JobResultRecord]):
             .limit(2)
             .all()
         )
-
-
--- a/lumigator/python/mzai/backend/backend/repositories/jobs.py
+++ b/lumigator/python/mzai/backend/backend/repositories/jobs.py
@@ -1,5 +1,6 @@
 from uuid import UUID
 
+from sqlalchemy import desc
 from sqlalchemy.orm import Session
 
 from backend.records.jobs import JobRecord, JobResultRecord
@@ -20,3 +21,5 @@ class JobResultRepository(BaseRepository[JobResultRecord]):
 
     def get_by_job_id(self, job_id: UUID) -> JobResultRecord | None:
         return self.session.query(JobResultRecord).where(JobResultRecord.job_id == job_id).first()
+
+
--- a/lumigator/python/mzai/backend/backend/repositories/jobs.py
+++ b/lumigator/python/mzai/backend/backend/repositories/jobs.py
@@ -3,6 +3,7 @@ from uuid import UUID
 from sqlalchemy import desc
 from sqlalchemy.orm import Session
 
+from backend.records.experiments import ExperimentRecord
 from backend.records.jobs import JobRecord, JobResultRecord
 from backend.repositories.base import BaseRepository
 
@@ -22,4 +23,13 @@ class JobResultRepository(BaseRepository[JobResultRecord]):
     def get_by_job_id(self, job_id: UUID) -> JobResultRecord | None:
         return self.session.query(JobResultRecord).where(JobResultRecord.job_id == job_id).first()
 
+
+    def get_jobs_by_experiment_id(self, experiment_id: UUID) -> list[ExperimentRecord]:
+        return (
+            self.session.query(ExperimentRecord)
+            .where(ExperimentRecord.id == experiment_id)
+            .order_by(desc(ExperimentRecord.created_at))
+            .limit(2)
+            .all()
+        )
 
--- a/lumigator/python/mzai/backend/backend/repositories/jobs.py
+++ b/lumigator/python/mzai/backend/backend/repositories/jobs.py
@@ -1,9 +1,7 @@
 from uuid import UUID
 
-from sqlalchemy import desc
 from sqlalchemy.orm import Session
 
-from backend.records.experiments import ExperimentRecord
 from backend.records.jobs import JobRecord, JobResultRecord
 from backend.repositories.base import BaseRepository
 
@@ -24,6 +22,7 @@ class JobResultRepository(BaseRepository[JobResultRecord]):
         return self.session.query(JobResultRecord).where(JobResultRecord.job_id == job_id).first()
 
 
+
     def get_jobs_by_experiment_id(self, experiment_id: UUID) -> list[ExperimentRecord]:
         return (
             self.session.query(ExperimentRecord)
@@ -32,4 +31,5 @@ class JobResultRepository(BaseRepository[JobResultRecord]):
             .limit(2)
             .all()
         )
+
 
--- a/lumigator/python/mzai/backend/backend/services/experiments.py
+++ b/lumigator/python/mzai/backend/backend/services/experiments.py
@@ -1,8 +1,18 @@
 from uuid import UUID
 
 import loguru
+<<<<<<< HEAD
 from fastapi import BackgroundTasks
 from lumigator_schemas.experiments import ExperimentCreate, ExperimentResponse
+=======
+from fastapi import BackgroundTasks, UploadFile
+from lumigator_schemas.datasets import DatasetFormat
+from lumigator_schemas.experiments import (
+    ExperimentCreate,
+    ExperimentResponse,
+    ExperimentResultDownloadResponse,
+)
+>>>>>>> 4fe14cd (add experiment methods)
 from lumigator_schemas.extras import ListingResponse
 from lumigator_schemas.jobs import (
     JobEvalLiteCreate,
@@ -112,6 +122,7 @@ class ExperimentService:
 
         return ExperimentResponse.model_validate(experiment_record)
 
+<<<<<<< HEAD
     # TODO Move this into a "composite job" impl
     def get_experiment(self, experiment_id: UUID) -> ExperimentResponse:
         """Gets an experiment by ID.
@@ -138,6 +149,49 @@ class ExperimentService:
             record = self._experiment_repo.update(experiment_id, status=JobStatus.SUCCEEDED)
 
         return ExperimentResponse.model_validate(record)
+=======
+    def _get_experiment_jobs(self, experiment_id: UUID):
+        records = self._job_service.job_repo.get_jobs_by_experiment_id(experiment_id)
+        return records
+
+    def get_experiment_result_download(
+        self, experiment_id: UUID
+    ) -> ExperimentResultDownloadResponse:
+        """Return experiment results file URL for downloading."""
+        s3 = S3FileSystem()
+        # get jobs matching this experiment
+        # have a query returning a list of (two) jobs, one inference and one eval,
+        # matching the current experiment id. Note that each job has its own type baked in
+        # (per https://github.com/mozilla-ai/lumigator/pull/576)
+        jobs = self._get_experiment_jobs(experiment_id)
+
+        # iterate on jobs and depending on which job this is, import selected fields
+        results = {}
+
+        for job in jobs:
+            # Get the dataset from the S3 bucket
+            result_key = self._job_service._get_results_s3_key(job.id)
+            with s3.open(f"{settings.S3_BUCKET}/{result_key}", "r") as f:
+                job_results = json.loads(f.read())
+
+            # we just merge the two dictionaries for now
+            results.update(job_results)
+
+        loguru.logger.error(results)
+
+        # Generate presigned download URL for the object
+        result_key = self._get_results_s3_key(experiment_id)
+        download_url = self.data_service.s3_client.generate_presigned_url(
+            "get_object",
+            Params={
+                "Bucket": settings.S3_BUCKET,
+                "Key": result_key,
+            },
+            ExpiresIn=settings.S3_URL_EXPIRATION,
+        )
+
+        return ExperimentResultDownloadResponse(id=experiment_id, download_url=download_url)
+>>>>>>> 4fe14cd (add experiment methods)
 
     def list_experiments(
         self, skip: int = 0, limit: int = 100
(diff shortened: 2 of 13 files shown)


lumigator/python/mzai/schemas/lumigator_schemas/experiments.py
```diff
@@ -22,7 +20,6 @@ class ExperimentResponse(BaseModel, from_attributes=True):
     id: UUID
     name: str
     description: str
-    status: JobStatus
```
veekaybee, line 25: this isn't being used at the experiment level

lumigator/python/mzai/schemas/lumigator_schemas/jobs.py
```diff
@@ -73,6 +73,7 @@ class JobEvalLiteCreate(BaseModel):
     name: str
     description: str = ""
     model: str
+    model_url: str | None = None
```
veekaybee, line 76: keeping consistent with other eval job

lumigator/python/mzai/schemas/lumigator_schemas/datasets.py
```diff
@@ -21,6 +21,6 @@ class DatasetResponse(BaseModel, from_attributes=True):
     size: int
     ground_truth: bool
     run_id: UUID | None
-    generated: bool
+    generated: bool | None
```
veekaybee, line 24: this was returning an error when we didn't fill it

lumigator/python/mzai/backend/backend/services/experiments.py
```diff
@@ -196,6 +200,48 @@ def create_experiment(
...
+        # iterate on jobs and depending on which job this is, import selected fields
+        results = {}
+
+        for job in jobs:
+            # Get the dataset from the S3 bucket
+            result_key = self._job_service._get_results_s3_key(job.id)
+            with s3.open(f"{settings.S3_BUCKET}/{result_key}", "r") as f:
+                job_results = json.loads(f.read())
+
+            # we just merge the two dictionaries for now
```
javiermtorres, line 227: What about a list of job responses? Merging the dicts means we and the user needs to keep track of what info goes into which dicts (which is not modelled currently), and some may be overwritten accidentally.

lumigator/python/mzai/backend/backend/services/experiments.py
```diff
@@ -196,6 +200,48 @@ def create_experiment(
...
+            results.update(job_results)
+
+        loguru.logger.error(results)
+
+        # Generate presigned download URL for the object
+        result_key = self._get_results_s3_key(experiment_id)
```
javiermtorres, line 233: Is the results dict being actually store in S3?

lumigator/python/mzai/schemas/lumigator_schemas/experiments.py
javiermtorres: It'll be introduced shortly, btw.

lumigator/python/mzai/schemas/lumigator_schemas/jobs.py
javiermtorres: We can try to get a deeper look at that issue, if you feel it should be left out.

lumigator/python/mzai/backend/backend/repositories/jobs.py
```diff
@@ -16,8 +17,7 @@ def __init__(self, session: Session):
...
     def get_by_job_id(self, job_id: UUID) -> JobResultRecord | None:
-        return (
-            self.session.query(JobResultRecord)
-            .where(JobResultRecord.job_id == job_id)
-            .first()
-        )
+        return self.session.query(JobResultRecord).where(JobResultRecord.job_id == job_id).first()
+
+    def get_jobs_by_experiment_id(self, experiment_id: UUID) -> list[JobRecord]:
+        return self.session.query(JobRecord).order_by(desc(JobRecord.created_at)).limit(2).all()
```
javiermtorres, line 23: I'd remove the order and the limit. Is there any change to put these constraints?

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -112,11 +112,11 @@ def _get_config_template(self, job_type: str, model_name: str) -> str:
 
         return config_template
 
-    def _set_model_type(self, request: BaseModel) -> str:
+    def _set_model_type(self, request) -> str:
         """Sets model URL based on protocol address"""
-        if request.model.startswith("oai://"):
+        if request.model_url.startswith("oai://"):
```
dpoulopoulos, line 117: You should now pass a `model_url` argument even if you don't need it (i.e., in the case of self-hosted models). Why don't we first check if `model_url` is set and if not return `None`. Currently, this breaks the "create new experiment" flow, since the second job that the experiment tries to create fails with:

```
backend-1      |   File "/mzai/lumigator/python/mzai/backend/backend/services/jobs.py", line 117, in _set_model_type
backend-1      |     if request.model_url.startswith("oai://"):
backend-1      |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
backend-1      | AttributeError: 'NoneType' object has no attribute 'startswith'
```

lumigator/python/mzai/backend/backend/repositories/jobs.py
dpoulopoulos: Why don't we use the `experiment_id` here and instead limit the result set to the first two results?

lumigator/python/mzai/backend/backend/services/experiments.py
```diff
@@ -196,6 +200,48 @@ def create_experiment(
 
         return ExperimentResponse.model_validate(experiment_record)
 
+    def _get_experiment_jobs(self, experiment_id: UUID):
+        records = self._job_service.job_repo.get_jobs_by_experiment_id(experiment_id)
```
dpoulopoulos, line 204: You've added the `get_jobs_by_experiment_id` method in the `JobResultRepository` class and you're trying to invoke it using a `JobRepository` instance. This doesn't work:

```
AttributeError: 'JobRepository' object has no attribute 'get_jobs_by_experiment_id'
```

I guess you'd want to move the `get_jobs_by_experiment_id` method in the `JobRepository` class.

dpoulopoulos (changes requested):

lumigator/python/mzai/backend/backend/repositories/jobs.py
aittalam: This is a mockup we set up while waiting for the grouping by experiment_id to be present. It allowed Vicki to get inference+eval job from the latest call to experiments_new without having job->experiment matching in place yet. It will be changed with the current method made available to get all experiments for a given job (which I think [has not been merged yet](https://github.com/mozilla-ai/lumigator/pull/604/files#diff-efa4d733ff129c97fe9750321306aa06551dd36851963370315d200d49e14ce0R13), is that correct?)

lumigator/python/mzai/backend/backend/services/experiments.py
aittalam: I can provide the rationale behind this.

This is by no way a general method we want to implement for composite jobs, but serves the purpose of merging results from inference+evaluation, whose outputs we know precisely. We will likely not need this anymore if/when we start pulling aggregated results from child runs in e.g. mlflow, and we would not need to do this if we did not want to be 100% compatible with the current API and the UI which hits "experiments" when it needs information about jobs.
All of this to say this is ad-hoc and temporary :-)

The reason why we merge these is that the UI expects from the API the outputs of an "experiment" (from the UI pov) with original samples, ground_truth, predictions, and metrics. We do not force evaluation to have original samples as an input though, so we need to get them from the inference job. The union of the jobs' outputs is our output. We don't mind about overwriting keys rn as if we have the same key in the two dictionaries the contents are the same too.

lumigator/python/mzai/backend/backend/services/experiments.py
aittalam: yes, both for inference and eval jobs

lumigator/python/mzai/backend/backend/services/jobs.py
aittalam: I think this was intended to be `model`, which always has a value (it is the URI we provide as input). The url, instead, is optional for non-api models.

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -112,11 +112,11 @@ def _get_config_template(self, job_type: str, model_name: str) -> str:
...
             model_url = settings.OAI_API_URL
-        elif request.model.startswith("mistral://"):
+        elif request.model_url.startswith("mistral://"):
```
aittalam, line 119:
```suggestion
        elif request.model.startswith("mistral://"):
```

lumigator/python/mzai/backend/backend/services/jobs.py
aittalam, line 117:
```suggestion
        if request.model.startswith("oai://"):
```

lumigator/python/mzai/schemas/lumigator_schemas/jobs.py
aittalam: The evaluator job takes models in because it runs inference with those models. The eval_lite job has a `model` parameter just to record the model name and it does not use `model_url` at all. What is the caller that passes this parameter to eval_lite? The ExperimentService passes [this dictionary](https://github.com/mozilla-ai/lumigator/blob/ced51dac2c7b6aa4bd2e6a5d75c764db221906b9/lumigator/python/mzai/backend/backend/services/experiments.py#L134).

lumigator/python/mzai/backend/backend/services/jobs.py
veekaybee: This is the stack trace I was getting without `model_url` on an eval_lite call: 

```
backend-1      |   File "/mzai/lumigator/python/mzai/backend/backend/api/routes/jobs.py", line 95, in create_evaluation_lite_job
backend-1      |     job_response = service.create_job(job_create_request)
backend-1      |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
backend-1      |   File "/mzai/lumigator/python/mzai/backend/backend/services/jobs.py", line 209, in create_job
backend-1      |     config_params = self._get_job_params(job_type, record, request)
backend-1      |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
backend-1      |   File "/mzai/lumigator/python/mzai/backend/backend/services/jobs.py", line 154, in _get_job_params
backend-1      |     "model_uri": request.model,
backend-1      |                  ^^^^^^^^^^^^^
backend-1      |   File "/mzai/lumigator/python/mzai/backend/.venv/lib/python3.11/site-packages/pydantic/main.py", line 856, in __getattr__
backend-1      |     raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
backend-1      | AttributeError: 'JobEvalLiteCreate' object has no attribute 'model'
```
It's clearly not correct because it does have that attribute: https://github.com/mozilla-ai/lumigator/blob/55eb2d9c31f707aa0e077af01d215c1d77cd63e7/lumigator/python/mzai/schemas/lumigator_schemas/jobs.py#L72 so I need to dig a bit deeper here

aittalam: My hypothesis is that the model URI (e.g. `mistral://...`) might be passed as `model_url` field instead of `model`. The URI identifies the model, the URL provides e.g. an IP for locally hosted models (e.g. localhost for llamafiles)

lumigator/python/mzai/schemas/lumigator_schemas/jobs.py
ividal: Are we sure it should stay consistent between inference and evaluation jobs? I understand the awkward situation while the `evaluator` combo is in the code, but we are phasing out at the same time `evaluator` and (by extension) the current `/experiments` at the same time, so I'm not sure all jobs need a `model_url`, no?

lumigator/python/mzai/backend/backend/services/experiments.py
aittalam: Agree: if we are free from the need to be compatible with the current API, I think we can come up with something better than this.
For instance, there's no need to move all this data around jobs: we could limit this to the bare minimum inputs and outputs, and then let the UI pick whatever they need to visualize at a given moment (e.g. they won't likely need a GB-large dataset if they just want to show aggregated metrics)

veekaybee: Yep, my suggestion here was to change this piece of the UI https://github.com/mozilla-ai/lumigator/blob/fde73e2084d5f00bca560eea675c55ee00e7c610/lumigator/frontend/src/services/experiments/experimentService.js#L56

lumigator/python/mzai/backend/backend/services/experiments.py
```diff
@@ -196,6 +200,48 @@ def create_experiment(
...
+        return records
+
+    def get_experiment_result_download(
```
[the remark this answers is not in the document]
veekaybee: I see these as two different routes in that case

lumigator/python/mzai/backend/backend/services/jobs.py
javiermtorres: Could it be that the request is passed typed as a BaseModel??? I don't know if Pydantic will intercept some of the method/attr calls to enforce the constraints :-/

lumigator/python/mzai/backend/backend/services/experiments.py
veekaybee: I've opted to put this in the experiment repo since it's logically related to experiments and we don't instantiate a jobs repo in the experiment service - let me know what you think

lumigator/python/mzai/backend/backend/services/experiments.py
njbrake: I'm still wrapping my head around everything, but maybe this ties into https://github.com/mozilla-ai/lumigator/issues/678 a bit? 

I'm trying to work through at what point the backend should prepare the file that the frontent will use. This become especially relevant when we move to mlflow, because then all the metrics (rouge/bert/etc) are stored in a db instead of a json file. This means that then for the frontend to use the data, the backend will need to access mlflow to prepare a json file for the frontend to use. 

I assume we would want to do this as a part of the experiment workflow, and would not want to wait for the frontend to ask for the results to create the file.

veekaybee: IMO, in the longer-term: 

+ communication between the backend and front-end should happen via JSON only - the front-end shouldn't know about any files, and calls to files should deserialize them to JSON objects which are rendered in the UI
+ The front-end should have a generic container that, as in 678 contains "metrics" that all get rendered recursively instead of with specific elements
+ the front-end shouldn't know about S3 or URLs

In the short-term, for this PR, we'd like to keep it tightly scoped which is why it's currently written this way

veekaybee:
Rebasing based on changes from https://github.com/mozilla-ai/lumigator/pull/657 and https://github.com/mozilla-ai/lumigator/pull/604

lumigator/python/mzai/backend/backend/repositories/jobs.py
veekaybee: still going off this assumption for now, unless it's changed?

lumigator/python/mzai/backend/backend/services/jobs.py
veekaybee: Did an investigation into this and fixed by passing correct model params

veekaybee:
@aittalam - ready for a re-review, I've incorporated all the changes since the last time this was reviewed and noted a place we might want to be careful

dpoulopoulos (approved):
Thanks!

lumigator/python/mzai/backend/backend/repositories/jobs.py
aittalam: We do have a real (ie non-mock) `get_by_experiment_id` method now in the jobs repository: https://github.com/mozilla-ai/lumigator/blob/572_experiment_results/lumigator/python/mzai/backend/backend/repositories/jobs.py#L13 , you can use that now and remove this one 🥳

javiermtorres (approved):

aittalam (approved):
TY Vicki! I think this is ready to merge after syncing with main.
Many thanks for addressing my comments 🙏

lumigator/python/mzai/schemas/lumigator_schemas/experiments.py
peteski22: @javiermtorres / @veekaybee does this mean [this doc comment](https://github.com/mozilla-ai/lumigator/blob/1c73e22cc815bf492017a24e1dedca8216f2cb88/lumigator/python/mzai/backend/backend/services/experiments.py#L123) doesn't need adjusting if `status` is coming back?

Issue #559, opened 2025-01-06
Add inference job tests to Makefile


aittalam:
Similarly to [eval_lite](https://github.com/mozilla-ai/lumigator/pull/523/files#r1901128003), we should call inference job unit tests from the Makefile so we have a single entry point to test all jobs.

PR #633, opened 2025-01-15
559 add inference job tests to makefile

aittalam:
# What's changing

Adds inference unit tests to the makefile, and fixes a couple of errors due to pydantic schema changes in eval + inference jobs.

Refs #559 

# How to test it

make test-jobs-inference
make test-jobs-evaluation

# I already...

- [x] Added some tests for any new functionality (added new test profiles to Makefile / fixed broken unit tests)


lumigator/python/mzai/jobs/evaluator_lite/tests/test_configs.py
```diff
@@ -11,7 +11,7 @@ def test_valid_config():
         name="my_eval_job",
         dataset=DatasetConfig(path="/path/to/dataset"),
         model=ModelConfig(path="/path/to/model"),
-        evaluation=EvaluationConfig(),
+        evaluation=EvaluationConfig(storage_path="/path/to/results"),
```
aittalam, line 14: `storage_path` is now compulsory

lumigator/python/mzai/jobs/inference/tests/data/config_full_hf.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "full_inference_config",
     "dataset": { "path": "s3://lumigator-storage/datasets/deaddead-dead-dead-dead-deaddeaddead/dataset_name.csv" },
-    "pipeline": { "model_uri": "hf://facebook/bart-large-cnn", "task": "summarization" },
+    "hf_pipeline": { "model_uri": "hf://facebook/bart-large-cnn", "task": "summarization" },
```
aittalam, line 4: typo was breaking the unit test

Makefile
```diff
@@ -137,15 +137,19 @@ else
 	$(call run_with_existing_containers, test-backend-integration-target)
 endif
 
-test-jobs-unit:
+test-jobs-evaluation:
 	cd lumigator/python/mzai/jobs/evaluator_lite; \
-	uv run pytest -o python_files="evaluator_lite/tests/test_*.py"
+    uv run --with pytest --with-requirements requirements.txt --isolated pytest
```
aittalam, line 142: This allows us to run in an isolated environment which just pulls up libraries from requirements.txt.
It means a lighter env for us but also avoids relying on libraries which are not in requirements.

pyproject.toml
```diff
@@ -4,11 +4,7 @@ version = "0.1.0"
 description = "Common deps to develop"
 requires-python = ">=3.11"
 dependencies = [
-    "huggingface-hub>=0.27.0",
```
aittalam, line 7: @veekaybee This libraries were added in https://github.com/mozilla-ai/lumigator/pull/523 for the eval job tests, I am removing them in this PR and adding an isolated env relying on requirements.txt only.

lumigator/python/mzai/backend/uv.lock
```diff
@@ -890,7 +913,7 @@ wheels = [
 
 [[package]]
 name = "lumigator-schemas"
-version = "0.1.0rc2"
+version = "0.0.2"
```
aittalam, line 916: @javiermtorres It seems like the schemas version has been downgraded since the last version of `uv.lock`, is this correct?

Makefile
```diff
@@ -137,15 +137,19 @@ else
...
+
+test-jobs-inference:
+	cd lumigator/python/mzai/jobs/inference; \
+	uv run --with pytest --with-requirements requirements.txt --isolated pytest
```
ividal, line 146: Is this to replicate what will happen on the ray cluster, which is why we need to separate the deps there from the "general" lumigator system deps?

Makefile
dpoulopoulos: Could you add this as a comment in the Makefile?

Makefile
```diff
@@ -137,15 +137,19 @@ else
...
 
 test-sdk-target: test-sdk-unit test-sdk-integration
 
-test-backend-target: test-backend-unit test-backend-integration test-jobs-unit
+test-backend-target: test-backend-unit test-backend-integration test-jobs-evaluation test-jobs-inference
```
dpoulopoulos, line 150: Should we add these as part of the backend target or create a new `test-jobs-target` target? We may have multiple jobs down the line, in which case the backend target will become inefficient. WDYT?

dpoulopoulos (commented):
Looks good to me! I just left a couple comments.

Makefile
```diff
@@ -137,15 +137,19 @@ else
 	$(call run_with_existing_containers, test-backend-integration-target)
 endif
 
-test-jobs-unit:
+test-jobs-evaluation:
```
veekaybee, line 140: are these unit or integration tests? i think these are both so would recommend including a qualifier so that we can run all unit and all integration separately

lumigator/python/mzai/jobs/evaluator_lite/tests/test_configs.py
veekaybee: I think we also need to change here https://github.com/mozilla-ai/lumigator/blob/39cc9384206f7dfbfe2eafb02250ccdb8860db04/lumigator/python/mzai/jobs/evaluator_lite/eval_config.py#L19

pyproject.toml
veekaybee: I learned about an even cooler trick you can do with uv now, which we probably don't want to replicate, but just for reference, you can include requirements inline to a script and uv will do everything from setting up env, to running tests, etc when you do `uv run myscript.py`

https://peps.python.org/pep-0723/

veekaybee (approved):
Approved after addressing review comments, thanks for this!

Makefile
aittalam: Will do!

pyproject.toml
aittalam: Wow! Adding to my TIL :-)

Makefile
aittalam: Exactly! This replicates the same env of the ray workers by just considering the deps listed in `requirements.txt`

Makefile
aittalam: Good point. As we only have unit for now (i.e. we mostly test job configs isolated from anything else) I'll just revert to test-jobs-unit

javiermtorres: AFAICT these tests will not require additional containers to run? Or will they need s3? In the latter case, I'd just add the S3 mocks in a conftest and put these tests as part of unit tests.

aittalam: I confirm they do not need S3 to run (you can just check https://github.com/mozilla-ai/lumigator/tree/main/lumigator/python/mzai/jobs/evaluator_lite/tests and https://github.com/mozilla-ai/lumigator/tree/main/lumigator/python/mzai/jobs/inference/tests)

Makefile
aittalam: Addressed in https://github.com/mozilla-ai/lumigator/pull/633/commits/824d86f6f5f0a6cb9c54d1cd78f58d65ef7d9e17

lumigator/python/mzai/backend/uv.lock
aittalam: Resolving after offline sync

Makefile
aittalam: Addresed in https://github.com/mozilla-ai/lumigator/pull/633/commits/824d86f6f5f0a6cb9c54d1cd78f58d65ef7d9e17

lumigator/python/mzai/jobs/evaluator_lite/tests/test_configs.py
aittalam: You mean to make it compulsory? Removing the default already does it, i.e. after removing that the tests started failing because `storage_path` was not provided

aittalam:
Hi all! I addressed comments and added a few changes:

- @dpoulopoulos I have modified the Makefile quite extensively, mostly moving different test targets around and adding comments so it would be easier for ppl to understand what happens. I took the liberty of removing the *-target expansion because it was giving me problems with recursion. The code seems a bit more readable to me rn but I'd be happy to have your blessing before committing 😇 

- @macaab26 as I refactored some test targets I had to update the [.github/workflows/lumigator_pipeline.yaml](https://github.com/mozilla-ai/lumigator/pull/633/files#diff-b2d113cb198175eef4f61a973cb31ed72b5a36d56b1b4b90f406f434761c34b6) file. Looks alright to me now (and tests pass) but I'd be happy to hear your thoughts about it 🙏

macaab26:
Hey @aittalam thanks for the heads up! Yep, looks good 😄

dpoulopoulos (approved):

PR #638, opened 2025-01-15
Check Minio health more often

njbrake:
# What's changing

Would it be ok if we check the health more often? My minio stack is normally up after ~7 seconds so this change would save me about 20 seconds every time I run local-up. I kept the same total duration by bumping up the retries.

# How to test it

Steps to test the changes:

1.`make local-up`

# I already...

- [ x] Tested the changes in a working environment to ensure they work as expected
- [x ] Checked if a (backend) DB migration step was required and included it if required

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -10,9 +10,9 @@ services:
       - 9001:9001
     healthcheck:
       test: ["CMD", "curl", "-f", "http://localhost:9000/minio/health/live"]
-      interval: 30s
+      interval: 5s
       timeout: 20s
-      retries: 3
+      retries: 18
     environment:
       - MINIO_ROOT_USER=minioadmin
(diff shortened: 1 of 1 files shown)


javiermtorres (approved):
LGTM

docker-compose.yaml
```diff
@@ -10,9 +10,9 @@ services:
       - 9001:9001
     healthcheck:
       test: ["CMD", "curl", "-f", "http://localhost:9000/minio/health/live"]
-      interval: 30s
+      interval: 5s
       timeout: 20s
-      retries: 3
+      retries: 18
```
njbrake, line 15: I would also be cool with updating this instead to 

```
      interval: 30s
      timeout: 20s
      retries: 3
      start_period: 10s
```

If you're not a fan of doing a 5s health check interval in perpetuity

PR #641, opened 2025-01-15
Bug fix: Assigned correct template to gpt-4o, removed gpt-3.5

aittalam:
# What's changing

Fixes missing gpt-4o template to be consistent with models offered in the UI

# How to test it

Run an inference / evaluator job passing `oai://gpt-4o` as model - it broke previously trying to load a HF model with that name, it now properly calls the API

# I already...

- [x] Added some tests for any new functionality (no new functionality, just a bug fix)
- [x] Updated the documentation (both comments in code and [product documentation](https://mozilla-ai.github.io/lumigator) under `/docs`) (no need as that model was already in documentation, just not captured in templates)


dpoulopoulos (approved):
Looks good to me! Thanks for this @aittalam!

PR #647, opened 2025-01-16
Docs: Mention Lumigator UI usage in README + minor improvements 

### LINKED ISSUE #634: [DOC] No mention of how to use Lumigator UI in README + minor README enhancements
Lumigator UI for uploading datasets and running experiments can be accessed via http://localhost:80/ 
However, this information is missing from the README. 

At the minimum, we could mention it in a few lines in the README.
Even better could be to add a section to lumigator docs with screenshots and step by step instructions (with the risk that the frontend/UI could be quite different couple of months from now and the screenshots would need to be constantly updated).

---
Other minor improvements to README. Mostly sentence rephrasing.

- [x] _Lumigator supports the evaluation of summarization tasks using sequence-to-sequence models such as BART and BERT_ → Lumigator supports the evaluation of summarization tasks using sequence-to-sequence models such as BART ~~and BERT~~ (BERT is inherently an encoder only model)
- [x] _Yet, in spite of the wealth of metrics available, there’s still no standard way to compare these models._ → However, one may find it challenging to compare these models for their specific domain and usecase. (Since researchers use [HELM](https://crfm.stanford.edu/helm/classic/latest/#/leaderboard) and other leaderboards as benchmarking 'standard')
- [x] _Without a clear ~~single~~ method for comparing models_ (since even with Lumigator we believe that users would use metrics suitable to their scenario for comparing, rather than a single universal method)
- [x] _[!NOTE]_ If you'd like to evaluate against (rendering issue)
- [x] _Similarly, one could provide a different `RAY_HEAD_NODE_HOST` to move compute to a remote ray cluster, and so on. See [here](https://github.com/mozilla-ai/lumigator/blob/7be2518ec8c6bc59ab8463fc7c39aad078bbb386/docker-compose.external.yaml) for an example of how to do this_ → that link does not exist on **main** branch
- [x] _The system Python; no version manager, such as uv, should be active_ → The system Python (version managers such as uv should be deactivated)

HareeshBahuleyan:
# What's changing
* Briefly introduce Lumigator UI usage in README.
* Rephrase several sentences in README for clarity
* Changes to [docs](https://mozilla-ai.github.io/lumigator/index.html): Clarify for users model is used in the automatic annotation (i.e., BART for summarization)

Addresses #634

# I already...

- [NA] Added some tests for any new functionality


HareeshBahuleyan:
As described in #634, I could further create a section in the main docs for step-by-step instructions with screenshots on how to use the Lumigator UI. Would that be something we would like to have?

README.md
```diff
@@ -23,12 +22,11 @@ To learn more about Lumigator's features and capabilities, see the
 As more organizations turn to AI for solutions, they face the challenge of selecting the best model
 from an ever-growing list of options. The AI landscape is evolving rapidly, with [twice as many new
 models released in 2023 compared to the previous year](https://hai.stanford.edu/research/ai-index-report).
-Yet, in spite of the wealth of metrics available, there’s still no standard way to compare these
-models.
+However, one may find it challenging to compare models for their specific domain and usecase.
```
ividal, line 25: Since you [mentioned it](https://github.com/mozilla-ai/lumigator/issues/634), what do you think of acknowledging research-oriented benchmarks?

```suggestion
However, in spite of existing [benchmarks and leaderboards](https://crfm.stanford.edu/helm/classic/latest/#/leaderboard) for some scenarios, one may find it challenging to compare models for their specific domain and use case.
```

ividal (approved):
Thanks @HareeshBahuleyan ! Minor suggestions, but already lgtm.

HareeshBahuleyan:
> As described in #634, I could further create a section in the main docs with step-by-step instructions and screenshots on how to use the Lumigator UI. Would that be something we would like to have?

@ividal Would you like this to be added to the docs in this PR or maybe keep it for later (when the UI is more mature)?

README.md
```diff
@@ -94,16 +92,18 @@ To verify that Lumigator is running, open a web browser and navigate to
...
-for an example of how to do this, and see the
+`RAY_HEAD_NODE_HOST` to move compute to a remote ray cluster, and so on. See the
 [operational guides](https://mozilla-ai.github.io/lumigator/operations-guide/kubernetes.html) in the
 documentation for more deployment options.
 
 Now that Lumigator is running, you can start using it. The platform provides a REST API that allows
 you to interact with the system. Run the [example notebook](/notebooks/walkthrough.ipynb) for a
 quick walkthrough.
 
+### Lumigator UI
```
dpoulopoulos, line 103: I suggest to not start a new section here. Just continue the paragraph without a heading.

HareeshBahuleyan: My thought was that having it as a heading catches the eye of a reader who is skimming through the README - by way of which they become aware that a UI exists. 
If we have it as just a paragraph, probably there is a higher chance of missing this detail. WDYT?

HareeshBahuleyan:
> As described in #634, I could further create a section in the main docs with step-by-step instructions and screenshots on how to use the Lumigator UI. Would that be something we would like to have?

Based on discussions with @ividal, I have opened a separate issue https://github.com/mozilla-ai/lumigator/issues/664 to track this. Since the UI changes are pending, we tackle the docs update with screenshots in a separate PR.

README.md
dpoulopoulos: IMO if you're adding a `###` section it means that you have more than one `###` sections under the same `##` heading. Think of it like this: Say you have chapter `1` in a book, and then sub-chapters `1.1` and `1.2`. Would you start a new `1.1.1` sub-heading if there was no `1.1.2`? It seems a bit redundant.

Also, you're now putting the "Lumigator UI" heading on the same level as "Terminate Lumigator session". Not sure if this is something you'd want.

In any case, that's a personal preference and it's very minor. If you think that it looks okay the way it is I'm fine with it! :)

dpoulopoulos (approved):

README.md
HareeshBahuleyan: > Also, you're now putting the "Lumigator UI" heading on the same level as "Terminate Lumigator session". 

Right, this I agree is not ideal.
What we can do is once we have a dedicated page in the docs with #664, we can get rid of these subsections (since Lumigator UI has a better visibility through the docs page by then).

PR #650, opened 2025-01-16
Change from EXPERIMENT to JOB to match schema

chainlink:
# What's changing

This brings into line the migration for datasets and the schema defined for datasets.

# How to test it

- [x] upload a dataset to lumigator

This error was not surfaced in sqlite, but did come up when running lumigator with postgres

# I already...

- [x] Checked if a (backend) DB migration step was required and included it if required (lol)


javiermtorres (approved):
LGTM

lumigator/python/mzai/backend/backend/alembic/versions/e75fa022c781_create_db.py
```diff
@@ -5,69 +5,73 @@
...
 """
+
 from collections.abc import Sequence
 
 import sqlalchemy as sa
 from alembic import op
 
 # revision identifiers, used by Alembic.
-revision: str = 'e75fa022c781' # pragma: allowlist secret
+revision: str = "e75fa022c781"  # pragma: allowlist secret
```
agpituk, line 15: thanks, it was bugging me that the only migration with single quotes was the first one :D

lumigator/python/mzai/backend/backend/alembic/versions/e75fa022c781_create_db.py
```diff
@@ -5,69 +5,73 @@
...
-        'created_at',
-        sa.DateTime(timezone=True),
-        server_default=sa.text('(CURRENT_TIMESTAMP)'),
-        nullable=False
-    ),
-    sa.PrimaryKeyConstraint('id')
+    op.create_table(
+        "datasets",
+        sa.Column("filename", sa.String(), nullable=False),
+        sa.Column("format", sa.Enum("JOB", name="datasetformat"), nullable=False),
```
agpituk, line 26: I haven't checked but, Are we certain we also don't need "EXPERIMENT" as well? I thought this change would be to add JOB, not to replace Experiment

chainlink: I'm going by the enum defined in schemas which only has `JOB` https://github.com/mozilla-ai/lumigator/blob/main/lumigator/python/mzai/schemas/lumigator_schemas/datasets.py#L9

agpituk (approved):

PR #656, opened 2025-01-16
Fix docker clean

aittalam:
# What's changing

`make clean-docker-all` breaks when no containers are running because the first command `clean-docker-containers` returns an error if the list of containers is empty.

This PR fixes that by using `xargs` to pass the list (when not empty) to `docker rm` and `docker rmi`.

Closes https://github.com/mozilla-ai/lumigator/issues/660

# How to test it

run `make clean-all` or `make clean-docker-all`. This should remove all containers + images if present, without breaking if there is none

I am not sure why we'd need both `make clean-all` and `make clean-docker-all`. The semantics I had in mind was "clean-all should call clean-docker-all plus something else" but that does not happen (actually this is a subset of `clean-docker-all`). WDYT? Should we remove `clean-all` or is there a good reason to keep it (perhaps with a different name)?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,10 +104,10 @@ clean-docker-buildcache:
 	docker builder prune --all -f
 
 clean-docker-containers:
-	docker rm -vf $$(docker container ls -aq)
+	docker container ls -aq | xargs -n1 docker rm -vf
 
 clean-docker-images:
-	docker rmi -f $$(docker image ls -aq)
+	docker image ls -aq | xargs -n1 docker rmi -f
 
 clean-docker-all: clean-docker-containers clean-docker-buildcache clean-docker-images
 
--- a/Makefile
+++ b/Makefile
@@ -104,10 +104,10 @@ clean-docker-buildcache:
 	docker builder prune --all -f
 
 clean-docker-containers:
-	docker container ls -aq | xargs -n1 docker rm -vf
+	docker container prune
 
 clean-docker-images:
-	docker image ls -aq | xargs -n1 docker rmi -f
+	docker image prune -a
 
 clean-docker-all: clean-docker-containers clean-docker-buildcache clean-docker-images
 
--- a/Makefile
+++ b/Makefile
@@ -106,8 +106,10 @@ clean-docker-buildcache:
 clean-docker-containers:
 	docker container prune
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -104,10 +104,10 @@ clean-docker-buildcache:
 	docker builder prune --all -f
 
 clean-docker-containers:
-	docker rm -vf $$(docker container ls -aq)
+	docker container ls -aq | xargs -n1 docker rm -vf
```
veekaybee, line 107: how about "just" `docker container prune`?

Makefile
```diff
@@ -104,10 +104,10 @@ clean-docker-buildcache:
...
 
 clean-docker-images:
-	docker rmi -f $$(docker image ls -aq)
+	docker image ls -aq | xargs -n1 docker rmi -f
```
agpituk, line 110: following on the comment by @veekaybee above, we could do 
`docker image prune -a` here so we have the same approach in both

Makefile
aittalam: `docker container prune` removes only stopped containers. The previous `docker rm -vf ...` removed all containers. I just kept the same semantics, thinking it'd help if we have any containers hanging (happened to me more than once).
+100 on only removing our own btw!

aittalam:
I added a commit with a possible solution for image removing:
- we run `docker images "mzdotai/*" -q | xargs -n1 docker rmi -f` to only remove our own images
- we run `docker image prune` to remove dangling images (useful for us because we generate tons of them)
WDYT @agpituk ?

agpituk:
> I added a commit with a possible solution for image removing:
> 
>     * we run `docker images "mzdotai/*" -q | xargs -n1 docker rmi -f` to only remove our own images
> 
>     * we run `docker image prune` to remove dangling images (useful for us because we generate tons of them)
> 
> 
> WDYT @agpituk ? Also please lmk if you have an answer to the question I raised in "Additional notes for reviewers" above

That works very well for me, thanks @aittalam !!

About the question, I agree, `clean-all` should call `clean-docker-all `plus additional stuff.  Happy for this to be merged as it is or for you to also review that one.

agpituk (approved):

PR #659, opened 2025-01-16
658 GPU support + related env vars on Linux + getting started instructions on Readme

### LINKED ISSUE #658: [BUG]: Linux issues (using GPU, passing env vars)
### Description

Two separate problems present only on Linux.

I tried:
* [ ] setting env vars, such as OPENAI_API_KEY, MISTRAL_API_KEY via `export OPENAI...` in the same terminal where I launch `make start-lumigator-build`
* [ ] enabling GPU use with RAY_WORKER_GPU, RAY_WORKER_GPU_FRACTION via export as above and also changing the value directly in docker-compose.yaml and launching with start-lumigator-build.
    * Note: This problem (even once the env var is passed) is due to the CPU image being forced in the [docker-compose.yaml here](https://github.com/mozilla-ai/lumigator/blob/main/docker-compose.yaml#L46).
 

### Reproduction

1. `export OPENAI_API_KEY=xxxxx`
2. `make start-lumigator-build`
3. See the message `"The OPENAI_... variable is not set. Defaulting to a blank string."`

For the second issue:
1. Change the values to: `RAY_WORKER_GPU: 1`and `RAY_WORKER_GPU_FRACTION: 1.0`
2. `make start-lumigator-build`
3. `docker ps` to see that the ray container is a cpu-only image: `...ray:2.30.0-py311-cpu`
4. (bonus) `docker inspect ray_container_id` and see nothing like "NVIDIA_VISIBLE_DEVICES=all" or "NV_CUDA_CUDART_VERSION"

### Relevant log output

```shell

```

### Expected behavior

All 4 environment variables would be used; the 2 first to pass on API keys to the Lumigator container and the 2 others to enable a GPU image of ray.

### System Info

* OS: Ubuntu 24.10
* Branch: main@ [80c51ff](https://github.com/mozilla-ai/lumigator/commit/80c51fff1cd5bc7a42300e16011d5c09fe1ee5ff)

ividal:
# What's changing

* Checks in the Makefile whether RAY_WORKERS_GPU was set and, if it was, uses the -gpu image from ray in docker-compose.
* Doc:
    * Adds instructions to users to enable a GPU and to pass env vars reliably (we have seen issues on a few methods, these seem to work for everyone so far). 
    * Updates instructions to refer to the UI once we run start-lumigator

Closes #658 

# How to test it

Steps to test the changes:

1. Follow the new instructions, in particular those related to env vars.
2. On Linux check whether the GPU image is used.

# I already...

- [P] Tested the changes in a working environment to ensure they work as expected
     - Tested this did not break start-lumigator on Mac 
     - Tested the correct GPU image is used on Linux and tests pass, but not that a GPU was actually used to run the test models


docker-compose.yaml
```diff
@@ -127,8 +127,8 @@ services:
       - RAY_HEAD_NODE_HOST
       - MISTRAL_API_KEY=$MISTRAL_API_KEY
       - OPENAI_API_KEY=$OPENAI_API_KEY
-      - RAY_WORKER_GPUS=0
-      - RAY_WORKER_GPUS_FRACTION=0
+      - RAY_WORKER_GPUS="1.0"
+      - RAY_WORKER_GPUS_FRACTION="1.0"
```
dpoulopoulos, line 131: Why are values for `RAY_WORKER_GPUS` and `RAY_WORKER_GPUS_FRACTION` hard-coded here? What are the implications if a worker has more than one GPU? What is it has none?

README.md
```diff
@@ -78,18 +78,37 @@ To start Lumigator locally, follow these steps:
     cd lumigator
     ```
 
-1. Start Lumigator using Docker Compose:
+1. If your system has an NVidia GPU, open a terminal and run:
+    ```bash
+    export RAY_WORKERS_GPU=1
+    export RAY_WORKERS_GPU_FRACTION=1.0
```
dpoulopoulos, line 84: Should we also add these variables in `.env.example`?

README.md
```diff
@@ -78,18 +78,37 @@ To start Lumigator locally, follow these steps:
...
+    ```
+    **Importat: Continue the next steps in this same terminal.***
```
dpoulopoulos, line 86:
```suggestion
    **Important: Continue the next steps in this same terminal.***
```

README.md
```diff
@@ -78,18 +78,37 @@ To start Lumigator locally, follow these steps:
...
+
+
+1. If you intend to use Mistral API or OpenAI API, use that same terminal and run:
+    ```bash
+    export MISTRAL_API_KEY=your_mistral_api_key
+    export OPENAI_API_KEY=your_openai_api_key
+    rm .env
+    ```
+    **Importat: Continue the next steps in this same terminal.***
```
dpoulopoulos, line 95:
```suggestion
    **Important: Continue the next steps in this same terminal.***
```

dpoulopoulos (changes requested):

docker-compose.yaml
ividal: Good catch! 🙏

ividal:
Finally. With this, on an ubuntu 24.10 with a GTX 4060 I'm seeing:

Screenshot From 2025-01-20 11-24-27

docker-compose.yaml
```diff
@@ -43,20 +43,27 @@ services:
...
-    image: rayproject/ray:2.30.0-py311-cpu${RAY_ARCH_SUFFIX}
+    image: rayproject/ray:2.30.0-py311${COMPUTE_TYPE}${RAY_ARCH_SUFFIX}
     ports:
       - "6379:6379"
       - "8265:8265"
       - "10001:10001"
     # https://docs.ray.io/en/releases-2.30.0/cluster/cli.html#ray-start for more info about the command
     command: bash -c "ray start --head --dashboard-port=8265 --port=6379 --dashboard-host=0.0.0.0 --ray-client-server-port 10001 --block" # pragma: allowlist secret
     shm_size: 2g
+    runtime: nvidia
```
ividal, line 54: Help needed: how can one introduce this only depending on COMPUTE_TYPE? Same as with the reservations section below.

Maybe @macaab26 knows how?

docker-compose.yaml
```diff
@@ -43,20 +43,27 @@ services:
...
     deploy:
       resources:
         limits:
           cpus: '2'
           memory: '5g'
+        reservations:
```
ividal, line 60: Help needed: see above. How can one introduce this whole section depending on COMPUTE_TYPE? 

Maybe @macaab26 knows how?

docker-compose.yaml
```diff
@@ -43,20 +43,27 @@ services:
...
+          devices:
+            - driver: nvidia
+              count: 1
+              capabilities: [gpu]
     environment:
+      - NVIDIA_VISIBLE_DEVICES=all
```
dpoulopoulos, line 66: I'd suggest making `all` and device count dynamic, such as `${GPU_COUNT:-0}` and `NVIDIA_VISIBLE_DEVICES=${NVIDIA_VISIBLE_DEVICES:-all}`.

docker-compose.yaml
dpoulopoulos: I'd suggest creating two distinct docker-compose files, on for each compute type.

davidmanzanoai: You can define conditions in a docker-compose file to control this (https://docs.docker.com/compose/how-tos/startup-order/). If we capture in the Makefile if the GPU is available we can control with a flag in that condidion. If it is not available the GPU that code won't be executed.

macaab26: There's no an "if" conditional accepted in Docker-compose files. Partially agree with @dpoulopoulos, the best solution is to create two different definitions. Probably it's not needed a different file, we can do it with a different profile in the actual one (one specific one for GPUs). That's up to you!

ividal: I was fearing ending up with 2 docker-compose files again, but it's likely much cleaner than some if statement.

Quick question: how about [composable docker-compose files](https://docs.docker.com/compose/how-tos/multiple-compose-files/include/)? As far as I understand, from the Makefile, we'd do `docker-compose up -f general-compose.yaml -f gpu-compose.yaml (...)` with `-f gpu-compose.yaml` depending on whether COMPUTE==gpu .

While @macaab26 is on something else, I can try that approach.

macaab26 (approved):

dpoulopoulos (approved):

PR #663, opened 2025-01-17
Add detailed experiment status

### LINKED ISSUE #571: Add get experiment status + tests

javiermtorres:
# What's changing

Adds detailed experiment status depending on the component job status.

Closes #571 

# How to test it

Integration tests are supplied with the change. Manually, the status can be retrieved and checked while the component jobs are cycling through their RUNNING and SUCCEEDED/FAILED status.

N/A

# I already...


lumigator/python/mzai/backend/backend/services/experiments.py
```diff
@@ -157,14 +157,30 @@ def get_experiment(self, experiment_id: UUID) -> ExperimentResponse:
...
                 all_succeeded = False
-                break
+                if self._job_service.get_job(job.id).status == JobStatus.FAILED:
+                    any_failed = True
+                    # Any failed job makes further inspection unnecessary
+                    break
+                if self._job_service.get_job(job.id).status == JobStatus.RUNNING:
+                    any_running = True
+                    # Any running job will move status to running, but
+                    # searching for failures is still necessary
```
aittalam, line 176: This is not needed by the current workflow as eval can fail only if inference succeded, but I am ok if we want to foresee different kind of composite tasks.
However, if we want this to work we should modify the sequence of `if`s below into disjoint events with `elif`s: otherwise, if something failed but something is still running we'll return `JobStatus.RUNNING` which is what we are trying to avoid by not having a `break` here.

lumigator/python/mzai/backend/backend/services/experiments.py
```diff
@@ -157,14 +157,30 @@ def get_experiment(self, experiment_id: UUID) -> ExperimentResponse:
...
 
         if all_succeeded:
             record = self._experiment_repo.update(experiment_id, status=JobStatus.SUCCEEDED)
+        if any_failed:
+            record = self._experiment_repo.update(experiment_id, status=JobStatus.FAILED)
+        if any_running:
+            record = self._experiment_repo.update(experiment_id, status=JobStatus.RUNNING)
```
aittalam, line 183:
```suggestion
        elif any_failed:
            record = self._experiment_repo.update(experiment_id, status=JobStatus.FAILED)
        elif any_running:
            record = self._experiment_repo.update(experiment_id, status=JobStatus.RUNNING)
```

The first `elif` is probably not required (as the other variables are set only if at least one job is not SUCCEEDED`) but the latter is needed if we want to enforce the behavior you are setting above (see previous comment)

aittalam: There is a slight difference in semantics between `created` and `pending` (created happens at DB insertion time, pending after ray submission), even if I cannot think about a situation where ray accepted the job at submission time and then never even put it into pending state. I think the above should cover all cases (ie. we'll have `created` only until ray answers us with information about the job)

aittalam (changes requested):
TY @javiermtorres ! I added minor fixes to be applied before merging.
Blocking this just because I want to make sure the above are already covered by tests before merging, but if you can confirm this I am happy to approve it!

lumigator/python/mzai/backend/backend/services/experiments.py
javiermtorres: I'm fine with the changes. I'll just re-run the tests with them and add documentation about the experiment states with these semantics 👍

lumigator/python/mzai/backend/backend/services/experiments.py
```diff
@@ -157,14 +157,32 @@ def get_experiment(self, experiment_id: UUID) -> ExperimentResponse:
         loguru.logger.info(f"Obtaining info for experiment {experiment_id}: {record}")
 
         all_succeeded = True
+        any_running = False
```
ividal, line 160: I feel I missed a step here: since the work around joining split jobs is in experiments_new, shouldn't collecting individual jobs' status also happen there?

This is more curiosity than change request.

javiermtorres: I'll check the route, but iirc the old experiment api is just a façade for calling an eval-old job through the jobs services, so no calls will be landed on the experiments service, right????

lumigator/python/mzai/backend/backend/services/experiments.py
javiermtorres: Ok, I just added your suggestion 👍

lumigator/python/mzai/backend/backend/services/experiments.py
ividal: Got it, the confusion comes from placing ExperimentsService, which is not in use in /experiments (since indeed that's a redirect to /jobs/evaluator right now) , inside experiments.py.

This might come around again when we make the switch, but at least that'll be the last time 🤞 :)

aittalam (approved):
TY @javiermtorres for implementing the new logic!
This looks good to me, I am happy to preapprove it. Re: impact on the current UI, there will be none as this service is hit only by the new endpoint

PR #674, opened 2025-01-17
Fixing db persistency in the local dev scenario

### LINKED ISSUE #515: [BUG]: sqlite database is not persisted (in dev environment at least)
### Description

Not sure if the problem is limited to the dev environment only, but the sqlite db is not persisted

### Reproduction

1. `make local-up`
2. save a dataset
3. check with sqlite3 client that the new data is in the `local.db` file in the container: ✅ 
5. check with sqlite3 client that the new data is in the `local.db` file on the host machine: ❌ 
6. `make local-down`
7. check that the `lumigator_database_volume` contains something: ❌ (size=0 bytes)
8. `make local-up`
9. check with sqlite3 client that the data is in the `local.db` file in the container: ❌ 

### Relevant log output

```shell

```

### Expected behavior

The volume should persist the database data, or there should be a sync with the `local.db` file on the host and the one in the container.
If the volume works, perhaps we do not need to also have the file?

### System Info

- OS: MacOs X
- Lumigator: [e6c3768](https://github.com/mozilla-ai/lumigator/commit/e6c37681d1ca641269fbb968cde2f5044fe907a5)

aittalam:
# What's changing

This PR binds the sqlite database in the backend container to a local file in `~/.lumigator/db`.
This allows the DB to persist from one session to another (i.e. even after `make local-down`+`make local-up`.

Closes #515 

# How to test it

Steps to test the changes:

1. run `make local-up`
2. add a dataset
3. run `make local-down`
4. run `make local-up`
5. get datasets list 

This is still a draft pr because:

- some integration tests run using this DB, which is not ideal. We should either prevent this or add in the docs that this might happen and how to avoid it
- while retrieving datasets works properly, jobs do not work as well because currently the list jobs method returns an intersection between jobs in the DB and jobs running on ray. Unless we persist ray data (which btw is possible), or start relying on the DB only to store/get job info, we won't be able to see jobs from a previous lumigator execution. 👉 NOTE that this is partially solved by https://github.com/mozilla-ai/lumigator/pull/744, we need to decide how to deal with logs in that case: see [this comment](https://github.com/mozilla-ai/lumigator/pull/744#pullrequestreview-2581646577))
- this is currently working in dev setup only. Do we want to have this as a stable feature in the `make start-lumigator` scenario? This still has to be discussed

### DIFF
--- a/.devcontainer/docker-compose.override.yaml
+++ b/.devcontainer/docker-compose.override.yaml
@@ -8,9 +8,13 @@ services:
       dockerfile: "Dockerfile"
       target: "dev_image"
     volumes:
-      -  database_volume:/mzai/backend/local.db
+      - type: bind
+        source: ~/.lumigator/db/
+        target: /lumigator_db/
     ports:
       - "5678:5678"
+    environment:
+      - SQLALCHEMY_DATABASE_URL=sqlite:////lumigator_db/local.db
     develop:
       watch:
         - path: lumigator/python/mzai/backend/


aittalam:
Closing as addressed in another PR

PR #676, opened 2025-01-17
fix: empty ground-truth column should mean no ground-truth

### LINKED ISSUE #620: [BUG]: Datasets with **empty** ground_truth column should still be considered as not having ground_truth
### Description

Currently if we upload a csv with an existing ground_truth column that is empty, we consider it as having a ground-truth. Instead, we should be checking whether there are any values in it and return false if it's truly empty.

Assumption for now: a value in any of the records already counts as ground-truth = True.

### Reproduction

1. git checkout main && git pull
2. make local-up
3. Browse to localhost (lumigator UI) 
4. Upload the attached CSV, which has an empty gt column
5. See on the UI gt column set as True

[dialogsum_mini_empty_gt.csv](https://github.com/user-attachments/files/18413388/dialogsum_mini_empty_gt.csv)

 

### Relevant log output

```shell

```

### Expected behavior

Ground-truth column should show "False".

### System Info

* OS and Linux
* Current main: [34873f6](https://github.com/mozilla-ai/lumigator/commit/34873f680f7d944610c65b077d6a44736607246c)

ividal:
# What's changing

Currently if we upload a csv with an existing ground_truth column that is empty, we consider it as having a ground-truth. Instead, we should be checking as well whether there are any values in it and return false if it's truly empty.

Closes #620 

# How to test it

Steps to test the changes:

1. git checkout this branch && git pull
1. make local-up
1. Browse to localhost (lumigator UI)
1. Upload [this CSV](https://github.com/user-attachments/files/18413388/dialogsum_mini_empty_gt.csv), which has an empty gt column (header is there, but no values)
1. See on the UI gt column set as **False**

Open to any **reasoned** choice between "ask for all rows to be annotated" and "1 sample is enough" to consider gt to be present. 

# I already...

- [X] Added some tests for any new functionality

### DIFF
--- a/lumigator/python/mzai/backend/backend/services/datasets.py
+++ b/lumigator/python/mzai/backend/backend/services/datasets.py
@@ -80,8 +80,8 @@ def dataset_has_gt(filename: str) -> bool:
     if GT_FIELD not in dataset.column_names:
         return False
 
-    # We'll accept that there's at least one value in the ground_truth column
-    return any(dataset[GT_FIELD])
+    # True only if every value in dataset[GT_FIELD] is not None or empty.
+    return all(value for value in dataset[GT_FIELD])
 
 
 class DatasetService:
--- a/lumigator/python/mzai/backend/backend/services/datasets.py
+++ b/lumigator/python/mzai/backend/backend/services/datasets.py
@@ -75,6 +75,9 @@ def validate_experiment_dataset(filename: str):
 
 
 def dataset_has_gt(filename: str) -> bool:
+    """Returns true if the dataset located at the supplied path (filename) has a ground truth
+    column with all of its rows correctly populated, otherwise false.
+    """
     dataset = load_dataset("csv", data_files=filename, split="train")
 
     if GT_FIELD not in dataset.column_names:
--- a/lumigator/python/mzai/backend/backend/services/datasets.py
+++ b/lumigator/python/mzai/backend/backend/services/datasets.py
@@ -81,7 +81,7 @@ def dataset_has_gt(filename: str) -> bool:
         return False
 
     # True only if every value in dataset[GT_FIELD] is not None or empty.
-    return all(value for value in dataset[GT_FIELD])
+    return all(value is not None and value.strip() != "" for value in dataset[GT_FIELD])
 
 
 class DatasetService:
--- a/lumigator/python/mzai/backend/backend/services/datasets.py
+++ b/lumigator/python/mzai/backend/backend/services/datasets.py
@@ -75,12 +75,13 @@ def validate_experiment_dataset(filename: str):
 
 
 def dataset_has_gt(filename: str) -> bool:
-    with Path(filename).open() as f:
-        reader = csv.DictReader(f)
-        fields = set(reader.fieldnames or [])
-        has_gt = GT_FIELD in fields
+    dataset = load_dataset("csv", data_files=filename, split="train")
 
-    return has_gt
+    if GT_FIELD not in dataset.column_names:
+        return False
+
+    # We'll accept that there's at least one value in the ground_truth column
+    return any(dataset[GT_FIELD])
 
 
 class DatasetService:
@@ -105,7 +106,7 @@ class DatasetService:
         return self.dataset_repo.get_by_job_id(job_id)
 
     def _get_s3_path(self, dataset_key: str) -> str:
-        return f"s3://{ Path(settings.S3_BUCKET) / dataset_key }"
+        return f"s3://{Path(settings.S3_BUCKET) / dataset_key}"
 
     def _get_s3_key(self, dataset_id: UUID, filename: str) -> str:
         """Generate the S3 key for the dataset contents.
--- a/lumigator/python/mzai/backend/backend/tests/conftest.py
+++ b/lumigator/python/mzai/backend/backend/tests/conftest.py
@@ -80,6 +80,16 @@ def valid_upload_file(valid_experiment_dataset) -> UploadFile:
 
 
 @pytest.fixture(scope="session")
+def valid_experiment_dataset_with_empty_gt() -> str:
+    """Minimal valid dataset without groundtruth."""
+    data = [
+        ["examples", "ground_truth"],
+        ["Hello World"],
+    ]
+    return format_dataset(data)
+
+
+@pytest.fixture(scope="session")
 def missing_examples_dataset() -> str:
     """Minimal invalid dataset without examples."""
     data = [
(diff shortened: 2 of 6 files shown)


lumigator/python/mzai/backend/backend/services/datasets.py
```diff
@@ -75,12 +75,13 @@ def validate_experiment_dataset(filename: str):
...
-        fields = set(reader.fieldnames or [])
-        has_gt = GT_FIELD in fields
+    dataset = load_dataset("csv", data_files=filename, split="train")
+
+    if GT_FIELD not in dataset.column_names:
+        return False
 
-    return has_gt
+    # We'll accept that there's at least one value in the ground_truth column
+    return any(dataset[GT_FIELD])
```
ividal, line 84: We need to settle on some agreement. All rows must be annotated? Bit extreme, but what's the threshold?

njbrake, line 84: If any rows don't have a target column, that will negatively bias all the aggregated metrics. although the bias would affect all the models, it might effect each of them to a different extent, since then the amount of text each model generates would have a bigger penalty. I think verbose predictions would be more impacted than concise predictions  (since the longer the summary the lower its precision score?).

My entire comment is irrelevant if there is already code later on from this logic that filters out items with no target/gt before running metric scoring
```suggestion
    # True only if every value in dataset[GT_FIELD] is not None or empty.
    return all(value for value in dataset[GT_FIELD])
```

njbrake (commented):
Regardless of which direction we take the function, maybe it also needs to be renamed to something like `dataset_has_some_gt` or `dataset_has_all_gt` to make what it's checking clearer, not sure.

lumigator/python/mzai/backend/backend/services/datasets.py
ividal: There isn't, we're quite accepting when it comes to data.
And, taking it to the extreme, even if we did we could be telling the user "Sure, we'll take you 100K sample dataset for evaluation.... Here are the results based on the single sample you provided with a label."

So I'll take all.

lumigator/python/mzai/backend/backend/services/datasets.py
```diff
@@ -75,12 +75,13 @@ def validate_experiment_dataset(filename: str):
...
-        fields = set(reader.fieldnames or [])
-        has_gt = GT_FIELD in fields
+    dataset = load_dataset("csv", data_files=filename, split="train")
+
+    if GT_FIELD not in dataset.column_names:
+        return False
 
-    return has_gt
+    # True only if every value in dataset[GT_FIELD] is not None or empty.
+    return all(value for value in dataset[GT_FIELD])
```
dpoulopoulos, line 84: This treats the empty string of length `1` or more (i.e., this `" "` not this `""`) as `True`. I'm not against that, just recognizing that an entry of `" "` is considered valid ground truth. Are we okay with that?

dpoulopoulos (commented):
Looks good to me, with a comment.

lumigator/python/mzai/backend/backend/services/datasets.py
peteski22: 👍🏼 good point. I think this also leads to us wanting to have more unit test cases that cover the scenarios where we say something has ground truth or not.

njbrake (approved):
LGTM

lumigator/python/mzai/backend/backend/services/datasets.py
```diff
@@ -75,12 +75,13 @@ def validate_experiment_dataset(filename: str):
 
 
 def dataset_has_gt(filename: str) -> bool:
-    with Path(filename).open() as f:
-        reader = csv.DictReader(f)
-        fields = set(reader.fieldnames or [])
-        has_gt = GT_FIELD in fields
+    dataset = load_dataset("csv", data_files=filename, split="train")
```
peteski22, line 78: ❓ why did this change from loading the CSV from a path as it's not been split up into HF formatted files etc. at this point?

I don't know if this makes it easier or harder to unit test properly as the return from `load_dataset` is a union of multiple types, vs. the `TextIO` that was returned before. 🤔 

We should add a doc string for the method too.

```
def dataset_has_gt(filename: str) -> bool:
    """Returns true if the dataset located at the supplied path (filename) has a ground truth
       column with all of its rows correctly populated, otherwise false."""
    with Path(filename).open() as f:
        reader = csv.DictReader(f)
        if GT_FIELD not in reader.fieldnames:
            return False

        # True only if every value in a ground truth cell is not None or empty.
        return all(row.get(GT_FIELD) and row[GT_FIELD].strip() for row in reader)
```

lumigator/python/mzai/backend/backend/tests/conftest.py
```diff
@@ -79,6 +79,16 @@ def valid_upload_file(valid_experiment_dataset) -> UploadFile:
     return fake_upload_file
 
 
+@pytest.fixture(scope="session")
+def valid_experiment_dataset_with_empty_gt() -> str:
```
peteski22, line 83: Please could you add some unit tests to `backend/tests/unit/services/test_dataset_service.py` to exercise the logic fully? 

e.g. testing with header missing, testing with header present, testing with header present and all relevant cells 👍🏼 , same with some relevant cells 👍🏼 , same with all relevant cells containing whitespace and so on.

`@pytest.mark.parametrize` should make this easier so you can just have a function along the lines of:

```
def test_dataset_has_ground_truth(dataset_csv_path, should_have_ground_truth):
    has_gt = dataset_has_gt(dataset_csv_path)
    assert has_gt == should_have_ground_truth
```

(*requires the CSV file to exist at the path too)

peteski22 (changes requested):
Thanks for doing this @ividal, I like the list comprehension stuff as a succinct way to validate this.

Sorry to seem like a nay-sayer, but I'd really like us to have good unit test coverage (test cases) for these smaller things as it prevents accidental regressions later.

lumigator/python/mzai/backend/backend/services/datasets.py
ividal: I don't feel too strongly about this one, but I had 2 reasons here:
* we're passing datasets around internally, so it seemed better to keep in line with that
* if we're going to start checking the full file for something, I'm willing to bet hf datasets have an efficient implementation and _since the dependency is already there_ throughout the codebase, let's benefit.

lumigator/python/mzai/backend/backend/tests/conftest.py
ividal: Yep, definitely, they're rather lacking - thanks for the nudge!

I liked the later suggestion of having it in a true unit test loading small, differently built files. Will add that in a min.

peteski22 (approved):
Thanks so much @ividal and thanks for being so open to feedback and collaborating! ❤️

Issue #480, opened 2024-12-09
[FEATURE] Evaluate experiment results


ividal:
The user can view and download evaluation scores after an experiment completes. 
 
* The system provides a comparison of model performance, including key metrics:
    * UI: #440, #454 
    * API: already present (return and download evaluation results)

PR #681, opened 2025-01-19
Issue 480 experiment results

george-mzai:
# What's changing
Enable users to compare job results within the same Experiment, provided that every job has a "SUCCEEDED" status.

# How to test it

1. Run an experiment with multiple models.
2. On the Experiments table click the experiment row in order to open the sidepanel
3. Make sure that the experiment is completed with status "SUCCEEDED".
4. At the bottom Experiment Details click the "View Results" button.
5. In the Results table expand the rows to display the results of each job.

Resolves: #480 
Closes: https://github.com/mozilla-ai/lumigator/issues/683


dpoulopoulos:
That's super useful! I just run it and it's looking great. One comment though: since this is running multiple experiments in parallel (at least two), I'm afraid we're going to bump into OOM issues for the workers. Is there any way to mitigate that? Running jobs one after another, even warning user that they may have to provide more resources if this fails?

HareeshBahuleyan:
Based on
> 4. At the bottom "Experiment Details click the "View Results" button.
>5. In the Results table expand the rows to display the results of each job.

I think, this https://github.com/mozilla-ai/lumigator/issues/683 bug could be resolved by this PR?

george-mzai:
> That's super useful! I just run it and it's looking great. One comment though: since this is running multiple experiments in parallel (at least two), I'm afraid we're going to bump into OOM issues in the Ray workers. Is there any way to mitigate that? Running jobs one after another, even warning user that they may have to provide more resources if this fails?

Yes, this is feasible.

george-mzai:
> Based on
> 
> > 4. At the bottom "Experiment Details click the "View Results" button.
> > 5. In the Results table expand the rows to display the results of each job.
> 
> I think, this #683 bug could be resolved by this PR?

Correct! This fixes #683

peteski22 (approved):
💪🏼 🥇

agpituk (approved):

PR #682, opened 2025-01-20
Add annotations to the SDK

### LINKED ISSUE #404: Add triggering annotations to the SDK
**Is your feature request related to a problem? Please describe.**
When the user's dataset lacked ground-truth, Lumigator offers the possibility to generate annotations. The SDK should allow users to access this.

**Additional context**
Requires #337

javiermtorres:
# What's changing

This PR adds an annotate Job type to the SDK.

Closes #404 

# How to test it

There is an integration test that takes a dataset with no ground truth, and checks that the result has indeed a `ground_truth` column.

N/A

# I already...

  - No DB migration needed


lumigator/python/mzai/schemas/lumigator_schemas/jobs.py
```diff
@@ -10,6 +10,7 @@ class JobType(str, Enum):
     INFERENCE = "inference"
     EVALUATION = "evaluate"
     EVALUATION_LITE = "eval_lite"
+    ANNOTATION = "annotate"
```
peteski22, line 13: ❓ Why do we not have this mapped to the string `annotation`? 

I see above that there's weird conflicting styles with `inference` vs `evaluate` (instead of `evaluation`) if we're choosing one way over the other, should we just update these to all be the same?

lumigator/python/mzai/sdk/lumigator_sdk/jobs.py
```diff
@@ -130,16 +134,18 @@ def create_job(self, type: JobType, request: JobEvalCreate | JobInferenceCreate)
 
         Args:
             type(JobType): The kind of job to create. It can be either
-                EVALUATION or INFERENCE.
+                EVALUATION_LITE, ANNOTATION or INFERENCE.
```
peteski22, line 137: ❓ Is `EVALUATION` not supported now (it looks like it is in the code below)? Sorry if this is a dumb question, I just haven't followed these developments.

Suggestion: make the job types lexicographical in their ordering.

```suggestion
                ANNOTATION, EVALUATION_LITE or INFERENCE.
```

peteski22 (approved):
Thanks Javier! I added some questions, but feel free to dismiss them if they make no sense.

😄

lumigator/python/mzai/schemas/lumigator_schemas/jobs.py
javiermtorres: Totally true. These strings correspond to the API endpoints. We should definitely use a single category (verbs vs nouns) at some point.

lumigator/python/mzai/sdk/lumigator_sdk/jobs.py
javiermtorres: Yes, standalone evaluation is still supported. I will use lexicographical order 👍

javiermtorres:
> Thanks Javier! I added some questions, but feel free to dismiss them if they make no sense.
> 
> 😄

Questions addressed! Of course they do make sense :)

PR #689, opened 2025-01-20
API: Models - add flag to indicate if a model requires an API key

peteski22:
# What's changing

API offers additional information on whether a model requires an API key to be configured.

# How to test it

Steps to test the changes:

1. `make local-up`

2. Hit the models API

```bash
curl -X 'GET' \
  'http://localhost:8000/api/v1/models/summarization' \
```

4. Check that `"requires_api_key"` appears for each model either `true` or `false`

image

Follow-up UI PR can be created to order the models in the UI.

# I already...


ividal (approved):
Thanks! LGTM!

Issue #670, opened 2025-01-17
Define interface between a job and the backend


njbrake:
### Motivation

Currently, jobs save a results.json file which is loaded by the backend and saved in the database. In order to add a level of abstraction between the job and the backend, let's split this out into a slightly more enforced interface, where the job saves a parameters, artifacts, and metrics json file with a loosely defined schema. The backend then load these files into pydantic validated classes before saving it to the database.

### Alternatives

1. Do nothing
2. Instead of multiple json files, save everything into a single results.json file

### Contribution

I'll do it! 

### Have you searched for similar issues before submitting this one?

- [x] Yes, I have searched for similar issues

njbrake:
During investigation I realized that currently there isn't any interface at all between an evaluation/evaluation_lite job and a backend, because the result.json is passed from the s3 bucket to the frontend without it being opened or validated by the backend. So this here issue I think is blocked until we de-couple the frontend and job artifact file structure, see #678

ividal:
Note that @javiermtorres had proposed a refactoring of jobs that might be relevant here. Just want to link the 2 of you in case it's useful!

njbrake:
> Note that [@javiermtorres](https://github.com/javiermtorres) had proposed a refactoring of jobs that might be relevant here. Just want to link the 2 of you in case it's useful!

thanks! I think this issue is relevant to that refactor request: https://github.com/mozilla-ai/lumigator/issues/706

Issue #678, opened 2025-01-17
Backend should return results data in a task agnostic format for frontend consumption


njbrake:
### Motivation

I believe this is also a relevant issue for the task that @HareeshBahuleyan is working on:

As I understand it currently, the metric names and settings are hardcoded into the frontend, e.g. https://github.com/mozilla-ai/lumigator/blob/main/lumigator/frontend/src/components/molecules/LExperimentResults.vue#L69

As we support new metrics and custom user metrics in the future, I'm thinking that the frontend should be able to remain ignorant about what the specific metric is.

In order to support this, I think we need to restructure the file whose path we provide to the frontend, which relates to https://github.com/mozilla-ai/lumigator/issues/670.

Currently, the file that the eval job creates and saves to S3 (which then the backend creates a presigned url so that the frontend can download the file) looks like this:

```
{
  "rouge": {
    "rouge1": [
      0.27272727272727276,
      0.27586206896551724
    ],
    "rougeLsum": [
      0.27272727272727276,
      0.20689655172413796
    ],
    "rouge1_mean": 0.27429467084639503,
    "rougeLsum_mean": 0.23981191222570536
  },
......
  "bertscore": {
    "precision": [
      0.8871338963508606,
      0.8618351817131042
    ],
    "recall": [
      0.8770406246185303,
      0.8888148069381714
    ],
...
    "hashcode": "hashymchashhash",
    "precision_mean": 0.8744845390319824,
    "recall_mean": 0.8829277157783508,
  },
  "predictions": [
    "a summary",
    "a summary2"
  ],
  "ground_truth": [
    "the real summary",
    "the real summary2"
  ]
} 
```

Can we re-org this to something that looks more like
```
{
  "metrics": {
    "rouge1": {
      "displayName": "ROUGE-1",
      "values": [
        0.27272727272727276,
        0.27586206896551724
      ],
      "mean": 0.27429467084639503
    },
    "rougeL": {
      "displayName": "ROUGE-L",
      "values": [
        0.27272727272727276,
        0.27586206896551724
      ],
      "mean": 0.27429467084639503
    },
    "bertscorePrecision": {
      "displayName": "BERTScore Precision",
      "values": [
        0.8871338963508606,
        0.8618351817131042
      ],
      "mean": 0.8744845390319824
    },
    "bertscoreRecall": {
      "displayName": "BERTScore Recall",
      "values": [
        0.8770406246185303,
        0.8888148069381714
      ],
[...]

PR #692, opened 2025-01-20
Use Lumigator Schemas to validate input and output of jobs (inference & eval_lite)

njbrake:
# What's changing
Type validation between backend and job: the job and backend both use the same pydantic class for input config and output file format. The backend also now loads and validates the inference job result as well as the evaluation_lite job result.

This helps to start positioning us to save the evaluation results into a system like mlflow.

This pull request introduces several changes to the codebase, focusing on enhancing job handling, validation, and schema management. The updates include modifications to configuration files, job service methods, and schema definitions.

### Configuration Updates:
* [`.vscode/settings.json`](diffhunk://#diff-a5de3e5871ffcc383a2294845bd3df25d3eeff6c29ad46e3a396577c413bf357L13-R14): Added `./lumigator/python/mzai/jobs` to `python.analysis.extraPaths`.
* [`docker-compose.yaml`](diffhunk://#diff-3fde9d1a396e140fefc7676e1bd237d67b6864552b6f45af1ebcc27bcd0bb6e9L110-R110): Updated `PYTHONPATH` to include `jobs` directory.

### Job Service Enhancements:
* `lumigator/python/mzai/backend/backend/services/jobs.py`: 
  - Imported new schemas for job configurations and outputs.
  - Added methods for validating job configurations and handling evaluation results. [[1]](diffhunk://#diff-b97303b13c281bbefe8de146d3b0431158c9f90a26f798be9072b224e26123bcR231-R238) [[2]](diffhunk://#diff-b97303b13c281bbefe8de146d3b0431158c9f90a26f798be9072b224e26123bcL157-R176)
  - Updated job creation logic to include validation and handling for evaluation jobs. [[1]](diffhunk://#diff-b97303b13c281bbefe8de146d3b0431158c9f90a26f798be9072b224e26123bcR333-R334) [[2]](diffhunk://#diff-b97303b13c281bbefe8de146d3b0431158c9f90a26f798be9072b224e26123bcR393-R401)

### Schema Refactoring:
* [`lumigator/python/mzai/jobs/evaluator_lite/eval_config.py`](diffhunk://#diff-2f03d9281f50a9d9215b5bbb6efae011b62ee80ae827999e0769fefde8390b13L1-L28): Removed old schema definitions.
* [`lumigator/python/mzai/jobs/evaluator_lite/schemas.py`](diffhunk://#diff-fec53d34a81cc91a6059e274e5b42909c5a6d79013d933d15642b1a74261a540R1-R67): Added new schema definitions for evaluation jobs.
* [`lumigator/python/mzai/jobs/inference/inference_config.py`](diffhunk://#diff-ff2c65426dffadb9c3ad89cac0f744875ff81776b687467f9ce8414ff5d89271L7-R13): Replaced local schema definitions with imports from shared schema file. [[1]](diffhunk://#diff-ff2c65426dffadb9c3ad89cac0f744875ff81776b687467f9ce8414ff5d89271L7-R13) [[2]](diffhunk://#diff-ff2c65426dffadb9c3ad89cac0f744875ff81776b687467f9ce8414ff5d89271L92-R118) [[3]](diffhunk://#diff-ff2c65426dffadb9c3ad89cac0f744875ff81776b687467f9ce8414ff5d89271L166-R163)
* [`lumigator/python/mzai/jobs/inference/schemas.py`](diffhunk://#diff-489dea071a533a3f96ea5e853359656b08fc78ec0a7ebe089f9e86656218c2caR1-R64): Added new schema definitions for inference jobs.

### Code Adjustments:
* [`lumigator/python/mzai/backend/backend/services/jobs.py`](diffhunk://#diff-b97303b13c281bbefe8de146d3b0431158c9f90a26f798be9072b224e26123bcL133-R141): Updated methods to use new schema validation and data handling.
* [`lumigator/python/mzai/jobs/inference/inference.py`](diffhunk://#diff-5aea7c090a840e4c817d85afa4c27df25402f32726005edff959425daab44bd3R19): Updated functions to use new schema definitions for saving and validating job outputs. [[1]](diffhunk://#diff-5aea7c090a840e4c817d85afa4c27df25402f32726005edff959425daab44bd3R19) [[2]](diffhunk://#diff-5aea7c090a840e4c817d85afa4c27df25402f32726005edff959425daab44bd3L31-R36) [[3]](diffhunk://#diff-5aea7c090a840e4c817d85afa4c27df25402f32726005edff959425daab44bd3L46-R47) [[4]](diffhunk://#diff-5aea7c090a840e4c817d85afa4c27df25402f32726005edff959425daab44bd3L126-R127)

These changes collectively improve the robustness and maintainability of the job handling and validation processes in the codebase.

Refs #670 

# How to test it

Steps to test the changes:

1. `make local-up`
2. Upload dataset
3. run experiment

you should see no changes.

# I already...


lumigator/python/mzai/backend/backend/config_templates.py
```diff
@@ -134,7 +134,8 @@
     "job": {{
         "max_samples": {max_samples},
         "storage_path": "{storage_path}",
-        "output_field": "{output_field}"
+        "output_field": "{output_field}",
+        "enable_tqdm": true
```
njbrake, line 138: Need this because the schema doesn't have a default of True. I could change the schema so that then this tqdm key doesn't need to be added

lumigator/python/mzai/backend/backend/ray_submit/submission.py
```diff
@@ -37,6 +38,11 @@ def submit_ray_job(client: JobSubmissionClient, entrypoint: RayJobEntrypoint) ->
     loguru.logger.info(
         f"Submitting {entrypoint.get_command_with_params}...{entrypoint.runtime_env}"
     )
+    # copy the lumigator_schema to the working directory so that it's uploaded with the source code
+    # This way, the jobs can use and adhere to the Lumigator schemas when running in ray.
+    working_dir = entrypoint.runtime_env["working_dir"]
+    os.system(f"cp -r ../schemas/lumigator_schemas/ {working_dir}")
```
njbrake, line 44: I'm open to thoughts about how else you would add lumigator_schemas to a job. I thought about adding it to the requirements.txt, but then you would need to publish it or somehow register it on the cluster.

veekaybee:
Thanks for the proposal! For some context, where we'd landed on this previously is including a lightweight set of validation schemas with each job (https://github.com/mozilla-ai/lumigator/blob/main/lumigator/python/mzai/jobs/evaluator_lite/eval_config.py). 

The problem and the pain points are real, but we don't want to ship an entire working directory to Ray, which we'd done with lm-buddy and now the old evaluator that we're looking to get rid of. We'd done that historically and since Ray "prefers" single scripts or smaller executables, we wound up on the solution of one (or as few as possible) .py scripts and a lightweight schema. 

However, as you've noticed it creates the problem we have now where we are doing schema validation at several boundary points of the application 🤦‍♀️ .

Knowing this context, what might you keep from what you have here? I have a couple thoughts but curious to hear what you and others think.

njbrake:
Thanks for the review and the background!


> The problem and the pain points are real, but we don't want to ship an entire working directory to Ray, which we'd done with lm-buddy and now the old evaluator that we're looking to get rid of. We'd done that historically and since Ray "prefers" single scripts or smaller executables, we wound up on the solution of one (or as few as possible) .py scripts and a lightweight schema.

Can you explain this more? As I understand it, the `experiments_new` endpoint and `evaluator_lite` job are still following the same strategy of zipping up the working directory. It's only a few files and takes up negligible space at time to send to ray when you compare it to the environment dependencies specified in the requirements.txt. I may be misunderstanding what you mean. Maybe if you could clarify why `we don't want to ship an entire working directory to Ray` that may help me get up to speed 🙏 

> 
> However, as you've noticed it creates the problem we have now where we are doing schema validation at several boundary points of the application 🤦‍♀️ .
> 
> Knowing this context, what might you keep from what you have here? I have a couple thoughts but curious to hear what you and others think.

I considered two options:

1. Move the jobs folder into the backend. This way the backend could directly call the schemas that are defined in the jobs. However, I decided against this, because each job is allowed to have its own set of dependencies, so it seems wrong to put it in the backend where you can try to import a module from a job that has dependencies that aren't supported by the backend. The more subtle issue with this is you could also have a version mismatch where for example the backend uses transformers version 4.34 but the job uses version 4.1. (this has caused me weeks of pain over the past few years in other projects 😆 )
2. Lumigator Schema is the source of truth for schemas between the client/sdk and the backend, as well as the jobs and the backend. I chose this because it seemed optimal to consolidate the schemas as much as possible into the package named "schemas" 😆 . So, knowing that I wanted to go with this option, I had to figure out how to provide the jobs with the lumigator_schemas dependency. The backend already consumes the lumigator_schema package, but not so with the jobs. Since they're run on a remote system (ray) I can't just specify that the package refers to a local path (as is done in the backend). Furthermore, I couldn't have it rely on a pypi package, since that would allow for the job to get out of sync with the schema, and slow down the development work.

This is why I went with adding the lumigator schema to the working directory that gets bundled up with the job source code. It's lightweight since it only copies a few schema files, it gives flexibility to develop the schemas and have them stay in sync with the job running on Ray, and it also gives us a shared interface between backend and job ✅ .

lumigator/python/mzai/schemas/lumigator_schemas/jobs/jobs.py
```diff
@@ -55,32 +55,34 @@ class JobSubmissionResponse(BaseModel):
     driver_exit_code: int | None = None
 
 
-class JobEvalCreate(BaseModel):
+class JobCreate(BaseModel):
```
javiermtorres, line 58: Instead of a base class, we could use a discriminated union to better support translation to openapi: https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions-with-str-discriminators

Otherwise, yes, I'd support having a "base" class that ties to the Lumigator job life cycle, while leaving the config details to specific classes.

lumigator/python/mzai/backend/backend/ray_submit/submission.py
javiermtorres: The schemas are (in the process of being) actually published to PyPI. However, I do agree on sharing them internally. Note that we are not adding jobs dynamically, so I'd do this kind of packaging (bundling it into a zip too might be worth the effort) once during startup.

Maybe adding a venv with the dependencies too is a possible strategy, but @veekaybee, if I got that right, you already explored this in lm_buddy and were not satisfied with the result?

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -215,7 +215,13 @@ def _set_model_type(self, request: BaseModel) -> str:
 
         return model_url
 
-    def _get_job_params(self, job_type: str, record, request: BaseModel) -> dict:
+    def _validate_config(self, job_type: JobType, config_template: str, config_params: dict):
+        if job_type == JobType.INFERENCE:
+            InferenceJobConfig.model_validate_json(config_template.format(**config_params))
```
javiermtorres, line 220: Note that having some sort of class representing a `JobPackage` would remove this kind of "if job_type == X" checks by just using the appropriate overloaded method, like `job.config_model.model_validate_json(...)`.

lumigator/python/mzai/schemas/lumigator_schemas/jobs/jobs.py
njbrake: Thanks for the reference! I didn't know about discriminators, better translation to OpenAPI sounds desireable. If I overhaul the class to be discriminator based, I think that would mean a change to the API exposed by the backend. We have a lot of work ahead to improve schema validation across the stack, and I'm trying to follow the old addage "how do you eat an elephant? One bite at a time!". If I add the JobCreate base class that the JobInferenceCreate , JobEvaluationCreate and JobEvalLiteCreate all inherit, this allows me to make the change transparent to the client.

Would you be ok if we delay the refactor of this class that would force a public API change to a later PR? It could be handled inside of https://github.com/mozilla-ai/lumigator/issues/678. I should say that this consolidation of the classes could be completely deferred, I was trying to make what I saw as a simple consolidation since I was already in the code, but if their is controversy about whether it's an improvement, I can undo this refactor of JobCreate to a later time.

The main part of this PR that I want to push through is the sharing of the lumigator_schema package between the backend and the jobs.

veekaybee:
> Can you explain this more? As I understand it, the experiments_new endpoint and evaluator_lite job are still following the same strategy of zipping up the working directory. It's only a few files and takes up negligible space at time to send to ray when you compare it to the environment dependencies specified in the requirements.txt. I may be misunderstanding what you mean. Maybe if you could clarify why we don't want to ship an entire working directory to Ray that may help me get up to speed 🙏

Yup, with this my note of caution is that if we end up including subdirectories and entrypoints as we did in the old `evaluator`, or grow to that,  it's much harder to debug issues rather than the lightweight way we have now of bundling job + schema.

lumigator/python/mzai/backend/backend/ray_submit/submission.py
veekaybee: yep, see here https://github.com/mozilla-ai/lumigator/pull/692#issuecomment-2604888575

lumigator/python/mzai/backend/backend/services/jobs.py
njbrake: to be addressed in https://github.com/mozilla-ai/lumigator/issues/706

docker-compose.yaml
```diff
@@ -107,7 +107,7 @@ services:
       - AWS_DEFAULT_REGION
       - AWS_ENDPOINT_URL
       - S3_BUCKET
-      - PYTHONPATH=/mzai/lumigator/python/mzai/backend
+      - PYTHONPATH=/mzai/lumigator/python/mzai/backend:/mzai/lumigator/python/mzai/jobs
```
javiermtorres, line 110: Instead of changing the PYTHONPATH directly, maybe we can use the features from uv: https://docs.astral.sh/uv/concepts/projects/dependencies/#path

njbrake: Thanks for the link! I actually had looked at that option, but we don't currently meet any of those preqrequisites because the job is not a package (no wheel, no zip, no pyproject.toml). Since each job may have its own dependencies, I think I would need to add a pyproject.toml to each job folder and then add a link to each individual job in the backend. Seemed like more overhead and one more place that woud need to be touched every time you add a new job definition, so I decided against it.

However, one thing I could do is to move this PYTHONPATH env var out of the docker-compose and instead put it into the Dockerfile. That might be a bettter place for the PYTHONPATH to be set anyways.

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -130,12 +132,13 @@ def _add_dataset_to_db(self, job_id: UUID, request: JobInferenceCreate, s3: S3Fi
         # Get the dataset from the S3 bucket
         result_key = self._get_results_s3_key(job_id)
         with s3.open(f"{settings.S3_BUCKET}/{result_key}", "r") as f:
-            results = json.loads(f.read())
+            results = InferenceJobOutput.model_validate(json.loads(f.read()))
+        # Validate that the output file adheres to
```
veekaybee, line 136: nit: let's leave the commend above the `results`, would also be helpful for legibility to type results

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -154,13 +157,23 @@ def _add_dataset_to_db(self, job_id: UUID, request: JobInferenceCreate, s3: S3Fi
...
         )
 
+    def _handle_evaluation_result(self, job_id: UUID, request: JobEvalLiteCreate, s3: S3FileSystem):
+        loguru.logger.info("Handling evaluation result")
+
+        # Get the dataset from the S3 bucket
+        result_key = self._get_results_s3_key(job_id)
+        with s3.open(f"{settings.S3_BUCKET}/{result_key}", "r") as f:
+            EvalJobOutput.model_validate(json.loads(f.read()))
+        # Currently all we do is load the output to validate.
```
veekaybee, line 174: nit: let's have these useful comments as the method docstring

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -154,13 +157,23 @@ def _add_dataset_to_db(self, job_id: UUID, request: JobInferenceCreate, s3: S3Fi
...
             generated=True,
-            generated_by=results["model"],
+            generated_by=results.model,
         )
 
         loguru.logger.info(
             f"Dataset '{dataset_filename}' with ID '{dataset_record.id}' added to the database."
         )
 
+    def _handle_evaluation_result(self, job_id: UUID, request: JobEvalLiteCreate, s3: S3FileSystem):
```
veekaybee, line 167: suggestion: rename to `_validate_evaluation_results` for now?

lumigator/python/mzai/jobs/evaluator_lite/schemas.py
```diff
@@ -0,0 +1,67 @@
+"""Schema file that is loaded by the backend to validate the input/output data.
```
veekaybee, line 1: 👍

lumigator/python/mzai/jobs/inference/inference.py
```diff
@@ -28,11 +29,11 @@ def predict(dataset_iterable: Iterable, model_client: BaseModelClient) -> list:
     return predictions
 
 
-def save_to_disk(local_path: Path, data_dict: dict):
+def save_to_disk(local_path: Path, results: InferenceJobOutput):
```
veekaybee, line 32: 👍

lumigator/python/mzai/jobs/evaluator_lite/schemas.py
```diff
@@ -0,0 +1,67 @@
...
+    return_input_data: bool = False
+    return_predictions: bool = False
+    storage_path: str
+    model_config = ConfigDict(extra="forbid")
+
+
+class EvalJobConfig(BaseModel):
+    name: str
+    dataset: DatasetConfig
+    model: ModelConfig
```
veekaybee, line 31: historically Pydantic has disallowed using `model` as a field name due to namespace conflicts with Model objects and might complain here

veekaybee (approved):
Thanks for taking this on! Left a few smaller comments but generally looks good.  I could foresee some issues in integration tests/python path issues that we don't catch until testing, but pre-approving, feel free to merge if you don't get any other comments once tests pass.

lumigator/python/mzai/jobs/evaluator_lite/schemas.py
njbrake: Hmm. It's not throwing any complaints. I know it doesn't like anything with model + an underscore like `model_foo` so maybe it doesn't care as long as the underscore isn't there 🤷

Issue #492, opened 2024-12-11
[FEATURE]: caching for HF models in local setting


aittalam:
### Motivation

Whenever lumigator is restarted, the first job takes way more time to complete. This is due to the fact that both caches for the ray pip env and the HF models are stored in the corresponding VM and are deleted at each restart.

Note that this issue probably does not exist in the hybrid setting, where we can assume the ray server to have longer uptimes and has persistent caching, but it would still occur anywhere ppl choose to run ray as a container.

### Alternatives

An idea could be to make these caches persistent, by mapping the corresponding directories to the host environment.

### Contribution

With the following change:

```
diff --git a/.devcontainer/docker-compose.override.yaml b/.devcontainer/docker-compose.override.yaml
index fb8ed8e..31d98d2 100644
--- a/.devcontainer/docker-compose.override.yaml
+++ b/.devcontainer/docker-compose.override.yaml
@@ -2,6 +2,19 @@ name: lumigator

 services:

+  ray:
+    volumes:
+      # map ray huggingface cache to local one
+      # (this allows to cache downloaded models across local env restarts)
+      - ${HOME}/.cache/huggingface:/home/ray/.cache/huggingface
+      # map ray_pip_cache to local directory under .lumigator
+      # (this allows to cache python environments across local env restarts)
+      - ${HOME}/.lumigator/ray_pip_cache:/ray_pip_cache

```

we get:

1. caching of huggingface models (which works with my currrent laptop cache, so anything I download for my local stuff is also cached on ray and vice-versa)
2. caching of pip environments, so we don't have to download the same packages over and over again each time we restart ray

The second step is a bit of a hack: for this to work, we just persist a custom dir (/ray_pip_cache) and then we substitute the pip cache directory with a symbolic link to our dir after the container is started:

```
docker exec `docker ps | grep lumigator-ray | awk '{print $1}'` sh -c "rmdir /tmp/ray/session_latest/runtime_resources/pip/ && ln -s /ray_pip_cache /tmp/ray/session_latest/runtime_resources/pip"
```

Not sure this is the ideal way of doing this, but I am happy to hear your thoughts and send a PR for it if you think it can be useful.

### Have you searched for similar issues before submitting this one?

- [x] Yes, I have searched for similar issues

javiermtorres:
We will limit the scope to caching HF models and datasets, and open a new PR for the pip env caching.

javiermtorres:
[...]

PR #698, opened 2025-01-21
Share HF cache from host home

javiermtorres:
# What's changing

The directory for the HF artifact cache will be shared with the host in docker-compose installs (`${HOME}/.cache/huggingface`).

In tests, this directory is created with permissions set to 777.

Refs #492 partially.

# How to test it

Start Lumigator with `make local-up` or `make start-lumigator`. Any HF models from the user's home should be already available in lumigator itself, decreasing the job run times.

N/A

# I already...

  - DB is not updated


docker-compose.yaml
```diff
@@ -152,6 +155,6 @@ services:
       - 80:80
 
 volumes:
-
     minio-data:
     database_volume:
+    ray-huggingface-cache:
```
peteski22, line 160: Is ‘ray-huggingface-cache‘ the one that’s commented out above now?

docker-compose.yaml
```diff
@@ -51,6 +51,9 @@ services:
     # https://docs.ray.io/en/releases-2.30.0/cluster/cli.html#ray-start for more info about the command
     command: bash -c "ray start --head --dashboard-port=8265 --port=6379 --dashboard-host=0.0.0.0 --ray-client-server-port 10001 --block" # pragma: allowlist secret
     shm_size: 2g
+    volumes:
+      - ${HOME}/.cache/huggingface:/home/ray/.cache/huggingface
+    # - ray-huggingface-cache:/home/ray/.cache/huggingface
```
peteski22, line 56: Can this be removed?

```suggestion
```

peteski22 (commented):
Cool! 😎

docker-compose.yaml
javiermtorres: Hum. After checking it out in depth, I'd rather:

* Keep this line uncommented in the standard docker.compose
* Keep the volume mount to the home dir in the docker override 

In that way, HF caches are persistent in the case of restarts in the default case, allowing sharing with the user cache in the best case. The clean make target should be able to remove the volume if anything happens.

@aittalam @agpituk any thoughts about this?

docker-compose.yaml
javiermtorres: Yes, afaict.

docker-compose.yaml
aittalam: Thanks Irina, I think it makes a lot of sense. We could map `${HF_HOME}` here and set it in `.env.template` so it defaults to `${HOME}/.cache/huggingface` if no value is already provided. WDYT?

docs/source/get-started/installation.md
```diff
@@ -28,6 +28,12 @@ Lumigator requires an SQL database to hold metadata for datasets and jobs. The l
 uses SQLite for this purpose.
 ```
 
+```{note}
+The Ray container will use the shared host folder `${HOME}/.cache/huggingface/` to store
```
aittalam, line 32: Just a note to update this too if you eventually choose to use HF_HOME 🙏

docker-compose.yaml
javiermtorres: LGTM @aittalam

dpoulopoulos (commented):
Looks good to me after fixing the `$HF_HOME` variable issue.

docker-compose.yaml
javiermtorres: Would you consider as well using a separate docker volume instead as indicated by a separate env var, or you don't think it's worth it @aittalam @ividal @dpoulopoulos @peteski22 ?

aittalam (approved):
LGTM!
I tested the branch and it works as expected, caching the models in the HF cache dir.
Feel free to merge after you sync with main

PR #691, opened 2025-01-20
UI/models/separate api hf

peteski22:
# What's changing

Separates the UI display of the API and non-API models as per the mock-ups.

Relates to: https://github.com/mozilla-ai/lumigator/pull/689

# How to test it

Steps to test the changes:

1. `make local-up`
2. Upload a dataset
3. Create an experiment
4. Models should be split up correctly
5. Created experiment should show all your selected models regardless of which section they came from.

image

image

# Additional notes for reviewers

NOTE: must be rebased/merged from main once (if) #689 merges.

# I already...

- [x] Tested the changes in a working environment to ensure they work as expected
- [ ] Added some tests for any new functionality
- [ ] Updated the documentation (both comments in code and [product documentation](https://mozilla-ai.github.io/lumigator) under `/docs`)
- [ ] Checked if a (backend) DB migration step was required and included it if required

PR #700, opened 2025-01-21
API/UI: Make model data more flexible for future use

peteski22:
# What's changing

https://github.com/mozilla-ai/lumigator/pull/689 updated the API to flag when a model required an API key.
https://github.com/mozilla-ai/lumigator/pull/691 added support for this in the UI so models could be displayed in different lists.

This PR refactors these changes to allow for more flexibility as we want to categorize models differently.

Instead of a `bool` flag returned in the API for `requires_api_key` we  can return `requirements` which contains a list (think set) of the requirements for this model. e.g. `api_key`, `llamafile`.

# How to test it

Steps to test the changes:

1. `make local-up`
2. Try to create an experiment in the UI and ensure no changes to the UI
3. Try to use the [API docs](http://localhost:8000/docs#/models/get_suggested_models_api_v1_models__task_name__get) to get models for `summarization`

In this screen shot you can see the different requirements being returned, the idea here is that as we want to further sub-divide the way we show models to users, it will be easier.

image

# I already...

### DIFF
--- a/lumigator/python/mzai/backend/backend/tests/data/models.json
+++ b/lumigator/python/mzai/backend/backend/tests/data/models.json
@@ -5,7 +5,7 @@
         "name": "facebook/bart-large-cnn",
         "uri": "hf://facebook/bart-large-cnn",
         "description": "BART is a large-sized model fine-tuned on the CNN Daily Mail dataset.",
-        "metadata": {},
+        "requirements": [],
         "info": {
           "parameter_count": "406M",
           "tensor_type": "F32",
@@ -28,7 +28,7 @@
         "name": "Falconsai/text_summarization",
         "uri": "hf://Falconsai/text_summarization",
         "description": "A fine-tuned variant of the T5 transformer model, designed for the task of text summarization.",
-        "metadata": {},
+        "requirements": [],
         "info": {
           "parameter_count": "60.5M",
           "tensor_type": "F32",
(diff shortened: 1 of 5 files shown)


george-mzai (approved):

lumigator/python/mzai/backend/backend/tests/data/models.json
```diff
@@ -103,7 +103,7 @@
         "name": "mistralai/Mistral-7B-Instruct-v0.2",
         "uri": "llamafile://mistralai/Mistral-7B-Instruct-v0.2",
         "description": "A llamafile package of Mistral's 7B Instruct model.",
-        "requires_api_key": false,
+        "requirements": [ "llamafile" ],
```
ividal, line 106: Nice 👍

ividal (approved):
This looks good, thanks @peteski22 !

lumigator/python/mzai/schemas/lumigator_schemas/models.py
```diff
@@ -12,6 +12,6 @@ class ModelsResponse(BaseModel):
     uri: str
     website_url: str
     description: str
-    requires_api_key: bool = False
+    requirements: list[str] = Field(default_factory=list)
```
aittalam, line 15: It might be good to capture the allowed keywords somewhere. WDYT about using an enumeration type here instead of freeform strings? I am also ok with merging this and capturing the improvement with an issue.

aittalam (approved):
Thanks @peteski22 , I left a comment for an improvement I think we should implement, but feel free to capture it within a follow-up issue.

lumigator/python/mzai/schemas/lumigator_schemas/models.py
peteski22: 💯 %! I initially left it quite, loose, because I wasn't sure what else this would get used for (when I had metadata with requirements inside it). 

I've updated this now: https://github.com/mozilla-ai/lumigator/pull/700/commits/a21f8537cae804b235c6400d0f73e504a202d575

PR #712, opened 2025-01-22
Replace pip cache of session_latest

javiermtorres:
# What's changing

Replaces the `/tmp/ray/session_latest/runtime_resources/pip` with a link to `/tmp/ray_pip_cache`. This location is placed in either a separate volume, or shared with the host.

Refs #492 partially

# How to test it

Start Lumigator with `make local-up`. The directory `/tmp/ray_pip_cache` needs to exist beforehand with 777 permissions. Submit an experiment using the UI. The pip cache should be present in the host directory.

N/A

# I already...

  - No DB changes


docker-compose.yaml
```diff
@@ -49,7 +49,14 @@ services:
       - "8265:8265"
       - "10001:10001"
     # https://docs.ray.io/en/releases-2.30.0/cluster/cli.html#ray-start for more info about the command
-    command: bash -c "ray start --head --dashboard-port=8265 --port=6379 --dashboard-host=0.0.0.0 --ray-client-server-port 10001 --block" # pragma: allowlist secret
+    entrypoint:
+      - /bin/bash
+      - -c
+      - |
+          ray start --head --dashboard-port=8265 --port=6379 --dashboard-host=0.0.0.0 --ray-client-server-port 10001
+          mkdir -p /tmp/ray/session_latest/runtime_resources/pip
```
dpoulopoulos, line 57: When the Ray container is up and running it will create the `/tmp/ray/session_latest/runtime_resources/pip` directory, won't it? So, why do we need the `mkdir` command? If I'm correct, we can simplify this to:

```
ray start --head --dashboard-port=8265 --port=6379 --dashboard-host=0.0.0.0 --ray-client-server-port 10001
rmdir /tmp/ray/session_latest/runtime_resources/pip
ln -s /tmp/ray_pip_cache /tmp/ray/session_latest/runtime_resources/pip
sleep infinity
```

Also, what happens if `/tmp/ray/session_latest/runtime_resources/pip` is not empty? Would it make more sense to run `rm -rf /tmp/ray/session_latest/runtime_resources/pip` instead of using `rmdir`?

docker-compose.yaml
```diff
@@ -78,6 +85,8 @@ services:
       - "localhost:host-gateway"
     profiles:
       - local
+    volumes:
+      - ray-pip-cache:/tmp/ray_pip_cache
```
dpoulopoulos, line 89: Being super pedantic here, but creating files or folders with predicable names under `/tmp` is a security risk. Could we randomize that?

docker-compose.yaml
dpoulopoulos: Should we also add a `make` target to clean up the pip cache locally?

javiermtorres: Re: the first comment, you're probably right. Let me recheck the process.

Re: the make target, we have talked about [using labels](https://docs.docker.com/engine/manage-resources/labels/) to specify what to remove. Let's address this issue then.

docker-compose.yaml
javiermtorres: Hmmmm... It is inside the container, after all, and for the current session the Ray tmp directories are well known. Furthermore, this is bound within the compose, and specifying a dynamic directory probably needs creating the name outside the compose, passing it as an env var and coordinating with the entrypoint too. If you can think of an easy-ish way to do it, I don't have anything against it.

docker-compose.yaml
javiermtorres: Oh, I found out (again) why the mkdir is needed :D The mount is only used for dev setups (local-up), so when starting a "standalone" setup this directory needs to be created. The -p ensures it won't complain because of the directory existing already.

docker-compose.yaml
dpoulopoulos: No, I'm okay with that. As I said, I'm being super pedantic...

dpoulopoulos (commented):
@javiermtorres looks good to me. I'll run some tests locally and then approve it!

docker-compose.yaml
```diff
@@ -65,10 +65,24 @@ services:
...
+          set -eaux
+          mkdir -p /tmp/ray_pip_cache
+          # If the file was mounted in a volume instead of
+          # a shared dir, permissions need to be setup
+          # ... || true allows this to fail (-e is set)
+          sudo chmod -R 777 /tmp/ray_pip_cache/ || true
+          RAY_REDIS_ADDRESS=redis:6379 RAY_JOB_ALLOW_DRIVER_ON_WORKER_NODES_ENV_VAR=1 ray start --head --dashboard-port=8265 --port=6379 --dashboard-host=0.0.0.0 --ray-client-server-port 10001
+          mkdir -p /tmp/ray/session_latest/runtime_resources/pip
+          rmdir /tmp/ray/session_latest/runtime_resources/pip/ && ln -s /tmp/ray_pip_cache /tmp/ray/session_latest/runtime_resources/pip
+          sleep infinity
```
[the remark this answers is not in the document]
javiermtorres: We cannot in this case because the Ray image is used as-is form the hub. The usual trick in Kubernetes would be to store this in a ConfigMap generated from some source files, but I'm not sure we can do the same in Docker.

dpoulopoulos (approved):

agpituk (approved):
Looks great! Do you think we can try to warm up this cache in the startup process? not for this PR but as an improvement thing

PR #715, opened 2025-01-22
Debugpy comes to Unit Tests!

njbrake:
# What's changing

I added a way for you to debug all of our unit tests!

# How to test it

Steps to test the changes:

1. Delete your .env file so that you'll get the new .env.template: `rm .env`
2. `export TESTS_USE_DEBUGPY=true`
3. `make test-backend-unit` : observe that execution is paused and you see `Waiting for debugger attach on port 5679...`
4. Go into Vscode Debugger  Screenshot 2025-01-22 at 3  02 26 PM  and click Screenshot 2025-01-22 at 3 02 46 PM 
5. Set breakpoints to your hearts content.

Putting the logic to optionally import debugpy in each __init__.py assures that it's the first file loaded when you run each module. It only tries to import debugpy if you specifically asked to debug

# I already...


docs/source/operations-guide/dev.md
```diff
@@ -144,3 +144,10 @@ To fix linting issues automatically:
 ```console
 user@host:~/lumigator/lumigator/frontend$ npm run lint:fix
 ```
+
+
+### Debugging with VSCode and DebugPy
+
+The backend and unit tests can be debugged with VsCode using Debugpy. If you're not familiar with how debugging python in vscode works, we recommend you read VSCode's documentation [here](https://code.visualstudio.com/docs/python/debugging). When running the command `make local-up`, the backend automatically creates the debugpy listener, which you can attach to using the `Attach to Backend` Debug configuration present in the `.vscode/launch.json` file of this repo.
```
HareeshBahuleyan, line 151: Nice, thanks for the documentation on debugging.

lumigator/python/mzai/sdk/tests/__init__.py
```diff
@@ -0,0 +1,12 @@
+# if the use_debugpy is set to True, the debugpy will be used to debug the test cases
```
HareeshBahuleyan, line 1: I was wondering if there was a way to avoid copy pasting the same across all the init files?
Like for example, if we could have it as a function `setup_debugpy()` in a file `debug_utils.py` somewhere (not sure where 😛 ) and then each init file would have
```
from <somewhere>.debug_utils import setup_debugpy
setup_debugpy()
```

njbrake:
> I'd prefer to keep IDE-specific settings out of our main codebase and test paths - is there a way to include this through the Makefile or Docker container commands as with the previous debugger PR?

The unit tests are a bit different from the previous PR.

1. The unit tests run outside of docker, so putting a debug call in the docker container isn't an option currently
2. When enabled, the unit test debugger pauses execution to wait for the developer to attach to the debugger. I do this because if unit tests are running quickly, you may attach the debugger too late and the test you wanted to debug may have already happened (aka race condition). Because of this "wait for client" configuration, the setting can't always be turned on. 

We could pipe new commands or extra logic into the makefile, but our Makefile is already pretty big and I hesitated to add this logic to further bloat. However, if that would be preferred over updating the __init__ files, I could do that instead. If we went that route, it would also address @HareeshBahuleyan 's comment about all the code copying: the Makefile could be that shared place to optionally enable the debugger for any unit test.

veekaybee:
> our Makefile is already pretty big

I share this concern and feel that! But I'd rather optional tooling (ie not everyone using lumigator will use vs code + this module) be included there than in the codepath directly and let folks control that based on their dev preferences via Make/env variables rather than have the app work to detect it in its internal logic. Let's see how this approach goes, and thanks for taking this.

njbrake:
> > our Makefile is already pretty big
> 
> I share this concern and feel that! But I'd rather optional tooling (ie not everyone using lumigator will use vs code + this module) be included there than in the codepath directly and let folks control that based on their dev preferences via Make/env variables rather than have the app work to detect it in its internal logic. Let's see how this approach goes, and thanks for taking this.

Thanks for the feedback! I refactored it so that it goes into the makefile. Now that it's in there, it does feel like it's cleaner than my initial solution, so thank you for the review :)

veekaybee (approved):
Looks good, let's make sure everyone working on the app is aware since there's makefile changes but approved from my side!

Issue #693, opened 2025-01-20
[BUG]: Link Checker Github Action is fooled on feature branches because of hardcoded links to /blob/main


njbrake:
### Description

I completed this PR https://github.com/mozilla-ai/lumigator/pull/614 which renamed a file from .env.example to .env.template. All of the actions passed, and I merged it to main, only to get a notification that a github action was failing and I had broken the main branch!  Serves me right for doing a merge to main at the end of the day on a Friday 😆 . The issue was that I had forgotten to update some documentation: PR to fix my mistake is here https://github.com/mozilla-ai/lumigator/pull/679.

This should not have been able to happen, the link checker should identify broken links before they happen.

The issue comes from how we reference links to files in our github repo.  We often reference files using HTTP links to the main branch, like

https://github.com/mozilla-ai/lumigator/blob/main/README.md?plain=1#L63 

`[link text](https://github.com/mozilla-ai/lumigator/blob/main/.env.template) `

This means that when the link checker runs on a feature branch, it's checking links that point to the main branch, which is why it missed that I had renamed a file and thus broken a link. 

### Reproduction

1. make a feature branch
2. delete the .env.template file, commit the change, push to github, make a PR
3. Observe that the github action will pass even though the file is gone and the links to that file in the repo should now be broken.

### Relevant log output

```shell

```

### Expected behavior

I think the links in the repo docs should be updated to be relative links, this way they will point to the files that exist on the branch itself, and not always to whatever is in main. This will become more important as we maintain versions of lumigator. (aka we don't want version 1.0 to accidentally be referencing to newer 2.0 docs). 

### System Info

n/a

### Have you searched for similar issues before submitting this one?

- [x] Yes, I have searched for similar issues

PR #717, opened 2025-01-22
Re-Organize Folders

njbrake:
# What's changing
Re-organize the folder structure to more closely align with standard python library packaging.
After re-organization, the hardest part was fixing the links in the docs: they were hardcoded links to git commit hashes or sometimes the main branch. This was a latent issue (see https://github.com/mozilla-ai/lumigator/issues/693) so this PR fixes that too. Now the sphinx build grabs the git commit and uses that to construct all the links, which is what we would want eventually anyways so that we could easily build the docs for any release and have it point to the correct files.

I also updated the github action "link checker" so that it is only responsible for checking the links outside of the docs folder. The github action link checker isn't compatible with the dynamic git commit insertion that is build into sphinx. Sphinx has a built in linkcheck extension which I activated, so we should still be all set with getting the links checked.

The old structure:
```
docs
lumigator
    frontend
    infra
        mzai
            helm
                lumigator
    python
        mzai
            backend
            jobs
            sample_data
            schemas
            sdk
```

The new structure I am proposing:

```
docs
infra
    helm
        lumigator
lumigator
    frontend
    backend
    jobs
    sample_data
    schemas
    sdk
```

# I already...

- [x ] Tested the changes in a working environment to ensure they work as expected
- [x ] Updated the documentation (both comments in code and [product documentation](https://mozilla-ai.github.io/lumigator) under `/docs`)
- [ x] Checked if a (backend) DB migration step was required and included it if required


javiermtorres:
Can you please provide a short summary of the relocations?

docs/Makefile
```diff
@@ -7,6 +7,7 @@ SPHINXOPTS    ?=
 SPHINXBUILD   ?= sphinx-build
 SOURCEDIR     = source
 BUILDDIR      = build
+GIT_COMMIT    ?= $(shell git rev-parse HEAD)
```
veekaybee, line 10: could you share what this change related to, is this the docs?

docs/source/conf.py
```diff
@@ -6,14 +6,19 @@
...
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
+
+# Get the commit hash from git
+import subprocess
 import sys
 from pathlib import Path
 
+commit_id = subprocess.check_output(["git", "rev-parse", "HEAD"]).strip().decode("utf-8")
+print(f"Git Commit ID: {commit_id}")
```
veekaybee, line 16: where does this print go, should we be logging it?

docs/source/conceptual-guides/endpoints.md
```diff
@@ -13,40 +13,40 @@ cluster, SQL database, and S3-compatible Storage) are the different services Lum
 The components inside the backend, shown in the image below, are the different abstraction layers
 the backend itself relies on:
 
-* The **API** makes backend functionalities available to the UI through different **routes** (see: [schema code](https://github.com/mozilla-ai/lumigator/tree/b1ea63ba3e1aae5907e46ffbe9bfd809253c6053/lumigator/python/mzai/backend/backend/api/routes)).
-  [**Schemas**](https://github.com/mozilla-ai/lumigator/tree/b1ea63ba3e1aae5907e46ffbe9bfd809253c6053/lumigator/python/mzai/schemas/schemas)
+* The **API** makes backend functionalities available to the UI through different **routes** (see: {{ '[schema code](https://github.com/mozilla-ai/lumigator/blob/{}/lumigator/backend/backend/api/routes)'.format(commit_id) }} ).
```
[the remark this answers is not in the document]
njbrake: It comes from the conf.py, which is executing a `git rev-parse` command to grab the commit hash. This is the way mysty enforces link substitution, per their docs https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#substitutions-and-urls

docs/source/operations-guide/kubernetes.md
```diff
@@ -27,7 +27,7 @@ with a minimal configuration ready to work with Lumigator.
 
 > [!NOTE]If the Mistral and/or the OpenAI API is used, there are two ways to provide
 > it to Lumigator:
-> 
+>
```
veekaybee, line 30: nit, extra space

infra/helm/lumigator/README.md
```diff
@@ -6,10 +6,10 @@ This Helm chart is the official way to deploy Lumigator in cloud environments wi
 
 This Helm chart is composed by two sub-charts:
 
-## `backend` 
+## `backend`
```
veekaybee, line 9: nit: i don't think we need to change this file (unless this was ruff, then feel free to ignore)

docs/Makefile
njbrake: This is for the docs, basically we currently are linking to hard-coded git hashs (which basically means we're linking in our current code to old code, which is a live link so it passes our CI checks, but it's actually an invalid link), or to the main branch, which will be an issue if we want to build our docs for a release branch.

lumigator/backend/backend/api/http_headers.py
```diff
@@ -1,3 +1,4 @@
 class HttpHeaders:
     """HttpHeaders represents constants used in HTTP headers"""
+
```
veekaybee, line 3: same here re: ruff

docs/source/operations-guide/kubernetes.md
njbrake: I think my PR is removing the space? Several docs never had the linter run over them, so you'll see this in several places. My changes are adding in the linting/formatting changes enforced by our pre-commit checks.

docs/source/conf.py
njbrake: The print goes in the github action. We could remove it, I"m fine with either way

lumigator/backend/backend/api/http_headers.py
njbrake: Yes, this stuff was either never edited after we added ruff, or was committed using the `--no-verify` tag.

Dockerfile
```diff
@@ -4,8 +4,8 @@ COPY --from=ghcr.io/astral-sh/uv:latest /uv /bin/uv
 # Copy the project into the image
 ADD . /mzai
```
veekaybee, line 5: `ADD . /mzai`
should we change this path since we're no longer using any variation of it?

njbrake:
> This looks ok to me at a high-level glance although it's fairly large (as we anticipated) 😅 . Is there an easy way to logically separate out the "renaming" aspect from fixing the links in the google docs? If possible, it would be good to have as two PRs.

Not really: since the links part is currently tied to main branch, the CI would fail if I just updated the paths. For example:

`docs/source/conceptual-guides/endpoints.md` currently has a link to **Schemas**](https://github.com/mozilla-ai/lumigator/tree/b1ea63ba3e1aae5907e46ffbe9bfd809253c6053/lumigator/python/mzai/schemas/schemas). This is bad because it's linking to a hash that is disconnected from the version of the code that the docs is built for. I have three options:

1.  I fix the bad link but keep the hardcoding: If I update that code in this PR to be

`**Schemas**](https://github.com/mozilla-ai/lumigator/tree/main/lumigator/schemas/schemas)` then it will break because the new file structure doesn't exist on main, so CI will be broken but then I'll have to trust that it will work when that file does exist on main

2. I update to point it to main but keep the hardcoding

`**Schemas**](https://github.com/mozilla-ai/lumigator/tree/main/lumigator/python/mzai/schemas/schemas)` this will work in my PR CI buiild, but then will cause a build failure as soon as I merge it to main, because that file path will no longer exist once the merge is finished, oh no!

3. I fix the hard coding and the bad link
This is the solution I went with: it makes it so that our docs don't point to the wrong version of code, and also makes it so that our CI builds will pass both on the PR branch as well as when we merge to main.

docs/source/conceptual-guides/endpoints.md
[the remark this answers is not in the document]
veekaybee: Nice. do we know if this then renders correctly on github?

veekaybee (approved):
Looks ok from the backend perspective. Thanks for taking this pass! Would love to get a few more eyeballs since this is a large one.

docs/source/conceptual-guides/endpoints.md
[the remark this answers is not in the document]
njbrake: Since we don't publish/host branch builds of the docs yet, I haven't been able to check it as a hosted github page.
However, I 
1. Built the docs locally and confirmed that the links rendered correctly
2. Looked at the github actions output logs to ensure that the GIT_COMMIT being parse by the action was the commit that I expected.

That seemed like a reasonable enough level of certainty to move forward. Once I merge I'll go through the hosted docs again just to check if there are any other anomalies.

.github/workflows/lumigator_pipeline.yaml
```diff
@@ -199,7 +199,7 @@ jobs:
           filters: |
             rebuild_fe:
               - '.github/workflows/**'
-              - 'lumigator/frontend/**'
+              - 'frontend/**'
```
macaab26, line 202: @njbrake if I'm not wrong you have to keep the lumigator here

njbrake: Great catch! Fixing 👍

macaab26 (approved):

javiermtorres (approved):

chainlink (approved):
Tested on kube, looks good

PR #725, opened 2025-01-23
HotReload for Jobs in Dev Mode

njbrake:
# What's changing

The Jobs folder needs to be watched by uvicorn so that it hot reloads when you make changes. Thanks @HareeshBahuleyan  for the find!

# I already...


HareeshBahuleyan (approved):
Thanks for the quick solution!

PR #726, opened 2025-01-23
Add warning about local-template env mismatch to Makefile

### LINKED ISSUE #723: [BUG]: Local env is out of sync with .env.template since we don't version control it
### Description

When anything in the .env variables changes, it doesn't reflect on the user's local `.env` because that file is private and not in version control, and we don't override it since we'd like to respect the user's preferences. 

We'd like to make sure that the user gets the latest from `.env.template` upon loading lumigator locally the first time with a check between the user's local .env file and the template in `make local-up`

### Reproduction

Use an env file different than what's in .env.template

### Relevant log output

```shell

```

### Expected behavior

I'd expect env to propogate correctly

### System Info

All OS

veekaybee:
# What's changing

When anything in the .env variables changes, it doesn't reflect on the user's local .env because that file is private and not in version control, and we don't override it since we'd like to respect the user's preferences.

We'd like to make sure that the user gets the latest from .env.template upon loading lumigator locally the first time with a check between the user's local .env file and the template in make local-up

* Updated section...
* Added new code to makefile `local-up` taking advantage of current makefile check 

Screenshot 2025-01-23 at 11 00 14 AM

[Closes #723](https://github.com/mozilla-ai/lumigator/issues/723)

# How to test it

run `make local-up` and you should see the warning

# I already...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,11 @@ DEV_DOCKER_COMPOSE_FILE:= .devcontainer/docker-compose.override.yaml
 
 check-dot-env:
 	@if [ ! -f .env ]; then cp .env.template .env; echo ".env created from .env.template"; fi
+	 @if [ -f .env ] && [ -f .env.template ] && ! cmp -s .env .env.template; then \
+      echo -e "\033[1;31m=====================================================================\033[0m"; \
+      echo -e "\033[1;31mWARNING: .env file differs from .env.template. Please review to stay in sync.\033[0m"; \
+      echo -e "\033[1;31m=====================================================================\033[0m"; \
+  fi
 
 # Launches Lumigator in 'development' mode (all services running locally, code mounted in)
 local-up: check-dot-env


Makefile
```diff
@@ -82,6 +82,11 @@ DEV_DOCKER_COMPOSE_FILE:= .devcontainer/docker-compose.override.yaml
 
 check-dot-env:
 	@if [ ! -f .env ]; then cp .env.template .env; echo ".env created from .env.template"; fi
+	 @if [ -f .env ] && [ -f .env.template ] && ! cmp -s .env .env.template; then \
```
aittalam, line 85: I think the semantics here is "check if `.env` has changed from the default" rather than "check if `.env.template` has been updated". Is that correct?

The result is that if someone customizes `.env` they will always see this message appear and won't notice the difference when `.env.template` is updated due to a pull.

WDYT about something like:

```
@if [ ! -f .env ]; then cp .env.template .env; diff .env.template .env > .env.diff; echo ".env created from .env.template"; fi
diff .env.template .env > .env.diff.new
@if [ -f .env ] && [ -f .env.template ] && ! cmp -s .env.diff .env.diff.new; then \
...
```
(I guess there's an `else` there to save the .new file into the .diff file too...)

Makefile
```diff
@@ -82,6 +82,11 @@ DEV_DOCKER_COMPOSE_FILE:= .devcontainer/docker-compose.override.yaml
...
+      echo -e "\033[1;31m=====================================================================\033[0m"; \
+      echo -e "\033[1;31mWARNING: .env file differs from .env.template. Please review to stay in sync.\033[0m"; \
```
aittalam, line 87: If we manage to do the check above, we can say something like:
```
.env and/or .env.template have been updated. Please review changes to stay in sync
```

(and if changes are ok we should also make sure the diff file is updated?)

Makefile
veekaybee: I can't think of a much more elegant way to do this (not saying that this is inelegant, just that it should be handled at the application level, so here makes sense for now!)

Makefile
```diff
@@ -81,7 +81,24 @@ LOCAL_DOCKERCOMPOSE_FILE:= docker-compose.yaml
...
+
+	# Check if files are out of sync and show warning
+	@if [ -f .env ] && [ -f .env.template ] && ! cmp -s .env.diff .env.diff.new; then \
+	echo -e "\033[1;31m====================================================================\033[0m"; \
+	echo -e "\033[1;31mWARNING: .env and .env.template are out of sync. Please review changes\033[0m"; \
+	echo -e "\033[1;31m====================================================================\033[0m"; \
+	fi
+
+	# Update diff file for next comparison
+	@mv .env.diff.new .env.diff 2>/dev/null || true
```
aittalam, line 101:
```suggestion
#   Create .env from template if it doesn't exist
	@if [ ! -f .env ]; then \
	cp .env.template .env; \
	echo ".env created from .env.template"; \
	fi

#   Generate new diff between template and current .env
	@diff .env.template .env > .env.diff.new 2>/dev/null || true

#   Check if files are out of sync and show warning
	@if [ -f .env ] && [ -f .env.template ] && ! cmp -s .env.diff .env.diff.new; then \
	echo -e "\033[1;31m====================================================================\033[0m"; \
	echo -e "\033[1;31mWARNING: .env and .env.template are out of sync. Please review changes\033[0m"; \
	echo -e "\033[1;31m====================================================================\033[0m"; \
	fi

#   Update diff file for next comparison
	@mv .env.diff.new .env.diff 2>/dev/null || true```

**Review by aittalam — APPROVED** [review]
Thank you!
I left a minor comment, good to merge for me after it is addressed!

File: Makefile [hunk 4]
```diff
@@ -81,7 +81,24 @@ LOCAL_DOCKERCOMPOSE_FILE:= docker-compose.yaml
 DEV_DOCKER_COMPOSE_FILE:= .devcontainer/docker-compose.override.yaml
 
 check-dot-env:
-	@if [ ! -f .env ]; then cp .env.template .env; echo ".env created from .env.template"; fi
+   # Create .env from template if it doesn't exist
```
aittalam, line 84: nit: the indented comments are shown on the terminal (at least in zsh), I'd suggest unindenting them e.g.

```suggestion
#   Create .env from template if it doesn't exist
```

Makefile
```diff
@@ -81,7 +81,24 @@ LOCAL_DOCKERCOMPOSE_FILE:= docker-compose.yaml
...
+	@if [ ! -f .env ]; then \
+	cp .env.template .env; \
+	echo ".env created from .env.template"; \
+	fi
+
+	# Generate new diff between template and current .env
+	@diff .env.template .env > .env.diff.new 2>/dev/null || true
```
aittalam, line 91: nit: I just realised `.env.diff` and `.env.diff.new` will appear as tracked files, we might want to add them to `.gitignore`

veekaybee: ah good catch!

PR #728, opened 2025-01-23
Create scaffolding for "workflows", merging experiments_new into experiments, hiding routes from OpenAPI schema that are in development

njbrake:
The goal of this PR is to re-organize and scaffold the backend in such a way to support the implementation of the features needed for workflow management. It does not solve any of the problems we know need to be worked on in the code,  but re-organizes things so that the missing pieces are clear and can be divided+conquered.

1. Hides from OpenAPI documentation the routes still in development (by setting `include_in_schema=False`)
2. Moves experiment based logic out of `experiments_new` and into the `experiments` route under the `/new` prefix (endpoints still hidden from OpenAPI schema to avoid confusing people looking at swagger docs etc)
3. Makes route called `workflows` which handles triggering a collection of jobs (currently inference + eval_lite) and having it be associated with an experiment
4. Create unimplemented routes to set groundwork for what remains to be implemented.
5. Re-organize integration test to follow the new steps of first creating an experiment (which is just a quick process to create an experiment id which is inserted into a DB), and then creating a workflow associated with that experiment.


lumigator/python/mzai/backend/backend/api/routes/runs.py
```diff
@@ -0,0 +1,61 @@
...
+from backend.api.deps import ExperimentServiceDep, JobServiceDep
+
+router = APIRouter()
+
+
+@router.post("/", status_code=status.HTTP_201_CREATED)
+async def create_run(
+    service: ExperimentServiceDep, request: ExperimentCreate, background_tasks: BackgroundTasks
+) -> ExperimentResponse:
+    return ExperimentResponse.model_validate(service.create_experiment(request, background_tasks))
```
njbrake, line 25: What we currently call 'create-experiment` is really a merging of `create-experiment` (aka make new experiment id and store in db) and `trigger-run` (aka trigger inference job and then evaluation job)

We would want to only call something like `create_run` here, which would be a new component in the service layer.

lumigator/python/mzai/backend/backend/api/deps.py
```diff
@@ -72,6 +73,17 @@ def get_experiment_service(
 ExperimentServiceDep = Annotated[ExperimentService, Depends(get_experiment_service)]
```
aittalam, line 73: Q: as the old `/experiments` only relied on `JobServiceDep` and the new `/workflows` relies on `WorkflowServiceDep`, does the new `ExperimentServiceDep` still need to rely on a job service and a dataset service?

```
def get_experiment_service(
    session: DBSessionDep, job_service: JobServiceDep, dataset_service: DatasetServiceDep
) -> ExperimentService:
```

Perhaps we could remove those deps already, making sure there's no stuff we do not need anymore.  WDYT?

lumigator/python/mzai/backend/backend/services/experiments.py
aittalam, line 1: I think there might be some dangling code here (ie. code that has been moved to the WorkflowService and is not reachable here, eg.  [run_eval](https://github.com/mozilla-ai/lumigator/blob/2c960962c365d72e0714a16333884f0f209d214e/lumigator/python/mzai/backend/backend/services/experiments.py#L81)). Would you mind removing that too?

lumigator/python/mzai/schemas/lumigator_schemas/workflows.py
```diff
@@ -0,0 +1,46 @@
...
+class WorkflowSummaryResponse(BaseModel, from_attributes=True):
+    workflow_id: UUID
+    experiment_id: UUID
+    run_ids: list[UUID]
+    metrics: dict
+    artifacts: dict
+    parameters: dict
+
+
+class WorkflowDetailsResponse(BaseModel):
```
aittalam, line 40: One reason we originally thought about returning a link to a downloadable file instead of directly a json from the API was to delegate the management of possibly very large files to some other service (in this case the object storage), while leaving the API (which can potentially run on low-resources hardware) free to deal with other requests.

I realise it is a bit overkill for the amount of data we had to deal with up to now, and it added an overhead for the SDK/UI (i.e. the need to manage an extra step for downloading data).

What is your take on this? What do you think are the pros/cons of one approach vs the other? I am in favor of simplifying things - and we could also have both if we wanted (e.g. /summaries and /summaries/download)

lumigator/python/mzai/backend/backend/api/routes/experiments.py
```diff
@@ -60,3 +61,54 @@ def get_experiment_result_download(
...
+
+####################################################################################################
+# "new" routes
+####################################################################################################
+# These "experiments_new" routes are not yet ready to be exposed in the OpenAPI schema,
+# because it is designed for the API when 'workflows' are supported
+# TODO: Eventually this route will become the / route,
+# but right now it is a placeholder while we build up the Workflows routes
+# It's not included in the OpenAPI schema for now so it's not visible in the docs
+@router.post("/new", status_code=status.HTTP_201_CREATED, include_in_schema=False)
```
javiermtorres, line 74: I proposed using the api version prefix to refer to the new endpoints. I'd prefer that to adding more endpoints that later need to be retrofitted.

lumigator/python/mzai/backend/backend/api/routes/experiments.py
```diff
@@ -60,3 +61,54 @@ def get_experiment_result_download(
...
+def create_experiment_id(
+    service: ExperimentServiceDep, request: ExperimentIdCreate
+) -> ExperimentResponse:
+    """Create an experiment ID."""
+    return service.create_experiment(request)
+
+
+@router.get("/new/all", include_in_schema=False)
```
javiermtorres, line 82: A general get should return all the applicable objects
```suggestion
@router.get("/new", include_in_schema=False)
```

lumigator/python/mzai/backend/backend/api/routes/workflows.py
```diff
@@ -0,0 +1,67 @@
...
+
+router = APIRouter()
+
+
+@router.post("/", status_code=status.HTTP_201_CREATED)
+async def create_workflow(
+    service: WorkflowServiceDep, request: WorkflowCreate, background_tasks: BackgroundTasks
+) -> WorkflowResponse:
+    """A workflow is a single execution for an experiment.
+    A workflow can be a single job or a collection of jobs.
```
javiermtorres, line 21: Why not always a collection of jobs? A single job would be a list of 1 element.

lumigator/python/mzai/schemas/lumigator_schemas/workflows.py
```diff
@@ -0,0 +1,46 @@
+import datetime
+from uuid import UUID
+
+from pydantic import BaseModel
+
+from lumigator_schemas.jobs import JobStatus
+
+
+class WorkflowCreate(BaseModel):
```
javiermtorres, line 9: I assume this would only be a first step, right? The logical replacement would be to include a list of jobcreates to be run in sequence.

lumigator/python/mzai/schemas/lumigator_schemas/workflows.py
javiermtorres: I'd just use a single data structure for outputs. These outputs can include dataset uuids or URLs; by using "links" or "ids" in all fields, the output will always be kept small. And retrieving the result itself is not such a big deal for the client.

lumigator/python/mzai/schemas/lumigator_schemas/workflows.py
njbrake: Yep absolutely! This will be subject to further changes as we migrate the frontend/sdk applications to use workflows.

lumigator/python/mzai/schemas/lumigator_schemas/workflows.py
njbrake: @aittalam this is a good point, I agree with you, I'll refactor this so that it's designed for links/ids.

@javiermtorres I agree with you, I'll give a try at restructuring this to match what you're expecting, please feel free to suggest some changes if I didn't capture your suggestion correctly.

lumigator/python/mzai/backend/backend/api/deps.py
njbrake: ooh great catch! We don't use the job_repo or experiment_repo anymore. However, we do still use the dataset_service and job_service repos to retrieve the dataset created by the inference job, and to check whether a job has been completed. I'll update to remove the deps we don't need 👍

lumigator/python/mzai/backend/backend/api/routes/experiments.py
njbrake: tl;dr I agree with you, but in this case I'm carrying forward an existing decision that was made, this is not a new issue i'm introducing here (I'm just not fixing the naming issue that already exists 😅 )

I thought about that, but in this case I was just moving this route out of "experiments_new" into the the "experiments" route, since these routes were the only routes from experiments_new that still belong in experiments after this change.

I agree with you that a better version prefixing process should be used next time. In this instance, we will need to retrofit the experiments endpoint in a breaking way, since "creating an experiment" will no longer actually trigger a job, it will just be creating an id.

Once the workflows routes are implemented with mlflow storage/management, I assume we'll do a PR where we swap in the new endpoints to replace the existing experiments endpoints, and update the SDK and frontend to use the new experiments + workflows strategy.

lumigator/python/mzai/backend/backend/api/routes/experiments.py
njbrake: Funny you should mention that here 😆 , so I had that originally, but apparently the way FastAPI works is such that if you have a `get(/new `and` get(/new/{experiments_id}` endpoint, it assumes that you always are calling that second route. It's a different FastAPI behavior from if you have `get(/` and `get(/{experiments_id}` which differentiates fine. 🤷 

Since these routes are really just hidden routes that will be moved to replace the current experiments endpoints hopefully soon, I didn't spend too much time worrying about these route names in particular. And they're hidden from the schema so that people browsing the swagger/openAPI docs will not see them.

lumigator/python/mzai/backend/backend/api/routes/workflows.py
njbrake: Ahhh great point! That is what I was thinking too, but my description is poor 😆 . I'll fix to clarify.

njbrake (commented):
Thank you for the helpful comments!

lumigator/python/mzai/backend/backend/services/experiments.py
```diff
@@ -33,6 +32,9 @@ def __init__(
         self._job_service = job_service
         self._dataset_service = dataset_service
 
+    def _raise_not_found(self, job_id: UUID):
+        raise HTTPException(status.HTTP_404_NOT_FOUND, f"Job {job_id} not found.")
+
```
peteski22, line 37:
```suggestion
```

We don't need this now, instead just raise a `JobNotFoundError` or `ExperimentNotFoundError`. 😄

lumigator/python/mzai/backend/backend/api/routes/experiments.py
```diff
@@ -12,11 +14,19 @@
...
 
-from backend.api.deps import JobServiceDep
+from backend.api.deps import ExperimentServiceDep, JobServiceDep
+from backend.services.exceptions.base_exceptions import ServiceError
+from backend.services.exceptions.experiment_exceptions import ExperimentNotFoundError
 
 router = APIRouter()
 
 
+def experiment_exception_mappings() -> dict[type[ServiceError], HTTPStatus]:
```
veekaybee, line 24: if we're returning a specific type here, makes sense to implement as Pydantic class directly and dump to dict for validation, wdyt?

lumigator/python/mzai/backend/backend/api/routes/experiments.py
```diff
@@ -60,3 +70,54 @@ def get_experiment_result_download(
...
+# These "experiments_new" routes are not yet ready to be exposed in the OpenAPI schema,
+# because it is designed for the API when 'workflows' are supported
+# TODO: Eventually this route will become the / route,
+# but right now it is a placeholder while we build up the Workflows routes
+# It's not included in the OpenAPI schema for now so it's not visible in the docs
+@router.post("/new", status_code=status.HTTP_201_CREATED, include_in_schema=False)
+def create_experiment_id(
+    service: ExperimentServiceDep, request: ExperimentIdCreate
+) -> ExperimentResponse:
+    """Create an experiment ID."""
```
veekaybee, line 87: nit: might want to also mention it's a UUID

veekaybee: we'll also want to validate the experiment response as we do in other methods

lumigator/python/mzai/backend/backend/api/routes/experiments.py
```diff
@@ -60,3 +70,54 @@ def get_experiment_result_download(
...
+    """Get an experiment by ID."""
+    return ExperimentResponse.model_validate(service.get_experiment(experiment_id).model_dump())
+
+
+@router.get("/new/{experiment_id}/jobs", include_in_schema=False)
+def get_experiment_jobs(
+    service: ExperimentServiceDep, experiment_id: UUID
+) -> ListingResponse[UUID]:
+    """Get all jobs associated with an experiment."""
+    return service.get_all_owned_jobs(experiment_id)
```
veekaybee, line 114: the method could be something a bit clearer like `get_all_jobs_by_experiment`

lumigator/python/mzai/backend/backend/api/routes/experiments.py
njbrake: This was something that was put in from https://github.com/mozilla-ai/lumigator/pull/754 so it's a question to @peteski22, I'll let him address it via a separate PR 👍

lumigator/python/mzai/backend/backend/api/routes/experiments.py
njbrake: Totally! it will actually be a str, because mlflow doesn't use UUIDs, per the PR that comes after this one. https://github.com/mozilla-ai/lumigator/pull/750/files#r1930972704

I'll add the model validation wrapper 👍

lumigator/python/mzai/backend/backend/api/routes/experiments.py
```diff
@@ -1,8 +1,10 @@
+from http import HTTPStatus
```
veekaybee, line 1: my understanding was that we were completely removing the current `experiments` route in favor of `experiments_new` and renaming _that_ to `experiments` - don't seem like we're going with that approach here

lumigator/python/mzai/backend/backend/api/routes/experiments.py
njbrake: Totally. I copied the naming from the way the function was currently named in the codebase, so I'll keep it this way for this PR but will update in a future PR 👍

lumigator/python/mzai/backend/backend/services/workflows.py
```diff
@@ -0,0 +1,166 @@
...
+    ):
+        self._job_service = job_service
+        self._dataset_service = dataset_service
+
+    NON_TERMINAL_STATUS = [
+        JobStatus.CREATED.value,
+        JobStatus.PENDING.value,
+        JobStatus.RUNNING.value,
+    ]
+    """list: A list of non-terminal job statuses."""
```
veekaybee, line 34: don't think we need this comment, or let's move above the variable for consistency

lumigator/python/mzai/backend/backend/services/workflows.py
```diff
@@ -0,0 +1,166 @@
...
+
+    # TODO: rely on https://github.com/ray-project/ray/blob/7c2a200ef84f17418666dad43017a82f782596a3/python/ray/dashboard/modules/job/common.py#L53
+    TERMINAL_STATUS = [JobStatus.FAILED.value, JobStatus.SUCCEEDED.value]
```
veekaybee, line 37: would make sense to define these at the pydantic/schema level since we use them throughout jobs and experiments

lumigator/python/mzai/backend/backend/services/workflows.py
```diff
@@ -0,0 +1,166 @@
...
+            await asyncio.sleep(5)
+            job_status = self._job_service.get_upstream_job_status(job_id)
+
+        match job_status:
+            case JobStatus.FAILED.value:
+                loguru.logger.error(f"Job {job_id} failed: not running task {str(task)}")
+            case JobStatus.SUCCEEDED.value:
+                loguru.logger.info(f"Job {job_id} finished successfully.")
+                if task is not None:
+                    task(*args)
```
veekaybee, line 61: would be helpful to leave a comment here understanding what args could potentially be passed, I'm always wary of args and kwargs in python since they lead to tricky debugging in the case of collected elements 😅

lumigator/python/mzai/backend/backend/services/workflows.py
```diff
@@ -0,0 +1,166 @@
...
+        # use the inference job id to recover the dataset record
+        dataset_record = self._dataset_service._get_dataset_record_by_job_id(inference_job_id)
+
+        # prepare the inputs for the evaluation job and pass the id of the new dataset
+        job_eval_dict = {
+            "name": f"{request.name}-evaluation",
+            "model": request.model,
+            "dataset": dataset_record.id,
+            "max_samples": request.max_samples,
+            "skip_inference": True,
```
veekaybee, line 86: do we still need skip_inference here since we separate out inference and eval?

lumigator/python/mzai/backend/backend/services/workflows.py
```diff
@@ -0,0 +1,166 @@
...
+            self._run_eval,
+            job_response.id,
+            request,
+            background_tasks,
+            request.experiment_id,
+        )
+
+        # TODO create a new workflow object which will be stored in
+        # by the tracking service (aka mlflow)
+        created_at = datetime.datetime.now()
```
veekaybee, line 146: nit, but may want to specify `datetime.datetime.utcnow()` for consistency across machines, for debugging, etc

lumigator/python/mzai/backend/backend/main.py
```diff
@@ -15,7 +15,7 @@
 from backend.api.router import api_router
 from backend.api.routes.completions import completion_exception_mappings
 from backend.api.routes.datasets import dataset_exception_mappings
-from backend.api.routes.experiments_new import experiment_exception_mappings
+from backend.api.routes.experiments import experiment_exception_mappings
```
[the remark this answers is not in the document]
njbrake: Similar to my earlier comment, this is changes related tohttps://github.com/mozilla-ai/lumigator/pull/754 so I'll defer to @peteski22, maybe a subject for a different PR

lumigator/python/mzai/backend/backend/services/workflows.py
njbrake: This is a copy past from some previous work, not anything new I added

lumigator/python/mzai/backend/backend/services/workflows.py
njbrake: Probably, but it's a copy-paste of existing code in our main. I'm not against changing it but I'd rather do it via a separate PR

lumigator/python/mzai/backend/backend/services/workflows.py
njbrake: Totally! This is existing code and I also have my eye on refactoring it, but would like to defer this work to a later PR.

lumigator/python/mzai/backend/backend/services/workflows.py
njbrake: Maybe not? Sorry to keep deferring stuff via my comments but this is existing code that I moved but didn't rewrite. I have thoughts about restructuring it but I would like to defer that to a future PR. This PR is focused on the re-organization effort to put things in locations to accomodate future work. I'm keeping all these comments in mind though so definitely good feedback for future!

lumigator/python/mzai/backend/backend/services/workflows.py
njbrake: Yeah this was really just a placeholder for this PR, it's going away in the next one https://github.com/mozilla-ai/lumigator/pull/750

lumigator/python/mzai/backend/backend/api/routes/experiments.py
njbrake: Correct. After a lot of discussion with Davide and Javier and Irina, we're migrating experiments to really just be a "container" not an executor, which means that experiments_new really shouldn't become experiments, because it is still functioning as an executor. With this goal in mind we split experiments_new into two chunks: one chunk goes into experiments, and is about creating an experiment_id and getting experiments. The second chunk goes into the new service/route callled "Workflows" which is in charge of triggering and managing composite jobs, and associating them with experiments via the use of a tracking service, which is coming in https://github.com/mozilla-ai/lumigator/pull/750

lumigator/python/mzai/backend/backend/services/workflows.py
veekaybee: no worries!

lumigator/python/mzai/backend/backend/api/routes/experiments.py
veekaybee: got it, thanks for the explanation!

lumigator/python/mzai/backend/backend/api/routes/experiments.py
peteski22: It was intended to allow each route 'container' (datasets, experiments, jobs, completions) that had a corresponding service, to specify mappings between service-level exceptions/errors and HTTP status codes that should be returned to the caller. 

It might have appeared cleaner if FastAPI allowed exception handler annotations (like `@router.exception_handler(JobNotFoundError)`) on the functions for `APIRouter`, as we wouldn't have to have this quasi-pattern where each route has to have a func to return the mappings and we add-em-all in `main`.

All this to say, I think it might become clearer how this can be improved now all the 'bits' are in, but I'm not sure I understand how making it a Pydantic class (in the schemas package) would improve things. If it's clear to you folks, please just chalk up an issue and we can look into it. 😄

lumigator/python/mzai/backend/backend/services/workflows.py
peteski22: It can be updated to make it clearer for sure. But, I'd leave it in and where it is, as it needs to be below the class attribute so that IDEs etc. pick up the docs to show in IntelliSense (not sure what the generic term is).

image

lumigator/python/mzai/backend/backend/api/routes/experiments.py
javiermtorres: There should be a way to structure this properly (possibly using the top level router). Please include a FIXME so we don't forget to do it in a later PR. I don't think we should block on this, but instead fix it in parallel.

njbrake: Will do 👍

lumigator/python/mzai/schemas/lumigator_schemas/workflows.py
```diff
@@ -0,0 +1,51 @@
...
+    artifacts: dict
+    parameters: dict
+
+
+# TODO: This schema will need to be refined when the WorkflowDetails route is implemented
+class WorkflowDetailsResponse(BaseModel):
+    workflow_id: UUID
+    experiment_id: UUID
+    job_ids: list[UUID]
+    metrics_urls: list[str]  # same length as run_ids
```
javiermtorres, line 45: Same as run_ids? Do you mean job_ids?
If so, please make a separate schema holding `JobResults = (job_id, metric_url, artifacts_url)` so we don't depend on checking the lengths of separate fields.

lumigator/python/mzai/backend/backend/api/routes/experiments.py
njbrake:
```suggestion
# TODO: FIXME this should not need the /all suffix.
# See further discussion https://github.com/mozilla-ai/lumigator/pull/728/files/2c960962c365d72e0714a16333884f0f209d214e#r1932176937
@router.get("/new/all", include_in_schema=False)
```

lumigator/python/mzai/schemas/lumigator_schemas/workflows.py
njbrake: Great point, I'll refactor.

lumigator/python/mzai/schemas/lumigator_schemas/workflows.py
```diff
@@ -0,0 +1,51 @@
...
+    name: str
+    description: str
+    status: JobStatus
+    created_at: datetime.datetime
+    updated_at: datetime.datetime | None = None
+
+
+class WorkflowSummaryResponse(BaseModel, from_attributes=True):
+    workflow_id: UUID
+    experiment_id: UUID
```
aittalam, line 33: If `experiment_id` is something we get at workflow creation time, I think perhaps we should add it to `WorkflowResponse` too (or as an alternative to here), wdyt?
I agree abt job_ids instead, as (at least for now) we do not have them all at workflow creation time so the only place where we can get them is either by querying jobs directly (when the workflow is live) or by getting the aggregated data here after the workflow completes

njbrake: Agree, I think that makes good sense so that the client can verify that the workflow was created for the same experiment_id that they requested it be associated with (since the experiment_id is required in the WorkflowCreate model). I'll update.

lumigator/python/mzai/schemas/lumigator_schemas/workflows.py
aittalam: Related question: now that introducing URLs the only difference wrt this schema and the summaries schema is the urls instead of some larger data, do we really need two separate responses/endpoints? Perhaps we can have everything in the same response as we already have all that data anyway, WDYT?

aittalam (approved):
Thanks Nathan! I think this is good to go after syncing with main (if needed) and closing the last open convos

PR #738, opened 2025-01-24
Bug fix for annotation/inference endpoint fails with long sequences as input

### LINKED ISSUE #667: [BUG]: Annotation with large texts fails
### Description

If we try to generate ground truth for datasets that contains large columns of text it currently fails. @aittalam suggests [this](https://stackoverflow.com/questions/67849833/how-to-truncate-input-in-the-huggingface-pipeline) might help

### Reproduction

Call the annotation endpoint with` thunderbird.csv` dataset

```
{
   dataset: "7c74038e-43c7-4527-842a-6252fd8db446" // this is the ID of thunderbird.csv dataset in my local db
   description: "Groundtruth generation for dataset 7c74038e-43c7-4527-842a-6252fd8db446"
   max_samples: -1
   name: "Groundtruth for thunderbird.csv"
}
```

### Relevant log output

```shell
Job entrypoint command failed with exit code 1, last available logs (truncated to 20,000 chars): File "/tmp/ray/session_2025-01-17_02-13-42_842760_1/runtime_resources/pip/617697551215d8488f42ccab087d2364573ba842/virtualenv/lib/python3.11/site-packages/transformers/models/bart/modeling_bart.py", line 116, in forward return super().forward(positions + self.offset) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/tmp/ray/session_2025-01-17_02-13-42_842760_1/runtime_resources/pip/617697551215d8488f42ccab087d2364573ba842/virtualenv/lib/python3.11/site-packages/torch/nn/modules/sparse.py", line 190, in forward return F.embedding( ^^^^^^^^^^^^ File "/tmp/ray/session_2025-01-17_02-13-42_842760_1/runtime_resources/pip/617697551215d8488f42ccab087d2364573ba842/virtualenv/lib/python3.11/site-packages/torch/nn/functional.py", line 2551, in embedding return torch.embedding(weight, input, padding_idx, scale_grad_by_freq, sparse) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ IndexError: index out of range in self
```

### Expected behavior

It should generate a groundtruth 

### System Info

MacOS latest version

HareeshBahuleyan:
# What's changing
When testing with Thunderbird dataset, @agpituk found that BART model fails for long sequences with the error:
```
IndexError: index out of range in self
```
As pointed out by @dpoulopoulos , the error comes from a limitation on the maximum number of positional embeddings that the model can have (token embeddings will always be within range).

* This could be fixed by setting the allowed `model_max_length` (the number of tokens that the model sees) for the respective model, e.g. [for BART](https://huggingface.co/facebook/bart-large-cnn/blob/main/config.json#L44) this is 1024 

* Additionally, one would also need to set the `truncation = True` in the HF pipeline to truncate the sequence at this max_length.

See [Colab Notebook](https://colab.research.google.com/drive/1ndu9U1ayy5KEXl-z1AZkd_8irb9W1hWN?usp=sharing) for a demo with these two changes that fixes the issue for long text input.

Closes #667 

# How to test it

Steps to test the changes:

1. Upload `thunderbird.csv` as the dataset
2. Run annotate endpoint for this dataset (uses BART by default)
```
{
  "name": "default annotation with BART",
  "dataset": "<dataset_id>",
  "max_samples": -1,
  "task": "summarization",
  "store_to_dataset": true
}
```

# I already...


lumigator/python/mzai/sdk/tests/integration/test_scenarios.py
```diff
@@ -131,24 +132,36 @@ def test_job_lifecycle_remote_ok(lumi_client_int, dialog_data, simple_eval_templ
     requests.get(download_info.download_url, allow_redirects=True)
 
 
-def test_annotate_dataset(lumi_client_int, dialog_data_unannotated):
+@pytest.mark.parametrize(
+    "dataset_name, dataset_fixture",
+    [("dialog", "dialog_data_unannotated"), ("thunderbird", "thunderbird_data_unannotated")],
```
HareeshBahuleyan, line 137: I guess we could skip `dialog_data_unannotated` from test_annotate_datasets, since `thunderbird_data_unannotated` tests unannotated scenario as well as long text sequences.

lumigator/python/mzai/jobs/inference/model_clients.py
```diff
@@ -150,7 +150,16 @@ def _chat_completion(
 
 class HuggingFaceModelClient(BaseModelClient):
     def __init__(self, config: InferenceJobConfig):
-        self._pipeline = pipeline(**config.hf_pipeline.model_dump())
+        pipeline_kwargs = config.hf_pipeline.model_dump()
+        if hasattr(config.hf_pipeline, "tokenizer") and config.hf_pipeline.tokenizer is not None:
```
javiermtorres, line 154: Maybe this is simpler, if no other falsy objects could be stored here:

```suggestion
        if hasattr(config.hf_pipeline, "tokenizer") and config.hf_pipeline.tokenizer:
```

lumigator/python/mzai/sdk/tests/integration/test_scenarios.py
javiermtorres: I'd leave both, so if the length handling alone fails we'll know about it.

lumigator/python/mzai/backend/backend/config_templates.py
```diff
@@ -154,6 +172,7 @@
 templates = {
     JobType.INFERENCE: {
         "default": default_infer_template,
+        "hf://facebook/bart-large-cnn": bart_infer_template,
```
javiermtorres, line 175: Will this work for a standalone inference?

lumigator/python/mzai/backend/backend/config_templates.py
```diff
@@ -111,14 +111,32 @@
...
+        "model_uri": "{model_uri}",
+        "task": "{task}",
+        "accelerator": "{accelerator}",
+        "revision": "{revision}",
+        "use_fast": "{use_fast}",
+        "trust_remote_code": "{trust_remote_code}",
+        "torch_dtype": "{torch_dtype}",
+        "max_length": 124,
+        "tokenizer": {{ "model_uri": "{model_uri}",
+                        "model_max_length": 1024,
```
dpoulopoulos, line 130: Can we make this dynamic? I don't even remember why we had this hardcoded in the temple in the first place. Also, if we make this dynamic, is there a way to limit its value, or log a warning if a user specifies something more than `1024`?

lumigator/python/mzai/jobs/inference/inference_config.py
```diff
@@ -115,8 +116,24 @@ class SamplingParameters(BaseSamplingParameters):
     model_config = ConfigDict(extra="forbid")
 
 
+class AutoTokenizerConfig(BaseAutoTokenizerConfig, arbitrary_types_allowed=True):
+    """Settings passed to a HuggingFace AutoTokenizer instantiation."""
+
+    model_uri: AssetPath = Field(title="The Model HF Hub repo ID", exclude=True)
+    model_max_length: int
+    use_fast: bool = True
+    trust_remote_code: bool = False
```
dpoulopoulos, line 125: Should we add a `truncation` argument here, with the default set to `True`?

lumigator/python/mzai/backend/backend/config_templates.py
HareeshBahuleyan: So since this is a transformers library bug:
image
technically the bart tokenizer max length should default to 1024, as per the config: https://huggingface.co/facebook/bart-large-cnn/blob/main/config.json#L44

This seems to be a model specific issue, when I check it seems to be configured correctly for some other models:
image

So I think it should be read from some config, but not exposing it to the user. That config would be per model.

dpoulopoulos: Thanks for the colab Notebook! Very helpful! 👍 

Why are you saying that it is a transformers library bug? From the [documentation](https://huggingface.co/docs/transformers/v4.17.0/en/main_classes/tokenizer#transformers.PreTrainedTokenizerFast.model_max_length) I read:

> model_max_length (int, optional) — The maximum length (in number of tokens) for the inputs to the transformer model. When the tokenizer is loaded with [from_pretrained()](https://huggingface.co/docs/transformers/v4.17.0/en/internal/tokenization_utils#transformers.PreTrainedTokenizerBase.from_pretrained), this will be set to the value stored for the associated model in max_model_input_sizes (see above). If no value is provided, will default to VERY_LARGE_INTEGER (int(1e30)).

This is exactly the behavior I'm seeing in the Notebook. The `bart-large-cnn` model has no `max_model_input_sizes` attribute set, thus, it defaults to `VERY_LARGE_INTEGER`. On the other hand, the `pegasus-cnn_dailymail` model explicitly sets `model_max_length` to `1024` in [`tokenizer_config.json`](https://huggingface.co/google/pegasus-cnn_dailymail/blob/main/tokenizer_config.json).

So, I don't think this is their bug. We should somehow fix this on our side.

lumigator/python/mzai/jobs/inference/inference_config.py
dpoulopoulos: I would move this to the `AutoTokenizerConfig` class because it is a tokenizer-specific argument, isn't it? What do you think?

lumigator/python/mzai/backend/backend/config_templates.py
HareeshBahuleyan: > If no value is provided, will default to VERY_LARGE_INTEGER (int(1e30)).

Yeah, this tokenizer_config is missing from some models. What I meant is that this missing information causes differences in inference vs. training sequence lengths and hence throws errors. You are right this is not a transformers library bug, but rather an issue caused by missing configs from some of the contributed models.

aittalam: +1, it would be definitely useful if we had something that works more generally. One thing we can do is change this `"mod_max_length": 1024` to a parameter we then fill up with what we can get from one of the `typical_fields` (assuming those fields never coexist and have inconsistent information 😅)

lumigator/python/mzai/backend/backend/config_templates.py
```diff
@@ -111,14 +111,32 @@
...
     "dataset": {{ "path": "{dataset_path}" }},
+    "hf_pipeline": {{
+        "model_uri": "{model_uri}",
+        "task": "{task}",
+        "accelerator": "{accelerator}",
+        "revision": "{revision}",
+        "use_fast": "{use_fast}",
+        "trust_remote_code": "{trust_remote_code}",
+        "torch_dtype": "{torch_dtype}",
+        "max_length": 124,
```
aittalam, line 128:
```suggestion
        "max_length": 1024,
```

lumigator/python/mzai/sample_data/thunderbird_mini_no_gt.csv
aittalam, line 1: While this data is publicly available on Thunderbird's ml archives, it still contains personal information that needs to be dealt with in an appropriate way. For this reason, I'd suggest to test against synthetic samples or a public academic dataset containing no personal information.

lumigator/python/mzai/backend/backend/config_templates.py
HareeshBahuleyan: I think max_length refers to the max length of the generation (output) and model_max_length is for the input token length. Indeed very confusing!
https://huggingface.co/docs/transformers/en/main_classes/text_generation#transformers.GenerationConfig

But, it seems you picked up a typo, for bart max_length according to [HF config](https://huggingface.co/facebook/bart-large-cnn/blob/main/config.json#L59) is 142 not 124 :) Thanks!

lumigator/python/mzai/sample_data/thunderbird_mini_no_gt.csv
HareeshBahuleyan: Good callout, I checked with @ividal earlier and we discussed to have the data with only 2 rows (those containing the long input sequences for testing). Plus, I removed any email ids that appear in those rows.
Would that be sufficient?

HareeshBahuleyan:
@aittalam @dpoulopoulos FYI we are deleting this branch (to get rid of the usage of thunderbird dataset from history) and opening a new PR here:
https://github.com/mozilla-ai/lumigator/pull/762

PR #740, opened 2025-01-24
First attempt at a parametrized JobCreate

### LINKED ISSUE #706: Consolidate the different creation jobs into an organized class/package.
This issue comes from a comment on a PR about schemas between job and backend.

https://github.com/mozilla-ai/lumigator/pull/692/files/fa86422b1c33d503a786bdc1f8a3e4d840624afd#r1923204888
also
https://github.com/mozilla-ai/lumigator/pull/692/files/fa86422b1c33d503a786bdc1f8a3e4d840624afd#r1923229499

Basically, the different types of Jobs (JobEvalLiteCreate, JobInferenceCreate) could be organized with pydantic descriminators and classes so that when a new job is added, logic to handle the new job would need to be put in as few places as possible.

FYI @javiermtorres  since this issue comes out of our discussion about it on the PR.

javiermtorres:
# What's changing

The JobCreate schema is changed to include a separate specific job_config. The openapi produced includes a `oneOf` constraint:

```
      "JobCreate": {
        "properties": {
          "name": {
            "type": "string",
            "title": "Name"
          },
[...]
          "job_config": {
            "oneOf": [
              {
                "$ref": "#/components/schemas/JobEvalConfig"
              },
              {
                "$ref": "#/components/schemas/JobEvalLiteConfig"
              },
              {
                "$ref": "#/components/schemas/JobInferenceConfig"
              },
              {
                "$ref": "#/components/schemas/JobAnnotateConfig"
              }
            ],
            "title": "Job Config",
            "discriminator": {
              "propertyName": "job_type",
              "mapping": {
                "annotate": "#/components/schemas/JobAnnotateConfig",
                "eval_lite": "#/components/schemas/JobEvalLiteConfig",
                "evaluate": "#/components/schemas/JobEvalConfig",
                "inference": "#/components/schemas/JobInferenceConfig"
              }
            }
          }
        },
```

The jobs and experiments services are changed accordingly.

Closes #706 

# How to test it

Tests should run correctly.

N/A

# I already...

  - No DB migration needed


lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -58,12 +56,16 @@ def create_annotation_job(
...
     """
-    inference_job_create_request = JobInferenceCreate(
-        **job_create_request.dict(),
+    inference_job_create_config = JobAnnotateConfig(
+        **job_create_request.job_config.dict(),
         model="hf://facebook/bart-large-cnn",
         output_field="ground_truth",
     )
-    inference_job_create_request.store_to_dataset = True
+    inference_job_create_config.store_to_dataset = True
```
njbrake, line 64: Should this be moved inside the JobAnnotateConfig constructor one line abouve?

lumigator/python/mzai/backend/backend/api/routes/jobs.py
```diff
@@ -58,12 +56,16 @@ def create_annotation_job(
...
+    inference_job_create_request_dict = job_create_request.model_dump()
+    inference_job_create_request_dict.job_config = inference_job_create_config
+
+    inference_job_create_request = JobCreate(**inference_job_create_request_dict)
```
njbrake, line 68:
```suggestion
    inference_job_create_request = JobCreate(
        **job_create_request.model_dump(), 
        job_config=inference_job_create_config
    )
```

lumigator/python/mzai/backend/backend/services/jobs.py
```diff
@@ -249,7 +249,7 @@ def _validate_config(self, job_type: str, config_template: str, config_params: d
         else:
             loguru.logger.info(f"Validation for job type {job_type} not yet supported.")
 
-    def _get_job_params(self, job_type: str, record, request: BaseModel) -> dict:
+    def _get_job_params(self, job_type: str, record, request: JobCreate) -> dict:
```
njbrake, line 252: Can the logic of this function be extracted out of the jobs service too? idk, maybe some sort of computed property inside each Job type?

njbrake (commented):
Looking good so far! Made a code suggestion but looks like a very logical refactor. My only question would be whether you plan on addressing the custom logic of `_get_job_params` in this PR. If you don't plan on addressing it here, can you make a separate issue to track elevating that out of the service layer?

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: If we make the distinction of request and config as whether it's handled by the backend or the job, I'd keep it where it is. The storage of results into a dataset is actually the a for the generic backend-job interaction and not something to be handled in the job itself.

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: IIRC Pydantic complained because it found some keys twice (so no "key overwriting" was possible with this syntax). I'll check it out again.

lumigator/python/mzai/backend/backend/services/jobs.py
javiermtorres: I'd make a plain function returning or mutating some job_params pre-filled with values from the request itself (or maybe the other way around, but that's not so important right now). And yes, that's one of the purposes of this whole change. It will also make the config_templates unnecessary, but we may still need to take the model into account when mapping things around.

These functions can be attached to the job config class or the job module. I'll make them static functions in the job config class for the moment, but I think the module works better as a namespace.

javiermtorres:
The SDK needs to be updated. I have checked the backend unit and integration tests locally and they seem to work.

lumigator/python/mzai/backend/backend/services/jobs.py
javiermtorres: I have placed the equivalent logic in `generate_inference_job_config` and `generate_evaluation_lite_job_config`. They cannot be made purely computed properties of the model, because we have dependencies on the data record, the dataset path, and the storage path, plus some small logic getting the API URL for external services. Let's forget about this detail right now, and address where to put these functions and how to best pass the extra params in a separate follow up PR.

javiermtorres:
The SDK and notebook tests have been updated. @veekaybee @aittalam I've changed the code of the notebook slightly. One important difference is that I have removed the model param in the eval lite job. AFAICT, it's not needed there. The notebook takes it from the initial model spec in the notebook and **not** from the output of the summarization job. Since the output is a csv, it didn't make sense to put the model there, but I'll check the results metadata.

lumigator/python/mzai/backend/backend/api/routes/jobs.py
javiermtorres: Ok, it did work O_o Updating and pushing.

javiermtorres: Yes, basically I get at https://github.com/mozilla-ai/lumigator/actions/runs/13139121520/job/36661667194?pr=740#step:7:932 :

```
>       inference_job_create_request = JobCreate(
            **job_create_request.model_dump(), job_config=inference_job_create_config_dict
        )
E       TypeError: lumigator_schemas.jobs.JobCreate() got multiple values for keyword argument 'job_config'
```

So I'll leave it as it is.

lumigator/backend/backend/services/jobs.py
```diff
@@ -309,108 +305,133 @@ def _validate_config(self, job_type: str, config_template: str, config_params: d
...
+            dataset=IDatasetConfig(path=dataset_path),
+            job=InferJobConfig(
+                max_samples=request.max_samples,
+                storage_path=storage_path,
+                output_field=request.job_config.output_field,
+            ),
+        )
+        # Maybe use just the protocol to decide?
+        match request.job_config.model:
+            case "oai://gpt-4o-mini" | "oai://gpt-4o":
```
njbrake, line 327: I think eventually we will want to elevate this logic maybe into a separate class (something like InferenceManager) as we support more complex prompts and models. However, I think for the time being this refactor is moving in the right direction and totally makes sense 👍

lumigator/backend/backend/services/jobs.py
```diff
@@ -309,108 +305,133 @@ def _validate_config(self, job_type: str, config_template: str, config_params: d
...
+                return_predictions=True,
+                storage_path=storage_path,
+            ),
+        )
+        return job_config
 
     def create_job(
         self,
-        request: JobEvalCreate | JobEvalLiteCreate | JobInferenceCreate,
+        request: JobCreate,
```
njbrake, line 397: Is this function dropping support for JobEval? From looking at the logic a few lines lower it seems like only JobType.EVALUATION_LITE and JobType.INFERENCE are supported, which will break the frontend API which is currently only using EVALUATION and not EVALUATION_LITE + INFERENCE

lumigator/backend/backend/services/jobs.py
```diff
@@ -430,7 +451,7 @@ def create_job(
         settings.inherit_ray_env(runtime_env_vars)
 
         # set num_gpus per worker (zero if we are just hitting a service)
-        if not request.model.startswith("hf://"):
+        if job_type == JobType.INFERENCE and not request.job_config.model.startswith("hf://"):
             worker_gpus = job_settings["ray_worker_gpus_fraction"]
         else:
             worker_gpus = job_settings["ray_worker_gpus"]
```
njbrake, line 457: I don't think it needs to be done in this PR necessarily, but I wonder if this logic can be moved to generate_inference_job_config

njbrake (commented):
My concern is that this PR is dropping support for the JobType.EVALUATION, which is needed to support the current frontend design. I may misunderstand the code. Other than that, only minor comments. Thanks for the work on this! (Let me know about JobType.EVALUATION and then I'll approve once that's worked out).

lumigator/backend/backend/api/routes/jobs.py
```diff
@@ -72,12 +70,16 @@ def create_annotation_job(
...
+    inference_job_create_config_dict = job_create_request.job_config.dict()
+    inference_job_create_config_dict["model"] = "hf://facebook/bart-large-cnn"
+    inference_job_create_config_dict["output_field"] = "ground_truth"
+    inference_job_create_config_dict["store_to_dataset"] = True
+    inference_job_create_config_dict["job_type"] = JobType.INFERENCE
+
+    inference_job_create_request_dict = job_create_request.model_dump()
+    inference_job_create_request_dict["job_config"] = inference_job_create_config_dict
+
+    inference_job_create_request = JobCreate(**inference_job_create_request_dict)
```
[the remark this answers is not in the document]
javiermtorres: > Could we have some sort of warning if the key was already set to something else?

The idea would be to prevent setting any param that would not be acceptable for an Annotation job config. However, there are params like model that should perhaps be moved to the jobs config, since they don't deal with job lifecycle. Maybe as an interim measure we could log a warning until we settle on the schemas.

Conceptually, an annotation is a different kind of job, even though internally is also mapped to an inference (for example, a model cannot be set in an annotation since the annotation job is designed to use exclusively the BART model [please @ividal confirm]).

lumigator/backend/backend/services/jobs.py
javiermtorres: As commented, my idea could be that the InferJobConfig would be passed roughly transparently to the jobs (and then we can definitely use a base class on top of WhateverJobConfig to hold the job-backend interaction details like storage_path or dataset_path, since here we aren't concerned about the OpenAPI translation).

I'll resolve then this issue, please unresolve if there's any other aspect you'd like to discuss.

lumigator/backend/backend/services/jobs.py
javiermtorres: This would be part of the general separation between job specific config and job-backend interaction config, so I'd leave it for another PR.

lumigator/backend/backend/api/routes/jobs.py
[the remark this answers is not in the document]
ividal: TLDR: an annotation job is a constrained inference job (one where the user is not choosing model or parameters). We should at least be transparent about the constraints (from API to UI).

This is more a product decision than a technical one: Lumigator at this point is supposed to be opinionated about the model used to generate annotations. So an annotation job request should not make the user _believe_ they can set any model they like, since it would be quietly ignored (replaced currently by BART, tomorrow who knows).

That's why I'd worry about the API offering (as this PR introduces):

```
{
  "name": "string",
  ...
  "job_config": {
    "job_type": "evaluate", # ouch, minor gripe, but this is a default taken from swagger for an annotation job
    "model": "string", # misleading: this would be promptly ignored by the backend
    "model_url": "string", # same
    "system_prompt": "string",
    "skip_inference": false
  }
}
```

What do you think - is there any reasonable way to avoid the confusions mentioned above (specially related to the model)?

ividal (changes requested):

ividal:
Thanks for this! One note, @javiermtorres this PR should still be in sync with the UI and keep an eye on how it interacts with /experiments.

lumigator/backend/backend/api/routes/jobs.py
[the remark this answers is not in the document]
javiermtorres: Maybe I'm not getting this, but the annotation does not allow a model to be set:
https://github.com/mozilla-ai/lumigator/pull/740/files#diff-851bf090de2849773e9c7486bebd855b586c8a24f6b20cdd2eb10d65139dd1c2R100-R103

If we are talking about the standalone EVALUATE job, then we should remove this one ASAP and port the current usage to INFERENCE+EVAL_LITE.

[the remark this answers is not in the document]
ividal: > TLDR: an annotation job is a constrained inference job (one where the user is not choosing model 

and

> the annotation does not allow a model to be set:

are both correct. 

I was raising that when one opens the swagger UI, under POST `/annotate`, one sees the snippet I quoted above (notice "job_type" = evaluate, "model" and "model_uri" are there), which could lead to confusion.


image


Since most users will not use the raw API, but only either the FE or the UI, I wouldn't make this a blocker, but it might trip contributors (internal or external) down the line.

ividal (changes requested):
Just a note that uploading a file w/o ground-truth and then clicking on "generate ground-truth" generates a "job not found error".

See log:
[2025-02-10 - PR740.log](https://github.com/user-attachments/files/18732963/2025-02-10.-.PR740.log)

ividal (commented):
Tested #847 in the context of reviewing #740 . I've only looked at the backend code (740), not the frontend code. But on the testing front, I did a demo:
- uploading a dataset with and w/o gt
- annotating one w/o gt
- launching an experiment with both local and API-based models


image

Overall, works as expected (🥳 ). There are a number of smaller issues, but let's get 847 in here and then 740 into main  - and iterate on smaller separate issues :)

ividal (approved):
With #847 in here and checks green, we have ourselves a usable new experiments workflow :)

THANK YOU for the effort @javiermtorres @khaledosman

lumigator/backend/backend/services/jobs.py
ividal: The conclusion here is the UI is switching to the new experiments workflow (separate inference + pure evaluation jobs) in #847 .

Scope creep is a (major) thing. No matter when you read this.

Resolving this now.

PR #744, opened 2025-01-27
Add persistence to Ray

### LINKED ISSUE #690: DB is persistent but ray jobs aren't
### Motivation

After we added minio, our DB became persistent but our jobs in ray aren't. That means when we restart the containers we can see what's in the database but not what is in ray, causing some issues. We need to discuss how to approach this (should we have a volume for ray to make it persistent? should we make the api fault tolerant anyway?)

### Alternatives

_No response_

### Contribution

_No response_

javiermtorres:
# What's changing

A Redis container has been added to the compose definition to hold the Ray metadata.

Closes #690

# How to test it

Steps to test the changes:

1. Start as usual with `make local-up`
2. Run the backend integration tests with `make test-backend-integration`
3. Check the jobs in the dashboards
4. Stop the containers with ^C
5. Restart them with `make local-up`
6. Check the dashboard again. The jobs information should still be visible.

N/A

# I already...

  - N/A
  - No DB changes done


docker-compose.yaml
```diff
@@ -42,14 +42,30 @@ services:
     profiles:
       - local
 
+  redis:
+    image: redis:8.0-M03-alpine
+    command: redis-server --save 60 1 --loglevel warning
+    profiles:
+      - local
+    volumes:
+      - redis-data:/data
```
dpoulopoulos, line 51: Should we add a `make` target to allow the user to clean this, if they choose to?

dpoulopoulos (commented):
Looks good to me! Thanks for this! I left a minor comment.

docker-compose.yaml
javiermtorres: We have talked about using labels to prune the right containers. IIRC the volumes can get pruned too with a certain flag. I'll check open PRs to see how far we've done of this. IMO this volume should have the same life cycle as the ray and redis containers, so there should be no need to clean it up separately.

aittalam (approved):
Thanks @javiermtorres , I tested it and it works as expected! Just a few notes:

This PR:

- persists job information ✅ 
- does not persist job logs ❌ 
- does not persist the PIP environment (✅ as there's your other draft PR for that)
 
I want to raise a flag on job logs for the following reason: right now, the UI gets all the jobs which are saved in the DB BUT ALSO available on Ray. The reason for this choice was that job information was not persisted in the DB and some things we need (input configuration, logs) are available on ray only.

When this PR is merged, the UI will find jobs on Ray which do not have logs if they are also persisted in the DB. This will not happen until https://github.com/mozilla-ai/lumigator/pull/674 is merged so it is not a blocker, but I will add a link there to this PR just to double check that things work properly before merging that one.

Issue #525, opened 2024-12-18
Write a tracking interface with an MLflow implementation


javiermtorres:
### Motivation

Lumigator should be able to record its own data about experiments, but also export it to MLflow for more complex analysis and post processing. A simple implementation should be easily added to the experiments endpoint. This feature should be optional, so Lumigator could still work without an existing mlflow. The interface of this implementation should be general enough that it could be adapted later on with minimal changes to other systems like Neptune or Weights & Biases, but making a completely generic interface should not be the main concern.

### Contribution

The idea is to start a PR for this issue.

### Have you searched for similar issues before submitting this one?

- [x] Yes, I have searched for similar issues

PR #750, opened 2025-01-27
Issue 525: Tracking Interface with Mlflow

njbrake:
# What's changing

This PR builds on top of the architecture changes introduced in https://github.com/mozilla-ai/lumigator/pull/728. 

It introduces the MLflow Tracking Client with a layer of abstraction to support other tracking backends in the future. I implemented a few mlflow methods to be sure that I correctly understand how to use the mlflow library. I have code that integrates it into our service and dependency layer, but I'll put that into a different PR.

This pull request includes significant updates to integrate MLflow for tracking machine learning experiments and workflows. The changes span environment configuration, Docker setup, backend settings, and the implementation of a new tracking client interface.

This pull request introduces significant changes to integrate MLflow for experiment tracking and management. The changes include updates to environment configurations, new MLflow-related classes and methods, and modifications to existing schemas to support MLflow.

### MLflow Integration:

* **Environment Configuration:**
  * Added `MLFLOW_TRACKING_URI` and related AWS environment variables to `.env.template` and `.devcontainer/docker-compose.override.yaml`. 
  * Introduced a new `mlflow` service in the Docker Compose override file.

* **Backend Settings:**
  * Added `TrackingBackendType` enum and `TRACKING_BACKEND` configuration in `backend/settings.py`.
  * Implemented a computed property `TRACKING_BACKEND_URI` for determining the tracking URI based on the backend type.

* **Tracking Client Implementation:**
  * Created `MLflowTrackingClient` and `MLflowClientManager` classes in `backend/tracking/mlflow.py` to handle MLflow operations.
  * Added abstract base classes `TrackingClient` and `TrackingClientManager` in `backend/tracking/tracking_interface.py`.
  * Implemented `TrackingClientFactory` to create tracking client instances based on settings.
  * Updated `__init__.py` to include the new tracking client manager.

* **Schema Updates:**
  * Modified `WorkflowCreate`, `WorkflowResponse`, `WorkflowSummaryResponse`, and `WorkflowDetailsResponse` schemas to use `str` instead of `UUID` for IDs.

* **Dependencies:**
  * Added `mlflow` to the project dependencies in `pyproject.toml`. 

Closes https://github.com/mozilla-ai/lumigator/issues/525 

# How to test it

This PR does not add the tracking client into the dependency or server layer, so there are no changes to behavior. However, to confirm that mlflow works in the docker-compose, run

`make local-up`

# I already...


lumigator/python/mzai/schemas/lumigator_schemas/workflows.py
```diff
@@ -9,7 +9,7 @@
 class WorkflowCreate(BaseModel):
     name: str
     description: str = ""
-    experiment_id: UUID
+    experiment_id: str
```
njbrake, line 12: mlflow created id's (run/experiment) are returned by the mlflow client as strings, not UUIDs. e.g. an experiment id is the string `252773820672265330` and a run id is the string `6c56b97ad4b44c2cafff799bc0e56de0`

javiermtorres: Ok, so the format does not comply with the uuid representation, and not just a difference in python type hint?

lumigator/python/mzai/backend/backend/settings.py
```diff
@@ -35,6 +36,24 @@ class BackendSettings(BaseSettings):
     RAY_WORKER_GPUS_ENV_VAR: str = "RAY_WORKER_GPUS"
     RAY_WORKER_GPUS_FRACTION_ENV_VAR: str = "RAY_WORKER_GPUS_FRACTION"
 
+    # Tracking
+    class TrackingBackendType(str, Enum):
+        """Enum for tracking backend types."""
+
+        MLFLOW = "mlflow"
+
+    TRACKING_BACKEND: TrackingBackendType = TrackingBackendType.MLFLOW
```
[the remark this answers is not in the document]
njbrake: Currently the `TRACKING_BACKEND` is not exposed to the .env layer, so it cannot be changed by an env var. However, if you did pipe the env var through the docker-compose file and set TRACKING_BACKEND=asdfasdf, you would get

```
backend-1      | pydantic_core._pydantic_core.ValidationError: 1 validation error for BackendSettings
backend-1      | TRACKING_BACKEND
backend-1      |   Input should be 'mlflow' [type=enum, input_value='asdfasdf', input_type=str]
backend-1      |     For further information visit https://errors.pydantic.dev/2.9/v/enum
```

lumigator/python/mzai/schemas/lumigator_schemas/workflows.py
njbrake: Correct. The string `id` i get back from mllfow throws an error when I try to convert it to a UUID, unfortunately.

lumigator/python/mzai/backend/backend/settings.py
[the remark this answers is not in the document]
javiermtorres: Ok, so marshalling from strs works out of the box 👍

.devcontainer/docker-compose.override.yaml
```diff
@@ -30,3 +36,23 @@ services:
             - .venv/
         - path: lumigator/backend/pyproject.toml
           action: rebuild
+
+  mlflow:
```
javiermtorres, line 40: Should we bind mlflow so tightly to lumigator? Maybe we can just configure its url in lumigator and leave the user to decide how to start it up?

lumigator/backend/backend/tracking/mlflow.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # this corresponds to creating a run in MLflow.
+    # The run will have n number of nested runs,
+    # which correspond to what we call "jobs" in our system
+    def create_workflow(self, experiment_id: str) -> str:
+        """Create a new workflow."""
+        workflow = self._client.create_run(experiment_id=experiment_id)
+        return workflow.info.run_id
+
+    def create_job(self, experiment_id: str, workflow_id: str, data: dict):
```
javiermtorres, line 34: I'd use the terms "workflow run" and "job run" instead of just "workflow" and "job", in the method names and the description. We're recording the _trace_ of those objects, not the objects themselves, and the word run is well understood in mlflow.

javiermtorres (approved):
A couple of comments, otherwise lgtm

javiermtorres (approved):
Just a couple of comments.

.devcontainer/docker-compose.override.yaml
njbrake: I think it is probably the same situation as the rest of the tools. We rely on the tool to exist (e.g. Ray, Sql) and if they have their own, they can disable the usage. Once mlflow is implemented (the next PR that I'm working on), mlflow will be a mandatory part of the stack, similar to Ray. A user can make decisions about mlflow the same way they make decisions about Ray/Sql/Minio?

With this PR I put mlflow into the overrride (because the tracking service isn't actually being used anywhere so people not doing development dont need it yet) , but with the next PR I was planning on moving it into the main docker-compose file.

lumigator/backend/backend/tracking/mlflow.py
njbrake: `word run is well understood in mlflow.`: since mlflow.py is implementing the tracking interface, changing the function name means I have to update the tracking_interface function name. The tracking_interface function name is meant to be implementation agnostic (e.g. mlflow/wandb/ourownthing). I try to avoid using the word "run" anywhere in my functions because that word means so many different things to different people 😆 .

In Lumigator, I was trying to use 'run' only as a verb, not as a noun. Let's forget for a moment that mlflow uses 'run' as a noun, that's part of the mlflow specific terminology that I'm trying to obfuscate away by the tracking interface.

Now having said that, I think that you make a really good point here: we probably need to adjust the naming at some point so that some functions are 'run_workflow' vs 'create_workflow'. E.g. to differentiate between storing the trace of the object, and executing the object.

So maybe we update the workflow routes so that in the backend route/service, you 'run_workflow' but for the tracking interface we 'create_workflow'. WDYT? Maybe we can make that naming update in https://github.com/mozilla-ai/lumigator/pull/768.

PR #755, opened 2025-01-28
Setting Docker image tag

macaab26:
# What's changing

Changing tag in the compose for the backend and frontend images.

# How to test it

Execute make start-lumigator

N/A

# I already...

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -80,7 +80,7 @@ services:
       - local
 
   backend:
-    image: mzdotai/lumigator:latest
+    image: mzdotai/lumigator:v0.1.0-alpha
     build:
       context: .
       dockerfile: "Dockerfile"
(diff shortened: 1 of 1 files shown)


docker-compose.yaml
```diff
@@ -82,7 +82,7 @@ services:
       - local
 
   backend:
-    image: mzdotai/lumigator:latest
+    image: mzdotai/lumigator:v0.1.0-alpha
```
njbrake, line 85: Generally I would have suggested we make some sort of "latest-release" tag so that we don't need to update this on every new release, but this is fine for the time being 👍

njbrake: Sounds good! To clarify, if you run `start-lumigator-build` it will run with the latest version on the main branch, not necessarily the latest release? By release I mean the latest published 'release'/ git release tag version of the code, which is not the same as the latest on the main branch.

PR #762, opened 2025-01-28
Redo Bug fix for annotation/inference endpoint fails with long sequences as input

### LINKED ISSUE #667: [BUG]: Annotation with large texts fails

HareeshBahuleyan:
# What's changing
When testing with Thunderbird dataset, @agpituk found that BART model fails for long sequences with the error:
```
IndexError: index out of range in self
```
As pointed out by @dpoulopoulos , the error comes from a limitation on the maximum number of positional embeddings that the model can have (token embeddings will always be within range).

* This could be fixed by setting the allowed `model_max_length` (the number of tokens that the model sees) for the respective model, e.g. [for BART](https://huggingface.co/facebook/bart-large-cnn/blob/main/config.json#L44) this is 1024
Edit: Instead of hard coding this in config_templates, we search for plausible params that are most like to correspond to `model_max_length` if it is set to the [default value](https://huggingface.co/docs/transformers/v4.17.0/en/main_classes/tokenizer#transformers.PreTrainedTokenizerFast.model_max_length) of VERY_LARGE_INTEGER (int(1e30)). 

* Additionally, one would also need to set the `truncation = True` in the HF pipeline to truncate the sequence at this max_length.

Closes #667 

See [Colab Notebook](https://colab.research.google.com/drive/1ndu9U1ayy5KEXl-z1AZkd_8irb9W1hWN?usp=sharing) for a demo with these two changes that fixes the issue for long text input.

Also removing `max_length` from the configs (this corresponds to the #tokens in the generated output) - this was being hard-coded in [default_infer_template](https://github.com/mozilla-ai/lumigator/blob/a7ad438beedbf8c36e1cec6d04a761d80586a3fa/lumigator/python/mzai/backend/backend/config_templates.py#L91) and [bart_infer_template](https://github.com/mozilla-ai/lumigator/blob/a7ad438beedbf8c36e1cec6d04a761d80586a3fa/lumigator/python/mzai/backend/backend/config_templates.py#L117). After removal, we would use the HF defaults provided in the respective model configs.

# How to test it

Steps to test the changes:

1. Upload `mock_long_sequences_no_gt.csv` (synthetically generated data with no GT) as the dataset
2. Run annotate endpoint for this dataset (uses BART by default)
```
{
  "name": "default annotation with BART",
  "dataset": "<dataset_id>",
  "max_samples": -1,
  "task": "summarization",
  "store_to_dataset": true
}
```

# I already...


lumigator/backend/backend/config_templates.py
```diff
@@ -99,7 +99,7 @@
         "use_fast": "{use_fast}",
         "trust_remote_code": "{trust_remote_code}",
         "torch_dtype": "{torch_dtype}",
-        "max_length": 500
+        "max_new_tokens": 500
```
dpoulopoulos, line 102: Since we have this option in `HfPipelineConfig`, I would change it to:

```suggestion
        "max_new_tokens": "{max_new_tokens}"
```

lumigator/backend/backend/config_templates.py
```diff
@@ -111,14 +111,27 @@
...
     "dataset": {{ "path": "{dataset_path}" }},
+    "hf_pipeline": {{
+        "model_uri": "{model_uri}",
+        "task": "{task}",
+        "accelerator": "{accelerator}",
+        "revision": "{revision}",
+        "use_fast": "{use_fast}",
+        "trust_remote_code": "{trust_remote_code}",
+        "torch_dtype": "{torch_dtype}",
+        "max_new_tokens": 142
```
dpoulopoulos, line 128: Same here:

```suggestion
        "max_new_tokens": "{max_new_tokens}"
```

lumigator/jobs/inference/model_clients.py
```diff
@@ -151,6 +152,51 @@ def _chat_completion(
...
     def __init__(self, config: InferenceJobConfig):
         self._pipeline = pipeline(**config.hf_pipeline.model_dump())
+        self._set_tokenizer_max_length()
+
+    def _set_tokenizer_max_length(self):
+        """Set the tokenizer's model_max_length if it's currently the default very large integer.
+        Checks various possible max_length parameters which varies on model architecture.
+        """
+        config = self._pipeline.model.config
+        logger.info(f"Initial model_max_length {self._pipeline.tokenizer.model_max_length}")
```
dpoulopoulos, line 162: In my opinion, don't say "initial ...". It kind of implies that you will definitely change it. You can also omit this log in favor of the other one, beneath.

```suggestion
        logger.info(f"The maximum number of input tokens is set to {self._pipeline.tokenizer.model_max_length}")
```

lumigator/jobs/inference/model_clients.py
```diff
@@ -151,6 +152,51 @@ def _chat_completion(
...
+        # If suitable model_max_length is already available, don't override it
+        if self._pipeline.tokenizer.model_max_length != VERY_LARGE_INTEGER:
+            logger.info(f"Using model_max_length = {self._pipeline.tokenizer.model_max_length} \
```
dpoulopoulos, line 165: Make it a bit more human readable:

```suggestion
            logger.info(f"Setting the maximum length of input tokens to {self._pipeline.tokenizer.model_max_length} \
```

lumigator/jobs/inference/model_clients.py
```diff
@@ -151,6 +152,51 @@ def _chat_completion(
...
+
+        # Check config parameters
+        for param in plausible_max_length_params:
+            if hasattr(config, param):
+                value = getattr(config, param)
+                if (
+                    isinstance(value, int) and value < VERY_LARGE_INTEGER
+                ):  # Sanity check for reasonable values
+                    self._pipeline.tokenizer.model_max_length = value
+                    logger.info(f"Setting model_max_length to {value} based on config.{param}")
```
dpoulopoulos, line 192:
```suggestion
                    logger.info(f"Setting the maximum length of input tokens to {value} based on the config.{param} attribute.")
```

lumigator/jobs/inference/model_clients.py
```diff
@@ -151,6 +152,51 @@ def _chat_completion(
...
+                    return
+
+        # If no suitable parameter is found, warn the user and continue with the HF default
+        logger.warning(
```
dpoulopoulos, line 196: Inform the user which is the default value, if you have it available. I agree with you that this should be a `warning`, however, currently warnings are not included in job logs. I would leave it as a `warning`, though, and fix this later. Could you create an issue for that?

lumigator/jobs/inference/model_clients.py
HareeshBahuleyan: Keeping this message, re-worded it.
Removing the one below (both will be the same value if `self._pipeline.tokenizer.model_max_length != VERY_LARGE_INTEGER`

lumigator/jobs/inference/model_clients.py
HareeshBahuleyan: Removed this logging statement in favor of the above.

lumigator/jobs/inference/model_clients.py
HareeshBahuleyan: Updated as suggested

lumigator/jobs/inference/model_clients.py
HareeshBahuleyan: I see, will add a Git issue for that. Added the default value being used in the log message.

HareeshBahuleyan:
Hey @dpoulopoulos, I've made the necessary changes and added my responses. Please check and let me know if we are good to go.

lumigator/backend/backend/config_templates.py
HareeshBahuleyan: As discussed, keeping the default for Bart.

lumigator/backend/backend/config_templates.py
HareeshBahuleyan: Done, removed from here and keep the default of 500 in HfPipelineConfig. This allows user to pass a different value if need be.

dpoulopoulos (approved):

PR #765, opened 2025-01-29
Revising documentation on running unit/integration tests

veekaybee:
# What's changing

It was previously hard to understand how to run granular tests (see https://github.com/mozilla-ai/lumigator/pull/763). This PR adds documentation on how to run these tests. This PR also removes references to TestContainers, which we no longer use. 

# I already...


njbrake (approved):
I would prefer that the Makefile be updated to allow a user to conditionally filter tests using something like `TEST_FILTER` arg (similar to how the DEBUGPY_ARGS works), but this PR still seems like a good improvement from what we have now so I'll approve :)

veekaybee:
> I would prefer that the Makefile be updated to allow a user to conditionally filter tests using something like `TEST_FILTER` arg (similar to how the DEBUGPY_ARGS works), but this PR still seems like a good improvement from what we have now so I'll approve :)

In the case of debugpy args - we work with env vars and run code based on that var. Are you thinking we'd also inject a TEST_FILTER env var and set that way? Will merge this PR but very happy to keep this convo going to get to a good efficient local testing strategy.

njbrake:
> > I would prefer that the Makefile be updated to allow a user to conditionally filter tests using something like `TEST_FILTER` arg (similar to how the DEBUGPY_ARGS works), but this PR still seems like a good improvement from what we have now so I'll approve :)
> 
> In the case of debugpy args - we work with env vars and run code based on that var. Are you thinking we'd also inject a TEST_FILTER env var and set that way? Will merge this PR but very happy to keep this convo going to get to a good efficient local testing strategy.

yep that's exactly what I was thinking: an env var that's designed to be passed in as an arg to the make command, e.g. `make test-thingy TEST_FILTER=a_certain_test_name`

veekaybee:
> yep that's exactly what I was thinking: an env var that's designed to be passed in as an arg to the make command, e.g. make test-thingy TEST_FILTER=a_certain_test_name

I think this would make sense if we knew for sure each test method name was granular enough, otherwise this approach seems ok to start with. I'd actually prefer to tackle this by reducing the amount of env vars we pass into pytest directly so we can run something much closer to `uv run pytest` like we used to be able to

javiermtorres:
Thanks a lot!!!!! ^_^

javiermtorres:
> so we can run something much closer to `uv run pytest` like we used to be able to

+1 to that. I sometimes just run pytest with the file::case syntax; the only nuisance is the need to fix all these env vars. Maybe we can instruct the user to source a script, or use a separate .env....?

Issue #527, opened 2024-12-18
Experiment implementation using a tracking server


javiermtorres:
### Motivation

As a follow-up to #525 and #526, write a dummy experiment implementation that records run items at the appropriate level (with no running tasks at this point). The main point is to check the usage of the interface defined in #525 in the MLflow server as deployed in #526.

### Contribution

The idea is to open a PR for this issue.

### Have you searched for similar issues before submitting this one?

- [x] Yes, I have searched for similar issues

Issue #588, opened 2025-01-09
[ENH]: Experiment logs should contain logs from all jobs linked to it


ividal:
### Motivation

Currently getting logs from /experiments delegates to the single ray job with the inference-evaluation combo. This means gettings logs automatically returns the logs of each, run sequentially.

When we have independent jobs (whether they are still inference and eval or anything else) and they are orchestrated inside experiments, getting logs from /experiments (right now, experiments_new) will require some logic to return logs for each of the spawned jobs in a way that is not confusing to the user. Since they are async and not necessarily sequential (evaluation will run after its inference job, but there may be other inference-evaluation runs or even single evaluations, without an inference, run in parallel), merging logs into a single file for the UI to show is dangerous.

This issue is an umbrella to find a sane structure/logic to give the user access to all logs generated, that works for the UI as well. It likely needs to keep an eye on what's planned for the upcoming #525, #526 and #527 .

### Alternatives

Simply returning the individual job logs to the user, with no added logic. At most, the reference of the common experiment_id after #570 us merged.

This implies the UI would need to show them on different tabs/x.

This is still a valid option.

### Have you searched for similar issues before submitting this one?

- [x] Yes, I have searched for similar issues

njbrake:
The initial workflows mlflow implementation handles retrieving all of the logs that are associated with a job

https://github.com/mozilla-ai/lumigator/pull/768 

I think this ticket can stay open though, since there is probably a better way to restructure the logic.

ividal:
The known issue/caveat is we get logs only after a job has finished, so both lumigator contributors and power users might want to see what's going on inbetween. However, for now we can close this issue.

PR #768, opened 2025-01-29
Mlflow implementation of Tracking Interface

njbrake:
# What's changing

Mlflow implementation of tracking interface to store and retrieve job results and workflow results. 

This means we no longer use the ExperimentRepository for tracking experiments. 

This PR also elevates workflow management code out of the Job service, and converts the _on_job_complete function into a "wait_for_complete" function this way we don't need to trigger nested background tasks: a workflow is run as a single background task, which encapsulates it into a single function of the workflows.py, to more easily accommodate an expanded range of workflow tasks (future work here).

This pull request includes several changes to the backend of the Lumigator project, focusing on enhancing the functionality of the experiment and workflow services, improving dependency management, and updating the Makefile for testing configurations. The most significant changes include the addition of tracking client dependencies, updates to experiment and workflow routes, and improvements to job creation processes.

### Enhancements to Experiment and Workflow Services:

* `lumigator/backend/backend/api/deps.py: Added `tracking_client_manager` dependency and updated `get_experiment_service` and `get_workflow_service` to include `tracking_client` as a parameter. 
* Implemented the workflow service to be able to run the `inf + eval ` workflow, which is the only type of workflow supported at the moment

### Updates to Experiment and Workflow Routes:

* `lumigator/backend/backend/api/router.py`: Included the workflows route in the OpenAPI schema.
* `lumigator/backend/backend/api/routes/experiments.py`: Updated the response types for `create_experiment_id` and `get_experiment_new` routes, and added a new route to delete experiments. 
* `lumigator/backend/backend/api/routes/workflows.py`: Added new routes for getting workflow logs and deleting workflows, and updated the request type for creating workflows.

### Improvements to Job Creation Processes:

* `lumigator/backend/backend/api/routes/jobs.py`: Removed `background_tasks` parameter from job creation methods and added asynchronous background tasks for handling job completion and dataset updates. 

### Exception Handling Enhancements:

* `lumigator/backend/backend/main.py`: Added workflow exception mappings to the FastAPI application. 
* `lumigator/backend/backend/services/exceptions/experiment_exceptions.py`: Updated `ExperimentNotFoundError` to use `str` instead of `UUID` for resource ID.
* `lumigator/backend/backend/services/exceptions/workflow_exceptions.py`: Added `WorkflowNotFoundError` class for handling workflow-related exceptions.

### Testing Configuration Updates:

* `Makefile`: Added `MLFLOW_TRACKING_URI` environment variable to `test-backend-unit` and `test-backend-integration` targets.

### Deployment Updates:
* Added mlflow to the main docker compose deployment

Refs #527 
# How to test it

This PR does not yet switch the frontend or SDK over to using workflows, so it can either be tested via curl commands or via the test. The test exercises the integration and makes sure that experiments/workflows/jobs can be created and deleted successfully, and that the logs can be accessed
Steps to test the changes:
1. make local-up
2 .make test-backend-integration

# I already...


lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
```diff
@@ -294,3 +283,124 @@ def test_job_non_existing(local_client: TestClient, dependency_overrides_service
...
+                "model": TEST_CAUSAL_MODEL,
+                "dataset": str(dataset.id),
+                "experiment_id": experiment_id,
+                "max_samples": 1,
+            },
+        ).json()
+    )
+
+    # Wait till the workflow is done
+    workflow_status = workflow_1.status
```
javiermtorres, line 340: Put a max number of polls, otherwise we can be spending millions on CI :D

lumigator/python/mzai/backend/backend/tracking/schemas.py
```diff
@@ -0,0 +1,7 @@
+from pydantic import BaseModel
+
+
+class RunOutputs(BaseModel):
```
javiermtorres, line 4: Does it make sense to just use dicts to Any?

lumigator/python/mzai/backend/backend/services/workflows.py
```diff
@@ -33,60 +31,79 @@ def __init__(
...
+    async def _run_inference_eval_pipeline(
         self,
-        inference_job_id: UUID,
-        request: WorkflowCreate,
-        background_tasks: BackgroundTasks,
-        experiment_id: UUID = None,
+        workflow: WorkflowResponse,
+        request: WorkflowCreateRequest,
     ):
+        """Currently this is our only "workflow. As we make this more flexible to handle different
```
javiermtorres, line 54: I think it could pay to start writing this as if we already had some workflow logic. At least, we can separate job creation from job coordination here. WDYT?

lumigator/python/mzai/backend/backend/services/workflows.py
```diff
@@ -33,60 +31,79 @@ def __init__(
...
+        # submit inference job first
+        inference_job = JobResponse.model_validate(
+            self._job_service.create_job(job_inference_create)
+        )
+        # workflow has now started!
+        self._tracking_client.update_workflow_status(workflow.id, WorkflowStatus.RUNNING)
+
+        # wait for the inference job to complete
+        status = await self._job_service.wait_for_job_complete(
+            inference_job.id, max_wait_time_sec=60 * 10
```
aittalam, line 79: It is rather difficult to know in advance whether a given time is a reasonable wait time for a job to complete (e.g. in this case, 10 minutes might not be enough if the dataset is not trivial).
Is there a way we could say "we don't want to be e.g. pending for too long, but as long as it is running we'll just wait for it to complete"?
Btw, depending on available resources one job might be pending for a while (e.g. one single GPU, 10 evaluations on models which require the whole GPU, each eval takes n minutes, the 10th job will have to wait 9*n minutes)

lumigator/python/mzai/backend/backend/services/workflows.py
njbrake: Totally! I think this ties very well into a comment that @dpoulopoulos made in our meeting today: moving workflow logic into a declarative file so that users can easily build workflows out of the jobs. Created https://github.com/mozilla-ai/lumigator/issues/774 to track this, since I think this totally makes sense to work on after this PR is merged, and can be done in parallel with the other mlflow integration efforts.

lumigator/python/mzai/backend/backend/tracking/schemas.py
njbrake: I was hoping to keep that specifying the metrics dict needs to have float values, since we do some basic sigfig rounding before sending to the client (I think right now I round to 3)

lumigator/python/mzai/backend/backend/tests/integration/api/routes/test_api_workflows.py
njbrake: 💵 💸 
I'll update, good call

lumigator/python/mzai/backend/backend/tracking/schemas.py
javiermtorres: Ok

lumigator/python/mzai/backend/backend/services/workflows.py
javiermtorres: Ok, let's take it in a separate PR

lumigator/backend/backend/api/routes/workflows.py
njbrake, line 1: You'll notice a bunch of deleted routes here. This is because, The get workflow/{workflow_id} route now gives you all the details you need! It gives both the aggregated info as well as the individual info for each job in the workflow. (The exception to this is the logs, which come via the workflow/{workflow_id} /logs route, which aggregates all the logs from the jobs in the workflow together.

lumigator/backend/backend/tracking/mlflow.py
```diff
@@ -14,66 +28,354 @@ class MLflowTrackingClient(TrackingClient):
...
+            metadata = json.loads(resp.text)
+            return JobLogsResponse(**metadata)
+        except json.JSONDecodeError as e:
+            loguru.logger.error(f"JSON decode error: {e}")
+            loguru.logger.error(f"Response text: {resp.text}")
+            raise HTTPException(
+                status_code=HTTPStatus.INTERNAL_SERVER_ERROR, detail="Invalid JSON response"
+            ) from e
+
+    def get_workflow_logs(self, workflow_id: str) -> JobLogsResponse:
```
njbrake, line 295: The caveat here is that it will only retrieve the logs of jobs that have finished: so this route won't give you any sort of live stream of the logs as it works now. I'm not sure if this is a dealbreaker? Or if it's ok since IIRC there were some other kinks in the logging to work out anyways? Theres more work to do with the logging and we are tracking that via https://github.com/mozilla-ai/lumigator/issues/588

docker-compose.yaml
```diff
@@ -168,6 +172,26 @@ services:
...
+    image: ghcr.io/mlflow/mlflow:v2.0.1
+    environment:
+      - MLFLOW_TRACKING_URI=${MLFLOW_TRACKING_URI}
+      - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
+      - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
+      - AWS_DEFAULT_REGION=${AWS_DEFAULT_REGION}
+      - BACKEND_STORE_URI=sqlite:///mlflow.db
+      - ARTIFACT_ROOT=s3://mlflow`
+    ports:
+      - "8001:5000"
```
javiermtorres, line 185: I'd actually move away from the well known range of 8xxx... Why not keep it at 5000?

lumigator/backend/backend/api/routes/jobs.py
```diff
@@ -52,7 +52,17 @@ def create_inference_job(
     response: Response,
     background_tasks: BackgroundTasks,
 ) -> JobResponse:
-    job_response = service.create_job(job_create_request, background_tasks)
+    job_response = service.create_job(job_create_request)
+
+    async def background_task(job_response: JobResponse, job_create_request: JobInferenceCreate):
```
javiermtorres, line 57: I'd say not to use local-scope functions unless the enclosing scope is really needed, and maybe use lambdas in that case. Passing params to module-scoped funs should be enough in most cases. I'm not sure what's the lifecycle of local-scope functions; we might be stuck with lots of copies of the same fun over and over that don't get GC'ed???

lumigator/backend/backend/services/workflows.py
```diff
@@ -120,104 +194,19 @@ def create_workflow(
...
-        result_key = self._job_service._get_results_s3_key(experiment_id)
-        download_url = self._dataset_service.s3_client.generate_presigned_url(
-            "get_object",
-            Params={
-                "Bucket": settings.S3_BUCKET,
-                "Key": result_key,
-            },
-            ExpiresIn=settings.S3_URL_EXPIRATION,
-        )
+    def delete_workflow(self, workflow_id: str) -> WorkflowResponse:
```
javiermtorres, line 206: I assume that this would mean "delete workflow run" until we have a proper workflow record associated with the experiment in which it's included, and with the jobs it contains, right?

lumigator/schemas/lumigator_schemas/workflows.py
```diff
@@ -3,13 +3,20 @@
...
 
-class WorkflowCreate(BaseModel):
+class WorkflowStatus(LowercaseEnum):
+    CREATED = "created"
+    RUNNING = "running"
+    FAILED = "failed"
+    SUCCEEDED = "succeeded"
+
+
+class WorkflowCreateRequest(BaseModel):
```
javiermtorres, line 16: I'd take the opportunity to put a list of JobCreates as a param. This avoids misusing params like "model", which is only necessary in the inference. In a similar way, the WorkflowResponse can carry a list of JobResponses (or a summary with only uuids and status, maybe).

javiermtorres (commented):
I'd insist on adding the list of JobCreate as workflow params. This will avoid making adaptations to the job and workflow params that would need to be removed later on. We can assume that only the [infer, eval] list will be sent, for the moment.

docker-compose.yaml
njbrake: on Macos the Airplay server uses port 5000 😢  https://nono.ma/port-5000-used-by-control-center-in-macos-controlce

Our backend is exposed to port 8000 so I figured I would keep it close to that port.

javiermtorres: It's not a big deal. But consider why Ray uses 8265 as port 🤣

njbrake:
> I'd insist on adding the list of JobCreate as workflow params. This will avoid making adaptations to the job and workflow params that would need to be removed later on. We can assume that only the [infer, eval] list will be sent, for the moment.

Documenting our convo from slack here: This PR will wait for https://github.com/mozilla-ai/lumigator/pull/740, since that PR is refactoring the JobCreation schemas. Following that PR being merged, we'll update this branch to be compatible with those changes, at which point we can merge.

lumigator/backend/backend/api/routes/jobs.py
```diff
@@ -78,7 +80,11 @@ def create_annotation_job(
         output_field="ground_truth",
     )
     inference_job_create_request.store_to_dataset = True
-    job_response = service.create_job(inference_job_create_request, background_tasks)
+    job_response = service.create_job(inference_job_create_request)
+
+    background_tasks.add_task(
+        service.handle_inference_job, job_response.id, inference_job_create_request
```
javiermtorres, line 86: The route package should not be responsible of starting up tasks, imho. Why should it be moved from the services package?

lumigator/backend/backend/services/jobs.py
```diff
@@ -248,34 +247,47 @@ def get_upstream_job_status(self, job_id: UUID) -> str:
         except RuntimeError as e:
             raise JobUpstreamError("ray", "error getting Ray job status", e) from e
 
-    async def on_job_complete(self, job_id: UUID, task: Callable = None, *args):
-        """Watches a submitted job and, when it terminates successfully, runs a given task.
-        Inputs:
-        - job_id: the UUID of the job to watch
-        - task: the function to be called after the job completes successfully
-        - args: the arguments to be passed to the function `task()`
+    async def wait_for_job_complete(self, job_id, max_wait_time_sec=None):
```
javiermtorres, line 250: Please give a sensible default for this (e.g. 5m). Otherwise we can wait arbitrarily if the job somehow goes astray.

lumigator/backend/backend/tests/integration/api/routes/test_api_workflows.py
```diff
@@ -214,75 +206,137 @@ def test_upload_data_no_gt_launch_annotation(
 def test_full_experiment_launch(
     local_client: TestClient, dialog_dataset, dependency_overrides_services
 ):
+    """This is the main integration test: it checks:
```
javiermtorres, line 209: Maybe we could put each of these items also as comments at the start of each section? Make separate functions?

lumigator/backend/backend/api/routes/jobs.py
njbrake: The reasoning was because I'm trying to elevate job-specific logic out of the `create_job` function. I could make a small wrapper function in the service that creates the background task instead? Adding it here in the route seemed like the cleanest solution but I'm fine with moving the task startup into the service layer if you'd like?

lumigator/backend/backend/services/jobs.py
njbrake: 😆  If you're a LOTR fan you'll appreciate this meme 😂 

image

I had an earlier comment from @aittalam about it being difficult to set a default wait time, due to the wide variability of how long a job may be running https://github.com/mozilla-ai/lumigator/pull/768/files/166375c48d00f642c18d41ad6d1306bb6cbac46c#r1935492921 

Similar to how the python requests lib works, I figured the default behavior would be to wait ♾️

lumigator/backend/backend/tests/integration/api/routes/test_api_workflows.py
njbrake: Ooh this is a good idea. I'll try to make the comments more clean.

lumigator/schemas/lumigator_schemas/workflows.py
njbrake: I agree with doing this in the long term, but for the moment this is consistent with the way we create workflows/(old experiments). We have an issue created to support declarative definitions for defining workflows https://github.com/mozilla-ai/lumigator/issues/774 and I'd like to have that issue be the one that updates the API to allow a client to specify the type of workflow configuration

lumigator/backend/backend/services/workflows.py
njbrake: I think we chatted about this in another comment, but I'd like to steer away from using `run` as a noun, since it's an mlflow specific term. the tracking_server api is creating and deleting workflows, while the backend service is running the workflows and checking their status. 

In this situation, this delete_worklfow function deletes from mlflow the workflow and any jobs associated with it. However, I don't handle the case where a user tries to delete a workflow that is still running. I'll update this code to prevent the deletion of a workflow that is in progress (aka has jobs attached to it that haven't yet completed) Thanks!

lumigator/backend/backend/tracking/mlflow.py
njbrake: @aittalam  I'd be interested in your feedback for this one: is this an ok starting point for log access? If we need the lives to be a log stream I can make that work, but it will take some refactoring (and further increase the size of this PR 😆 )

lumigator/backend/backend/services/jobs.py
javiermtorres: Yes, I remember. I don't mind a ludicrous limit - as long as there's a limit. If it can be set right now in tests, I guess that's enough for me at this point.

I assume the python requests will depend on other timeouts like TCP stack timers not to hang indefinitely. But in our case, we don't have other timers that we can rely on, afaict.

lumigator/backend/backend/api/routes/jobs.py
javiermtorres: If you don't mind, yes, I'd prefer it. In that way the routes would handle only adaptation to services, and schema changes, but not other mechanisms like job execution or scheduling. Thanks a lot ^_^

lumigator/backend/backend/tests/integration/api/routes/test_api_workflows.py
javiermtorres: TBH this is much better than my default level of commenting ^_^

lumigator/backend/backend/services/jobs.py
javiermtorres: Maybe we can indicate the last job state as retrieved from Ray, so we can see if the job has been stopped due to this timer (final state still PENDING).

njbrake: Sounds good! At the moment I'll keep this as-is since we define limits in the functions that make the calls, but this is definitely something to look into as we move forwards. I'm afraid to even suggest what I consider a ludicrous limit because then it seems like we risk setting something that we think is ludicrous but a user doesn't, and then we time out their job on them and they get frustrated 😆 . Very hypothetical of course but that's where my mind is going haha

javiermtorres (approved):
Approved with a couple of caveats.
* Adding the jobs directory directly in the PYTHONPATH causes a few issues when the job "pseudo package" has a deeper structure, and the backend starts importing job implementation packages
* The worklow will need to expose the list of jobs configs but this can be done in #740

PR #776, opened 2025-01-30
Only run integration tests on lumigator code changes

chainlink:
# What's changing

This only run the integration tests if the lumigator code has changed (not the helm chart or docs)

# How to test it

N/A

### DIFF
--- a/.github/workflows/lumigator_pipeline.yaml
+++ b/.github/workflows/lumigator_pipeline.yaml
@@ -68,21 +68,35 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Check for modified paths
+        uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            test:
+              - '.github/**'
+              - 'lumigator/**'
+
       - name: Install uv
+        if: steps.filter.outputs.test == 'true' || contains(github.ref, 'refs/tags/')
         uses: astral-sh/setup-uv@v3
 
       - name: Install python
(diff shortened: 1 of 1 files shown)


.github/workflows/lumigator_pipeline.yaml
```diff
@@ -68,21 +68,35 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Check for modified paths
+        uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            test:
+              - '.github/**'
```
javiermtorres, line 77: Please add also the .env.template file (and I'm sure there are other paths we need to check too but I can't remember them right now :-/)

javiermtorres (approved):

.github/workflows/lumigator_pipeline.yaml
macaab26: Probably we should execute those with everything Docker-related, WDYT?

javiermtorres: Probably... In the end, the template will provide env var values that will get into the images / containers 🤔

macaab26: Yeah, if we change something in those files, and we don't test it, it could be problematic

macaab26: Could you add those @chainlink ?

chainlink:
Closed in favour of https://github.com/mozilla-ai/lumigator/pull/1144

PR #778, opened 2025-01-30
Backend: Dataset errors

peteski22:
# What's changing

More dataset errors returned from the `dataset_service`.

* When the dataset is not found when getting the S3 path (`DatasetNotFoundError`)
* When the dataset identified to a job ID (run ID) is not available (`DatasetNotAvailableError`)

`DatasetNotAvailableError` is mapped to [HTTP status code `422`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) 

> The HTTP 422 Unprocessable Content [client error response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses) status code indicates that the server understood the content type of the request content, and the syntax of the request content was correct, but it was unable to process the contained instructions.

>Clients that receive a 422 response should expect that repeating the request without modification will fail with the same error.

Also, updated the `uv` lock file as it seems to be trying to get my attention on every PR and I've ignored it thus far.

# How to test it

Steps to test the changes:

1.
2.
3.

This PR will be updated from `main` and readied up once the integration test issues have been resolved.

# I already...

### DIFF
--- a/lumigator/backend/backend/services/datasets.py
+++ b/lumigator/backend/backend/services/datasets.py
@@ -18,6 +18,7 @@ from backend.repositories.datasets import DatasetRepository
 from backend.services.exceptions.dataset_exceptions import (
     DatasetInvalidError,
     DatasetMissingFieldsError,
+    DatasetNotAvailableError,
     DatasetNotFoundError,
     DatasetSizeError,
     DatasetUpstreamError,
@@ -170,8 +171,6 @@ class DatasetService:
         """Attempts to upload and convert the specified dataset (CSV) to HF format which is then
         stored in S3.
 
-        :return: Information on the uploaded dataset
-        :rtype: DatasetResponse
         :raises DatasetSizeError: if the dataset is too large
         :raises DatasetInvalidError: if the dataset is invalid
         :raises DatasetMissingFieldsError: if the dataset is missing any of the required fields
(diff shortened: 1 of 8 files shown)


lumigator/backend/backend/services/datasets.py
```diff
@@ -224,17 +221,32 @@ def get_dataset(self, dataset_id: UUID) -> DatasetResponse | None:
...
+        :param job_id: job ID to find the associated dataset for
+        :return: information on the associated dataset, for example filename, size, if it contains
+                 ground truth data.
+        :rtype: DatasetResponse
+        :raises DatasetNotAvailableError: if there is no dataset associated with the supplied job ID
+        """
         record = self._get_dataset_record_by_job_id(job_id)
         if record is None:
-            return None
+            raise DatasetNotAvailableError(f"not found using job ID: {job_id}") from None
```
javiermtorres, line 235: Why not the same DatasetNotFoundError?

peteski22: It was to indicate the semantics, with `DatasetNotFoundError` the dataset resource, as identified by it's ID isn't available, but here the job ID is being used to try and find a dataset with a corresponding run ID. 

It felt better to suggest that the dataset (for this job/run ID) wasn't available, to say: the job/run hasn't generated a dataset that has the attribute.

javiermtorres (approved):

PR #779, opened 2025-01-31
Fix for evaluate issues

javiermtorres:
Test PR for evaluate issues blocking the CI chain.

### DIFF
--- a/.github/workflows/lumigator_pipeline.yaml
+++ b/.github/workflows/lumigator_pipeline.yaml
@@ -199,7 +199,6 @@ jobs:
           filters: |
             build_fe:
               - '.github/workflows/**'
-              - 'lumigator/frontend/**'
 
       - name: Truncate commit SHA
         if: steps.filter.outputs.build_fe == 'true' || contains(github.ref, 'refs/tags/')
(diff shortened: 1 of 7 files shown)


.github/workflows/lumigator_pipeline.yaml
```diff
@@ -199,7 +199,6 @@ jobs:
           filters: |
             build_fe:
               - '.github/workflows/**'
-              - 'lumigator/frontend/**'
```
agpituk, line 202: why do we need to remove this? Don't we need to trigger the build when there's a change in this path?

javiermtorres: @macaab26 advised me to remove it for the moment. If you check the previous test, it tries to push the FE but it didn't build the FE previously, so the PR fails.

We can also override the tests and merge this directly.

agpituk (approved):

.github/workflows/lumigator_pipeline.yaml
javiermtorres: I'll take this asap with @chainlink and @macaab26 👍

Issue #770, opened 2025-01-30
Testing and documenting usage of local LLMs


HareeshBahuleyan:
* Llamafile
* vLLM models

Issue #729, opened 2025-01-23
[FEATURE] Support inference for any model tagged with "summarization" on the HF Hub


ividal:
Users want to evaluate their own AI models within Lumigator, but currently, they are limited to pre-integrated options. We currently support only a subset of models tagged with `summarization` on the HF Hub (#339). However, users will likely want to browse the HF model hub and try any model tagged for [summarization]((https://huggingface.co/models?pipeline_tag=summarization)), not just those listed in [`models.yaml`](https://github.com/mozilla-ai/lumigator/blob/1c73e22cc815bf492017a24e1dedca8216f2cb88/lumigator/python/mzai/backend/backend/models.yaml).

This feature will allow users to bring any model (e.g., from Hugging Face or OpenAI API compatible models) and run evaluations seamlessly.

### Problem Statement
Today, Lumigator users can only evaluate pre-integrated models, leaving out a major part of our user base who wants to evaluate other models for summarisation. This was a key pain point raised in our latest Discord event, where:

User X asked: “What kind of models do you support?”
Another user asked: “Do you support all Hugging Face models?”
A third user asked: “Are there any models Lumigator won’t support?”

These questions highlight the uncertainty users experience. Many are assuming Lumigator is limited to certain models we pre+selected for summarisartion when, in reality, we aim to make it more flexible.

### Goal:

Reduce churn by eliminating a key limitation mentioned in Discord and user feedback.
The Priority is to focus on:

Prio 1: OpenAI-compatible APIs (Bedrock, etc. but also ollama, vllm…)
Prio 2: other HF hub models (e.g. generic ones, not summarization specific)

ividal:
Closing, as we are not planning on extending support for more HF models for now (e.g. older GPT2 and others that require custom prompts).

PR #782, opened 2025-01-31
Run eval/inference on any Ollama or Llamafile or vLLM hosted locally

### LINKED ISSUE #781: Run eval/inference on any Ollama or Llamafile hosted locally
User should be able bring their model of choice, supported in Ollama or Llamafile.

HareeshBahuleyan:
# What's changing

Context: Lumigator supports locally hosted models from Llamafile, only a specific model according to the config_templates:
https://github.com/mozilla-ai/lumigator/blob/57da763edfb06fdeb8c1a069db8c56f77568c27d/lumigator/backend/backend/config_templates.py#L160

With this change, users can bring any Llamafile or Ollama model, host it locally and run inference with lumigator. It is straightforward since we can interact with both of them with open AI client.

Refs #729 
Closes #781

# How to test it

Refer to docs PR: https://github.com/mozilla-ai/lumigator/blob/issue-770-docs-usage-of-local-llms/docs/source/user-guides/local-models.md

Steps to test the changes:
Locally spin up the models:

For llamafile:
1. Download any llamafile from here: https://github.com/Mozilla-Ocho/llamafile?tab=readme-ov-file#other-example-llamafiles
2. `chmod 755` on it 
3. Start it as `./<downloaded_model>.llamafile`
4. You should be able to see it running on http://localhost:8080/

For ollama:
1. Download and setup ollama https://ollama.com/download
2. Run `ollama run deepseek-r1:7b` or any other [listed model](https://ollama.com/search) - consider the model size and your system memory.
3. You should be able to see it running on http://localhost:11434/

Then run inference on Lumigator
1. `make local-up`, upload dataset
4. Run inference, specify 
- For Ollama
```
"model": "ollama://deepseek-r1:7b",  # [the model name has to be the same as the hosted model]
"model_uri": "http://localhost:11434/v1" 
...
```
- For Llamafile 
```
"model": "llamafile://deepseek-r1:7b",  # [actually, after the llamafile prefix, you can have any string - it isnt used]
"model_uri": "http://localhost:8080/v1" 
...
```
^courtesy @aittalam for sharing the steps

Not updating the usage of `evalutation_templates` since the old evalutor will be deprecated soon in favor evaluator_lite.

Will be updating this in our docs as a part Issue #770 

# I already...


HareeshBahuleyan:
### Changes to Frontend?

Currently there is an option of selecting `mistralai/Mistral-7B-Instruct-v0.2` which is a llamafile model. Technically we could replace it we `llamafile` (or `ollama`) and let the user use any underlying model. 

But for it to be useful during comparison, I think it makes sense to query the llamafile endpoint and populate the model name in the final tables we show in comparison. This way for example, the user can spin up one `llamafile`, run eval on lumigator, then comeback spin up another model, run eval and so on. In the end, they have a comparison across different llamafile models. 

Seeking any suggestions on how to tackle this^

HareeshBahuleyan:
For llamafile, the output generated contains an end of turn token `<|eot_id|>` - need to check how to get rid of this.

HareeshBahuleyan:
Page with run/testing instructions: https://github.com/mozilla-ai/lumigator/blob/issue-770-docs-usage-of-local-llms/docs/source/user-guides/local-models.md
which will be added to the docs as part of this [PR-806](https://github.com/mozilla-ai/lumigator/pull/806)

lumigator/backend/backend/config_templates.py
```diff
@@ -151,28 +151,38 @@
...
+        inference_templates = {
+            "default": default_infer_template,
+            "oai://gpt-4o-mini": oai_infer_template,
+            "oai://gpt-4o": oai_infer_template,
+            "mistral://open-mistral-7b": oai_infer_template,
+        }
+        # If no config template is provided, use the default one for the job type
+        return inference_templates.get(model_name, default_infer_template)
+
+    elif job_type == JobType.EVALUATION:
```
dpoulopoulos, line 169: This could be just an `if` since there is a `return` statement in every block. Same for the next `elif` statement.

lumigator/backend/backend/config_templates.py
```diff
@@ -151,28 +151,38 @@
...
-        "mistral://open-mistral-7b": oai_eval_template,
-        "llamafile://mistralai/Mistral-7B-Instruct-v0.2": oai_eval_template,
-    },
+def lookup_template(job_type, model_name):
+    if job_type == JobType.INFERENCE:
+        # Locally hosted models using oai client
+        oai_compatible_local_model_prefixes = ("ollama://", "llamafile://")
+        if model_name.startswith(oai_compatible_local_model_prefixes):
+            return oai_infer_template
+        inference_templates = {
```
dpoulopoulos, line 160: Maybe we could make these dictionaries constant variables, outside the `lookup_template` function. WDYT?

HareeshBahuleyan: Sure, that works as well. Pushed the change.

HareeshBahuleyan:
Merged vLLM PR-https://github.com/mozilla-ai/lumigator/pull/792 into this. The instructions to test are available on the same documentation page: https://github.com/mozilla-ai/lumigator/blob/issue-770-docs-usage-of-local-llms/docs/source/user-guides/local-models.md

PR #786, opened 2025-01-31
Basic documentation for the FE

george-mzai:
# What's changing

Added JSDocs & TODOs in services and  store


khaledosman:
Just saw this, I think JSDoc is now redundant that we have typescript, but I would still like to double check to see if some of the comments are helpful and can be cherry picked

khaledosman:
Copied over some of the changes in https://github.com/mozilla-ai/lumigator/pull/823

PR #791, opened 2025-02-03
chore: setup typescript, unit & e2e testing and vue-devtools

khaledosman:
# What's changing

Used similar config to that of a standard new vue 3 project with vite to:
* Setup TypeScript
* Setup Playwright for e2e testing
* Setup Vitest and vue-test-utils for unit testing
* Integrate vue-dev-tools for better debugging experience
* Use vue's standard eslint + typescript config
* Integrate prettier for consistent formatting with typescript and eslint
* Add shims-vue.d.ts to allow typescript to understand .vue files

# How to test it

Steps to test the changes:

1. `npm run-script test:unit` (or `yarn test:unit` to run unit tests (currently there's only a sample test that fails)
2. `npm run-script test:e2e` to run e2e tests (currently there's only a sample test that fails)
3. `npm run-script build` to test type checking and build
4. `npm run-script format` for formatting with prettier

There should probably be upcoming RPs or commits to fix any formatting, linting or type checking warnings or erros

all `.js` files have been renamed to `.ts` extensions

All the config are the standard vue configuration without customizations which are generated when setting up a new project via `npm create vue`

I temporarily set `strict: false` in tsconfig.json as not to fail the build

Also let me know if you have a preference when it comes to cypress vs playwright for e2e testing

# I already...


lumigator/frontend/src/App.vue
```diff
@@ -50,7 +50,7 @@
               class="toaster-content"
               :class="slotProps.message.severity"
             >
-              <span :class="slotProps.message.messageicon" />
+              <span :class="slotProps.message.messageicon" ></span>
```
khaledosman, line 53: span tags are not self-closing

lumigator/frontend/src/App.vue
```diff
@@ -65,7 +65,7 @@
           >
             <component
               :is="Component"
-              @s-disable-scroll="disableScroll = $event"
+              @s-disable-scroll.stop=""
```
khaledosman, line 68: `disableScroll` was not set or used in the component

lumigator/frontend/src/components/molecules/LFileUpload.vue
```diff
@@ -54,8 +54,7 @@ function confirmUpload() {
     },
     reject: () => {
       cancelUpload();
-    },
-    close: false
```
khaledosman, line 58: `close` is not a real attribute in primevue's `confirm` parameter types

lumigator/frontend/src/router/index.ts
```diff
@@ -1,25 +1,24 @@
...
-		icon: 'pi pi-dataset'
+		icon: 'pi pi-dataset',
 	},
 	{
 		path: '/experiments',
 		name: 'experiments',
-		disabled: false,
 		component: LExperiments,
-		icon: 'pi pi-experiments'
+		icon: 'pi pi-experiments',
```
khaledosman, line 21: `disabled` is not a valid vue-router option, it didn't seem to be used explicitly elsewhere in the code either

ideally we should remove the `icon` from there aswell, but thats being used through `slotProps`, there's an opportunity to refactor this in the future to separate vue-router's config from that of the side menu

lumigator/frontend/src/services/models/modelsService.ts
```diff
@@ -8,7 +8,8 @@ async function fetchModels(task_name = 'summarization') {
       return response.data.items;
     }
   } catch (error) {
-    throw new Error('Fetching Models failed.', { cause: error });
+    console.error('Fetching Models failed.', { cause: error });
+    throw error;
```
khaledosman, line 12: `new Error` expects only one param which is a string, so I just logged the error and bubbled it to the caller

agpituk:
Looking great. Should we add documentation about anything new to be installed? I had to run `npx playwright install` in order to be able to run the scripts.

lumigator/frontend/playwright.config.ts
```diff
@@ -0,0 +1,110 @@
+import process from 'node:process'
+import { defineConfig, devices } from '@playwright/test'
+
+/**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// require('dotenv').config();
```
peteski22, line 8: ❓ Should this be removed?

lumigator/frontend/e2e/vue.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { test, expect } from '@playwright/test';
+
+// See here how to get started:
+// https://playwright.dev/docs/intro
+test('visits the app root url', async ({ page }) => {
+  await page.goto('/');
+  await expect(page.locator('div.greetings > h1')).toHaveText('You did it!');
+})
```
peteski22, line 8: ❓Please forgive my ignorance, but is this just an example test or should it actually pass? (should we remove it if not).

lumigator/frontend/playwright.config.ts
```diff
@@ -0,0 +1,110 @@
...
+    //   use: {
+    //     channel: 'msedge',
+    //   },
+    // },
+    // {
+    //   name: 'Google Chrome',
+    //   use: {
+    //     channel: 'chrome',
+    //   },
+    // },
```
peteski22, line 93: ❓ same with these commented out blocks of code (can they be removed)?

lumigator/frontend/playwright.config.ts
```diff
@@ -0,0 +1,110 @@
...
+  ],
+
+  /* Folder for test artifacts such as screenshots, videos, traces, etc. */
+  // outputDir: 'test-results/',
```
peteski22, line 97: ❓ as we can see, my input to this PR isn't low level 😳. Can these go too?

lumigator/frontend/src/components/__tests__/HelloWorld.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { describe, it, expect } from 'vitest'
+
+import { mount } from '@vue/test-utils'
+import LPageHeader from '../molecules/LPageHeader.vue'
+
+describe('LPageHeader', () => {
+  it('renders properly', () => {
+    const wrapper = mount(LPageHeader, { props: { msg: 'Hello Vitest' } })
+    expect(wrapper.text()).toContain('Hello Vitest')
```
peteski22, line 9: 🤔 I don't fully understand the test, but I'm happy we have them. 😄

peteski22 (approved):
Thanks @khaledosman 🥳 

I added some non-blocking comments that were more for my own understanding than anything else.

khaledosman:
> Looking great. Should we add documentation about anything new to be installed? I had to run `npx playwright install` in order to be able to run the scripts.

I think `yarn test:e2e` already informs you to run `npx playwright install` if running it for the first time to install the browser(s), so this is something thats coming from playwright that doesn't need to be explicitly documented in our code/product IMO as it might also change by them

lumigator/frontend/playwright.config.ts
khaledosman: This was autogenerated as is when creating a new vue project, so I think its nice to have, we will likely even need to uncomment it at some point when we have real e2e tests that need to read some environment variables from a `.env` file before running

lumigator/frontend/e2e/vue.spec.ts
khaledosman: Yes, its only an example test that is failing in that case, it was there to test running the e2e tests, I can potentially replace it with a sample running test that applies to our application instead

lumigator/frontend/playwright.config.ts
khaledosman: I would also keep them since we will likely want to test different devices/screen sizes and browser combinations when e2e testing, we can potentially define these with @hasangzl later based on the breakpoints for example, but idk if we care about mobile at the moment anyway, if not we can remove them

lumigator/frontend/playwright.config.ts
khaledosman: yeah this can go since its already the default and added to gitignore 😛

lumigator/frontend/src/components/__tests__/HelloWorld.spec.ts
khaledosman: it basically mounts/renders the component to a headless NodeJS "[JSDOM](https://github.com/jsdom/jsdom)" environment under the hood using [vue-test-utils](https://test-utils.vuejs.org/guide/) as the vue testing library to mount vue components and interact with the vue component apis, then runs tests and assertions against the component using [vitest](https://vitest.dev/) as the test-runner (similar to jest, but much faster and easier to configure since it uses already configured vue's vite setup and understands typescript out of the box).

So in this case this specific test mounts the `LPageHeader` component with the given props (think of it as it creates a new vue app with only this component mounted), then tries to interact with the resulting component/DOM structure to check that it has a text with `Hello Vitest` which it doesn't, but its just a sample test

khaledosman:
I will merge this to unblock the follow up PRs, but feel free to leave more comments or suggestions if you feel strongly about something, I can tackle them in [this follow up PR](https://github.com/mozilla-ai/lumigator/pull/799)

PR #794, opened 2025-02-04
Fix allocation of dead head nodes

javiermtorres:
# What's changing

When storing the Ray data in Redis, the cluster information goes in as well - including the dead head nodes from previous Ray executions. In some situations, e.g. when a job doesn't have node requirements, the scheduler will place it on the node head, but chances are a _dead_ head node is retrieved.

# How to test it

This should avoid error messages mentioning that the assigned node for a job is "infeasible".

N/A

# I already...


njbrake (approved):
Would it be possible to leave some kind of comment in the code explaining what that environment variable is?

javiermtorres:
Tracking this issue in https://discuss.ray.io/t/dead-head-nodes-selected-in-scheduling/21686

PR #799, opened 2025-02-04
chore: enable typescript strict mode and fix type checking errors

khaledosman:
# What's changing

Follow up on https://github.com/mozilla-ai/lumigator/pull/793 and https://github.com/mozilla-ai/lumigator/pull/791

* In the previous PR I set `strict: false` to get type-checking and the CI to pass, this PR aims to remove it and fix underlying type checking issues and add some types

# How to test it

Steps to test the changes:

1. run `npm run-script type-check` or `yarn type-check` to test type-checking issues
2. run `yarn start` or `npm run-script start` to run the project and make sure there are no regression bugs or errors

The types are probably not fully correct and don't differentiate between what the ui components need versus the api response, but its a good starting point where things are working.

# I already...


lumigator/frontend/src/helpers/index.ts
```diff
@@ -41,23 +43,23 @@ export function retrieveEntrypoint(job) {
     // See: lumigator/backend/backend/config_templates.py
 
     // Normalize the max_samples
-    if (jsonObject?.job?.max_samples !== undefined) {
+    if (jsonObject?.job?.max_samples) {
```
khaledosman, line 46: if `max_samples = 0` it won't go into the if branch since `0` is a falsey value

lumigator/frontend/src/stores/datasets/store.ts
```diff
@@ -1,28 +1,29 @@
...
     selectedDataset.value = await datasetsService.fetchDatasetInfo(datasetID);
   }
 
   function resetSelection() {
-    selectedDataset.value = null;
+    selectedDataset.value = undefined;
   }
 
-  async function uploadDataset(datasetFile) {
+  async function uploadDataset(datasetFile: Blob) {
```
khaledosman, line 26: more likely a File type here?

lumigator/frontend/src/components/molecules/LExperimentDetails.vue
```diff
@@ -160,39 +163,39 @@ const { experiments, selectedExperiment, jobs, inferenceJobs, selectedJob } =
...
-const isJobFocused = computed(() => selectedJob.value !== null);
+const isJobFocused = computed(() => Boolean(selectedJob.value));
 const allJobs = computed(() => [...jobs.value, ...inferenceJobs.value]);
 
 // TODO: this needs refactor when the backend provides experiment id
 const currentItemStatus = computed(() => {
   if (isJobFocused.value) {
-    const selected = allJobs.value.filter((job) => job.id === selectedJob.value.id)[0];
-    return selected ? selected.status : selectedJob.value.status;
+    const selected = allJobs.value.filter((job) => job.id === selectedJob.value?.id)[0];
```
khaledosman, line 180:
```suggestion
    const selected = allJobs.value.find((job) => job.id === selectedJob.value?.id);
```

lumigator/frontend/src/components/molecules/LExperimentForm.vue
```diff
@@ -154,9 +157,9 @@ async function triggerExperiment() {
 function resetForm() {
   experimentTitle.value = '';
   experimentDescription.value = '';
-  dataset.value = null;
+  dataset.value = undefined;
   modelSelection.value.selectedModels = [];
-  maxSamples.value = null;
+  maxSamples.value = undefined;
```
khaledosman, line 162: This is more of a preference, I tend to avoid `null` in JavaScript and favor using `undefined` since its already the default value in JavaScript for any empty or uninitialised value

Mostly because `null` has to be explicitly set, and it has to be explicitly checked and `null !== undefined` in JavaScript, so its an easier mental model to only expect the value to either be set to a value or undefined, without having to explicitly set or check against `null` as a potential third value

These are some good reads about the topic

https://medium.com/@hbarcelos/why-i-banned-null-from-my-js-code-and-why-you-should-too-13df90323cfa
https://www.split.io/blog/bias-against-null-javascript/

lumigator/frontend/src/components/molecules/LExperimentTable.vue
```diff
@@ -120,28 +121,30 @@ const columnStyles = computed(() => {
...
   emit('l-experiment-selected', experiment);
 }
 
-function retrieveStatus(experimentId) {
+function retrieveStatus(experimentId: string) {
   const experiment = experiments.value.find((exp) => exp.id === experimentId);
   if (!experiment) {
-    return null;
+    return undefined;
   }
```
khaledosman, line 148: no need to explicitly `return undefined`, just an early `return` achieves the same result

lumigator/frontend/src/components/organisms/LDatasetDetails.vue
```diff
@@ -47,36 +51,36 @@
       </div>
       <div class="l-dataset-details__content-item">
         <span class="l-dataset-details__content-label">submitted</span>
-        <span class="l-dataset-details__content-field">
+        <span class="l-dataset-details__content-field" v-if="selectedDataset">
```
khaledosman, line 54: the whole component should probably not be rendered if there's no `selectedDataset` as it can just receive that as a required prop instead of having these checks everywhere, but that's an opportunity for future refactoring

lumigator/frontend/src/services/datasets/datasetsService.ts
```diff
@@ -70,7 +83,11 @@ async function downloadDataset(id) {
     });
     return fileResponse.data;
   } catch (error) {
-    console.error('Error downloading dataset: ', error.message || error);
+    if (error instanceof Error) {
+      console.error('Error downloading dataset: ', error.message || error);
+    } else if (error instanceof AxiosError) {
+      console.error('Error downloading dataset: ', error.response?.data || error);
+    }
```
khaledosman, line 90: here and in similar cases is also a future refactoring opportunities to properly bubble up the errors to the components or any consumers so they can handle them and get the real error message instead of silently swallowing and logging it here

lumigator/frontend/src/types/Experiment.ts
```diff
@@ -0,0 +1,134 @@
...
+  name: string;
+  experimentStart: string;
+  jobs: Job[];
+  useCase: string;
+  runTime: string;
+  samples?: number;
+  models: Model[];
+  max_samples?: number;
+  status: string;
+};
```
khaledosman, line 15: These types are likely neither accurate nor comprehensive, but its a good start to have some typechecking and know what the data structure looks like

lumigator/frontend/src/types/Job.ts
khaledosman, line 1: This empty file can be removed, or we can split the types to separate files later

agpituk (approved):

PR #800, opened 2025-02-04
build: always pull latest docker image before start-lumigator

khaledosman:
# What's changing

Ensure `make start-lumigator` command always pulls the latest docker image versions before starting, as a way to avoid the need for manually removing and rebuilding images

# How to test it

Steps to test the changes:

1. run `make start-lumigator`
2. potentially push a new version of the docker image
3. run `make start-lumigator` again and ensure that the latest version is running and that the new image has been pulled

Or just double check that docker attempts to pull the images before running
image

# I already...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,7 +118,6 @@ local-logs:
 
 # Launches lumigator in 'user-local' mode (All services running locally, using latest docker container, no code mounted in)
 start-lumigator: check-dot-env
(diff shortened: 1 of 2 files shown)


njbrake:
Would this docker compose argument do the trick? 

https://docs.docker.com/reference/compose-file/services/#pull_policy

Might be a little more streamlined than adding an extra command to the makefile.

agpituk (commented):
thanks for this Khaled!

agpituk (approved):

Issue #122, opened 2024-07-30
Incorporate causal-style models (decoders) from the HF Hub for inference


veekaybee:
We currently support only a subset of models tagged with `summarization` on the HF Hub (https://github.com/mozilla-ai/lumigator/issues/339). Expand the list to support causal-style models (decoders), such as GPT and Llama, as well.

Previews work: https://github.com/mozilla-ai/lumigator/pull/117

PR #808, opened 2025-02-05
Add support for HF text-generation pipelines

dpoulopoulos:
# What's changing

Add support for models filtered by the "text-generation" tag on the HF Hub. This means adding support for causal models, such as Llama, Gemma, Qwen, etc.

Causal models can complete a wide variety of tasks; therefore, users should provide a proper system prompt to inform the models about the task at hand. The user-provided system prompt, along with the data, is then transformed into a model-specific prompt using the chat template defined in the tokenizer config (a Jinja2 string template). This step will fail for older models, which don't have a defined chat template.

Refs #122

# How to test it

Steps to test the changes:

1. Deploy Lumigator with `make local-up`
2. Upload a dataset through the OpenAPI UI.
3. Create an inference job with the following body:
    ```json
    {
      "name": "string",
      "model": "hf://microsoft/Phi-3.5-mini-instruct",
      "dataset": "<dataset-uuid>",
      "max_samples": 1,
      "system_prompt": "You are an advanced AI trained to summarize documents accurately and concisely. Your goal is to extract key information while maintaining clarity and coherence.",
      "task": "text-generation"
    }
    ```

Recent models include a `chat_template` attribute in their `tokenizer_config.json` file. This attribute is essentially a Jinja2 string, such as the one defined [here](https://huggingface.co/microsoft/Phi-3.5-mini-instruct/blob/main/tokenizer_config.json#L119) for `microsoft/Phi-3.5-mini-instruct`.

The pipeline takes the user's input and uses it to render the template into a valid prompt, which then serves as the model's input. The user input follows the same standardized format:

```python
messages = [
    {"role": "system", "content": "You are a helpful assistant."},
    {"role": "user", "content": "What is the meaning of life?"},
]
```

However, some older models, such as [`openai-community/gpt2`](https://huggingface.co/openai-community/gpt2) do not include a chat template. These models break this process.

We should investigate how we could let the user specify a valid chat template. Passing it through the API as another argument doesn't seem to be the best approach, as they would have to do a lot of manual work, escaping special characters to convert the Jinja2 template to a JSON serializable object. I created #809 to track this.

# I already...

- - [x] Added some tests for any new functionality
- - [x] Updated the documentation (both comments in code and [product documentation](https://mozilla-ai.github.io/lumigator) under `/docs`)
- - [x] Checked if a (backend) DB migration step was required and included it if required

### DIFF
--- a/lumigator/backend/backend/api/routes/jobs.py
+++ b/lumigator/backend/backend/api/routes/jobs.py
@@ -31,6 +31,7 @@ from backend.services.exceptions.job_exceptions import (
     JobNotFoundError,
     JobTypeUnsupportedError,
     JobUpstreamError,
+    JobValidationError,
 )
 from backend.settings import settings
 
@@ -42,6 +43,7 @@ def job_exception_mappings() -> dict[type[ServiceError], HTTPStatus]:
         JobNotFoundError: status.HTTP_404_NOT_FOUND,
         JobTypeUnsupportedError: status.HTTP_501_NOT_IMPLEMENTED,
         JobUpstreamError: status.HTTP_500_INTERNAL_SERVER_ERROR,
+        JobValidationError: status.HTTP_400_BAD_REQUEST,
     }
 
 
@@ -55,9 +57,7 @@ def create_inference_job(
 ) -> JobResponse:
     job_response = service.create_job(job_create_request)
 
-    service.add_background_task(
-        background_tasks, service.handle_inference_job, job_response.id, job_create_request
-    )
+    service.add_background_task(background_tasks, service.handle_inference_job, job_response.id, job_create_request)
 
     url = request.url_for(get_job.__name__, job_id=job_response.id)
     response.headers[HttpHeaders.LOCATION] = f"{url}"
@@ -243,9 +243,7 @@ def _get_all_ray_jobs() -> list[RayJobDetails]:
     """Returns metadata that exists in the Ray cluster for all jobs."""
     resp = requests.get(settings.RAY_JOBS_URL, timeout=5)  # 5 seconds
     if resp.status_code != HTTPStatus.OK:
-        loguru.logger.error(
-            f"Unexpected status code getting all jobs: {resp.status_code}, error: {resp.text or ''}"
-        )
+        loguru.logger.error(f"Unexpected status code getting all jobs: {resp.status_code}, error: {resp.text or ''}")
         raise HTTPException(
             status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
             detail="Unexpected error getting job logs",
@@ -257,9 +255,7 @@ def _get_all_ray_jobs() -> list[RayJobDetails]:
     except json.JSONDecodeError as e:
         loguru.logger.error(f"JSON decode error: {e}")
         loguru.logger.error(f"Response text: {resp.text}")
-        raise HTTPException(
-            status_code=HTTPStatus.INTERNAL_SERVER_ERROR, detail="Invalid JSON response"
-        ) from e
+        raise HTTPException(status_code=HTTPStatus.INTERNAL_SERVER_ERROR, detail="Invalid JSON response") from e
 
 
 def _get_ray_job(job_id: UUID) -> RayJobDetails:
@@ -267,17 +263,14 @@ def _get_ray_job(job_id: UUID) -> RayJobDetails:
     resp = requests.get(urljoin(settings.RAY_JOBS_URL, f"{job_id}"), timeout=5)  # 5 seconds
 
     if resp.status_code == HTTPStatus.NOT_FOUND:
-        loguru.logger.error(
-            f"Upstream job metadata not found ({resp.status_code}), error:  {resp.text or ''}"
-        )
+        loguru.logger.error(f"Upstream job metadata not found ({resp.status_code}), error:  {resp.text or ''}")
         raise HTTPException(
             status_code=HTTPStatus.NOT_FOUND,
             detail=f"Job metadata for ID: {job_id} not found",
         )
     elif resp.status_code != HTTPStatus.OK:
         loguru.logger.error(
-            "Unexpected status code getting job metadata text: "
-            f"{resp.status_code}, error: {resp.text or ''}"
+            f"Unexpected status code getting job metadata text: {resp.status_code}, error: {resp.text or ''}"
         )
         raise HTTPException(
             status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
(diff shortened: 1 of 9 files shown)


lumigator/backend/backend/services/jobs.py
```diff
@@ -381,6 +381,9 @@ def create_job(
             job_type = JobType.INFERENCE
             if not request.output_field:
                 request.output_field = "predictions"
+
+            if request.task == "text-generation" and not request.system_prompt:
+                raise ValueError("System prompt is required for text generation tasks.")
```
peteski22, line 386: Hey @dpoulopoulos , could you create a `JobValidationError` in `job_exceptions.py` and raise that instead?

It can then be explicitly documented as a type of error raised by the service, and you could add a mapping in `job_exception_mappings` for the `400` too.

lumigator/jobs/inference/inference_config.py
```diff
@@ -165,6 +165,7 @@ class InferenceJobConfig(BaseInferenceJobConfig):
     name: str
     dataset: DatasetConfig
     job: JobConfig
+    system_prompt: str | None = Field(title="System Prompt", default=None, exclude=True)
```
HareeshBahuleyan, line 168: I was wondering if this `system_prompt` field should belong to `HfPipelineConfig` class (instead of directly in `InferenceJobConfig`) because it is being used for HF models. WDYT?

HareeshBahuleyan: @javiermtorres I see that `InferenceServerConfig` has its own `system_prompt` field
https://github.com/mozilla-ai/lumigator/blob/3bb7551b3bd6984dda5b05fd87534d0d7d3ba3eb/lumigator/jobs/inference/schemas.py#L25

lumigator/backend/backend/services/jobs.py
dpoulopoulos: @peteski22 @javiermtorres can you check if it's okay now? I tested it locally and it seems it raises the right error.

> None: This code is now outdated.

lumigator/backend/backend/api/routes/jobs.py
```diff
@@ -52,7 +52,13 @@ def create_inference_job(
...
     background_tasks: BackgroundTasks,
 ) -> JobResponse:
-    job_response = service.create_job(job_create_request, background_tasks)
+    try:
+        job_response = service.create_job(job_create_request, background_tasks)
+    except ValueError as e:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=f"Invalid request: {e}",
+        ) from e
```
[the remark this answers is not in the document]
dpoulopoulos: Check Peter's comment above ^

lumigator/jobs/inference/model_clients.py
```diff
@@ -150,13 +150,30 @@ def _chat_completion(
 
 class HuggingFaceModelClient(BaseModelClient):
     def __init__(self, config: InferenceJobConfig):
+        logger.info(f"System prompt: {config.system_prompt}")
+        self._system = config.system_prompt
```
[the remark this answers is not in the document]
dpoulopoulos: I agree with you, and I'm all for clarity, however, this is how the OpenAI calls it (e.g., `role = "system"`) and this is how we've used it so far in the `APIModelClient`, the `OpenAIModelClient`, and `MistralModelClient`. So, it's going to be a big refactor, irrelevant to this commit. If you insist on changing that, I'll add a chore commit, doing the full refactor on top. @HareeshBahuleyan, WDYT?

lumigator/jobs/inference/inference_config.py
dpoulopoulos: @HareeshBahuleyan the idea is to slowly remove the `system_prompt` inside the `InferenceServerConfig`, cause now it is a duplicate parameter, and make it available in the general context. I'll do that in a different PR though, if that's okay with you.

lumigator/jobs/inference/model_clients.py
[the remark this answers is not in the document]
aittalam: Ehmmm I am afraid this was my fault 😅, sorry about that!
I agree that if we want to change this we need to keep this consistent across the other classes and its place is probably a follow-up commit. @dpoulopoulos we can also create an issue for that, could be good as a first issue!

lumigator/jobs/inference/model_clients.py
```diff
@@ -199,10 +202,24 @@ def _set_tokenizer_max_length(self):
...
 
     def predict(self, prompt):
-        prediction = self._pipeline(prompt)[0]
-
-        # The result is a dictionary with a single key, which name depends on the task
-        # (e.g., 'summary_text' for summarization)
-        # Get the name of the key and return its value
-        result_key = list(prediction.keys())[0]
-        return prediction[result_key]
+        # When using a text-generation model, the pipeline returns a dictionary with a single key,
```
aittalam, line 205: Thanks for providing these details in comments, they help better understand the differences between different tasks 🙏

aittalam (approved):
Thank you Dimitris for this PR!

I tested this with the following and it worked out as expected:

```
#!/bin/bash

fi

EVAL_NAME="test_inference_phi"
EVAL_DESC="Test inference with microsoft/Phi-3.5-mini-instruct"
EVAL_MODEL="hf://microsoft/Phi-3.5-mini-instruct"
EVAL_TASK="text-generation"
EVAL_SYSTEM_PROMPT="You are a helpful assistant, expert in text summarization. For every prompt you receive, provide a summary of its contents in at most two sentences."
EVAL_DATASET=$DATASET_ID
EVAL_MAX_SAMPLES="5"

JSON_STRING=$(jq -n \
                --arg name "$EVAL_NAME" \
                --arg desc "$EVAL_DESC" \
                --arg model "$EVAL_MODEL" \
                --arg task "$EVAL_TASK" \
                --arg system_prompt "$EVAL_SYSTEM_PROMPT" \
                '{name: $name, description: $desc, model: $model, task: $task, dataset: $dataset_id, max_samples: $max_samples, system_prompt: $system_prompt}' )

curl -s $BACKEND_URL/api/v1/jobs/inference/ \
  -d "$JSON_STRING" | jq
```

I left minor comments, feel free to merge after you have resolved them

lumigator/jobs/inference/inference_config.py
HareeshBahuleyan: Sounds good 👍

lumigator/jobs/inference/model_clients.py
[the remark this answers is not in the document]
dpoulopoulos: Created a new issue here: https://github.com/mozilla-ai/lumigator/issues/865

PR #814, opened 2025-02-06
Bugfix for old experiments endpoint

njbrake:
# What's changing

We have unit/integration testing around the new experiments (e.g. /experiments/new and /workflows) but apparently not for the old experiments endpoint 🙈 

Please feel free to approve and merge this without me, since this is needed to get the frontend working (the frontend is the only thing using the old experiments endpoint.

### DIFF
--- a/lumigator/backend/backend/api/routes/experiments.py
+++ b/lumigator/backend/backend/api/routes/experiments.py
@@ -1,7 +1,7 @@
 from http import HTTPStatus
 from uuid import UUID
 
-from fastapi import APIRouter, BackgroundTasks, status
+from fastapi import APIRouter, status
 from lumigator_schemas.experiments import (
     ExperimentCreate,
     ExperimentIdCreate,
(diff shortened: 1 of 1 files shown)


lumigator/backend/backend/api/routes/experiments.py
```diff
@@ -67,9 +65,9 @@ def get_experiment_result(
 def get_experiment_result_download(
     service: JobServiceDep,
     experiment_id: UUID,
-) -> ExperimentResultDownloadResponse:
+) -> JobResultDownloadResponse:
     """Return experiment results file URL for downloading."""
-    return ExperimentResultDownloadResponse.model_validate(
+    return JobResultDownloadResponse.model_validate(
```
njbrake, line 70: the create_job function returns JobResultDownloadResponse and we were saying it was ExperimentResultDownloadResponse. This worked before because those classes were identical. But now they aren't (because experiment_id is now a str and not a uuid)

javiermtorres (approved):

PR #815, opened 2025-02-06
Jobs return standardized and flexible output

njbrake:
# What's changing
    
Tl;dr this design de-couples the presentation (frontend) layer from the data management/storage layer (backend), as well as the execution layer (job). Although the frontend still has hardcoded references to several keys, they are now organized in such a way that the frontend can implement more dynamic techniques and management of keys when they're ready.

In order to make it easier to add an manage jobs as they are added, this PR creates a loose interface between job result (in the result.json file) and the backend. 

The job can create a schema as they please as long as the main structure of the file they save has only the keys "artifacts", "parameters", and "metrics". This allows for easy storage in the tracking interface (e.g. mlflow) and easy access for the frontend. 

The design focuses on keeping the jobs as a flexible entity that doesn't require knowledge about how the frontend works. The design also allows for new visualizations of existing data, which means that the frontend is able to make decisions about how to display the data, and it isn't tied to the backend.

Refs https://github.com/mozilla-ai/lumigator/issues/678

# How to test it

The unit tests handle testing of the SDK and the backend, but this should be tested by running `make local-up` and then using the Web UI to upload a dataset, create an experiment, and view all the results.

# I already...


njbrake:
@khaledosman adding you to this review for the frontend perspective. I ran the UI to ensure that the changes I made to the Vue code don't break anything I could see. I think that these changes should enable you to simplify how you load and organize the data you get back from the backend.

lumigator/schemas/lumigator_schemas/jobs.py
```diff
@@ -148,6 +148,19 @@ class JobResults(BaseModel):
...
+    """This is a very loose definition of what data
+    should be stored in the output settings.S3_JOB_RESULTS_FILENAME.
+    As long as a job result file only has the fields defined here,
+    it should be accepted by the backend.
+    """
+
+    model_config = ConfigDict(extra="forbid")
+    metrics: dict | None
+    parameters: dict | None
+    artifacts: dict | None
```
javiermtorres, line 161: Maybe dict[str, Any] to a be a bit more, er, pydantic?
Also, I'd rather use an empty dict as the default value instead of allowing None. I don't think there's any difference in semantics, but this allows the user of these fields to handle them in an homogeneous way.

lumigator/backend/backend/services/jobs.py
```diff
@@ -176,15 +177,21 @@ def _add_dataset_to_db(self, job_id: UUID, request: JobInferenceCreate, s3: S3Fi
...
-        with s3.open(f"{settings.S3_BUCKET}/{result_key}", "r") as f:
-            # Validate that the output file adheres to the expected inference output schema
-            results = InferenceJobOutput.model_validate(json.loads(f.read()))
+        results = self._validate_results(job_id, s3)
+
+        # make sure the artifacts are present in the results
+        if not all(
+            key in results.artifacts for key in ["examples", "ground_truth", request.output_field]
+        ):
+            raise ValueError("Missing required fields in the evaluation results.")
```
javiermtorres, line 186: Just to keep it generic...

```suggestion
            raise ValueError("Missing required fields in the job results.")
```

lumigator/backend/backend/services/jobs.py
```diff
@@ -176,15 +177,21 @@ def _add_dataset_to_db(self, job_id: UUID, request: JobInferenceCreate, s3: S3Fi
...
 
-        # define the fields we want to keep from the results JSON
-        # and build a CSV file from it as a BytesIO object
-        fields = ["examples", "ground_truth", request.output_field]
-        bin_data = self._results_to_binary_file(results.model_dump(), fields)
+        dataset_to_save = {
+            "examples": results.artifacts["examples"],
+            "ground_truth": results.artifacts["ground_truth"],
+            request.output_field: results.artifacts[request.output_field],
+        }
```
javiermtorres, line 192: Hum. I'd add an additional layer under `artifacts`. What if a job produces more than one dataset?
Something like:
```suggestion
        dataset_to_save = {
            "examples": results.artifacts[output_artifact]["examples"],
            "ground_truth": results.artifacts[output_artifact]["ground_truth"],
            request.output_field: results.artifacts[output_artifact][request.output_field],
        }
```
Maybe this `output_artifact` is a constant at this stage, or a well known value, so it doesn't need to be specified in the job itself.

lumigator/backend/backend/services/workflows.py
```diff
@@ -152,14 +153,18 @@ async def _run_inference_eval_pipeline(
...
                 metrics={
-                    "rouge1_mean": round(eval_output.rouge.rouge1_mean, 3),
-                    "rouge2_mean": round(eval_output.rouge.rouge2_mean, 3),
-                    "rougeL_mean": round(eval_output.rouge.rougeL_mean, 3),
-                    "rougeLsum_mean": round(eval_output.rouge.rougeLsum_mean, 3),
-                    "bertscore_f1_mean": round(eval_output.bertscore.f1_mean, 3),
-                    "bertscore_precision_mean": round(eval_output.bertscore.precision_mean, 3),
-                    "bertscore_recall_mean": round(eval_output.bertscore.recall_mean, 3),
-                    "meteor_mean": round(eval_output.meteor.meteor_mean, 3),
+                    "rouge1_mean": round(eval_output.metrics["rouge"]["rouge1_mean"], 3),
```
javiermtorres, line 156: Let's think about making this more generic in a future PR. I don't know exactly how, and it'll depend on how the rest of the system develops, but let's keep an eye on this.

javiermtorres (approved):
LGTM. Only a couple of issues:
* Add another layer between `artifacts` and the columns (admittedly waivable if we assume this specific jobs will contain only these columns and will only accept one input, but would make it more generic, like having several in/out params)
* Try to shape the result object fields (params, metrics, etc) into a more "appropriate" type that can be more flexible and at the same time more restrictive. Since I don't have a firm proposal yet :-/ we can leave it for next PR, but just let me know your thoughts on this.

lumigator/backend/backend/services/workflows.py
njbrake: Absolutely! I think we've got some more work to do on how we specify and store metrics.

lumigator/backend/backend/services/jobs.py
njbrake: `What if a job produces more than one dataset?` I agree that we should cover this corner case. However, at the moment we do assume that a job only produces a single dataset. Whenever we want to support the use case where a job is saving multiple datasets, we'll need to refactor a bunch of code. I'd rather not try to support a new feature in this PR, but I think it's a good idea to track that we should do this. Created https://github.com/mozilla-ai/lumigator/issues/835 to keep track of this.

lumigator/schemas/lumigator_schemas/jobs.py
njbrake: Good point. I was trying to make the result object as flexible as possible but maybe I went too far 😆 . I'll update.

njbrake (commented):
Thanks for the review, Javier! You made some great points 💯

njbrake:
> LGTM. Only a couple of issues:
> 
>     * Add another layer between `artifacts` and the columns (admittedly waivable if we assume this specific jobs will contain only these columns and will only accept one input, but would make it more generic, like having several in/out params)

Agree that this needs to be improved. I created https://github.com/mozilla-ai/lumigator/issues/835 to track this, because implementing this change will require refactors in a handful of places and I'd like to prevent this PR from getting too unwieldy. 

> 
>     * Try to shape the result object fields (params, metrics, etc) into a more "appropriate" type that can be more flexible and at the same time more restrictive. Since I don't have a firm proposal yet :-/ we can leave it for next PR, but just let me know your thoughts on this.
This is a good comment and I think in a way it will be evergreen: there's this push and pull between making the output type standardized, opinionated, but also flexible. I liked your suggestion of turning the fields into at least Pydantic BaseModels and initializing them as empty dicts if the fields aren't provided; I implemented that change. I think that this will require more refinement down the road, but at least now this gets us into a world where we have three high level buckets to put things in for the SDK and frontend to access: parameters, metrics, and artifacts.

lumigator/schemas/lumigator_schemas/jobs.py
javiermtorres: Note that your proposal is not strictly bad!!!! I was just trying to come up with something with better properties, and these were only suggestions.

njbrake:
> > 
> > * Try to shape the result object fields (params, metrics, etc) into a more "appropriate" type that can be more flexible and at the same time more restrictive. Since I don't have a firm proposal yet :-/ we can leave it for next PR, but just let me know your thoughts on this.
> > 
> 
> This is a good comment and I think in a way it will be evergreen: there's this push and pull between making the output type standardized, opinionated, but also flexible. I liked your suggestion of turning the fields into at least Pydantic BaseModels and initializing them as empty dicts if the fields aren't provided; I implemented that change. I think that this will require more refinement down the road, but at least now this gets us into a world where we have three high level buckets to put things in for the SDK and frontend to access: parameters, metrics, and artifacts.

Update: I had to undo the suggestion of changing the JobResultOutput fields into Pydantic BaseModels: pydantic models require you to know all of the keys that you want the class to hold, so it can't function as an arbitrary dict. :(

Issue #475, opened 2024-12-05
[FEATURE]: LiteLLM for LLM provider backends


njbrake:
### Motivation

I wanted to get this written down while it's on my mind 😆 
## Motivation
New LLMs and platforms are being introduced by the week, most of which have slightly different APIs and configurations. My understanding is that currently the code at 

https://github.com/mozilla-ai/lumigator/blob/e42fd90f25aa232c00da8b821900741b4daed837/lumigator/python/mzai/jobs/inference/model_clients.py#L18 

is designed to support specific models, but if we want to support additional backends like bedrock, ollama, etc, this class would need to be expanded. the LiteLLM library is build to manage this issue with a long list of supported providers: https://docs.litellm.ai/docs/providers 

## What would it do for Lumigator
Per the [github repo](https://github.com/BerriAI/litellm) :

LiteLLM manages:
    - Translate inputs to provider's completion, embedding, and image_generation endpoints
    - [Consistent output](https://docs.litellm.ai/docs/completion/output), text responses will always be available at ['choices'][0]['message']['content']
    - Retry/fallback logic across multiple deployments (e.g. Azure/OpenAI) - [Router](https://docs.litellm.ai/docs/routing)

Using LiteLLM would give 🐊  faster support for a variety of backends, which could aid in community adoption and contribution. The built in [logging of pricing](https://docs.litellm.ai/docs/completion/token_usage) is also really pretty nice so that users have an idea how much each task is costing them.

## How could it be implemented
The usage of the model_clients pydantic models could be replaced with calls to the LiteLLM library.

## Downsides

Interaction with the LLM is a hugely important piece of 🐊  of course, and tying the project to another library adds reliance on LiteLLM to quickly address bugs and upgrade their dependencies. Additionally, if Lumigator wants to use an LLM not supported by LiteLLM and LiteLLM doesn't want to add support for it for some reason, then we would be back to needing to implement it ourselves

Feel free to close this out if this is not a direction you want to head in.

### Alternatives

https://github.com/langroid and https://www.langchain.com/ appear to also have support for a variety of models, but their use case is more to provide all of the code to build agents etc. 

It looks like Mlflow AI Gateway would also be able to support this, but would involve spinning up a separate service to manage credentials and api calls: https://mlflow.org/docs/latest/llms/deployments/index.html

As I understand it) Lumigator is trying to test LLMs/Agents that are already deployed/designed someplace else.

[...]

PR #818, opened 2025-02-06
Support for Deepseek and theoretically ∞ different models

njbrake:
# What's changing

This PR adds LiteLLM as our backend for connecting to other hosted models. This enables us to connect to the Deepseek APi.

This pull request includes several changes to add support for the Deepseek API and refactor the model clients in the Lumigator application. The changes include updating environment variable references, modifying configuration templates, and refactoring code to use a new LiteLLM model client.

Full transparency: I used copilot to generate the below overview and reviewed them myself, I'm not actually this detail oriented of a person 😆 .
### API Key Support Updates:
* [`README.md`](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L63-R63): Added `DEEPSEEK_API_KEY` to the list of required environment variables for evaluating against LLM APIs. [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L63-R63) [[2]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L90-R94)
* [`docker-compose.yaml`](diffhunk://#diff-3fde9d1a396e140fefc7676e1bd237d67b6864552b6f45af1ebcc27bcd0bb6e9R88): Added `DEEPSEEK_API_KEY` to the environment variables section. [[1]](diffhunk://#diff-3fde9d1a396e140fefc7676e1bd237d67b6864552b6f45af1ebcc27bcd0bb6e9R88) [[2]](diffhunk://#diff-3fde9d1a396e140fefc7676e1bd237d67b6864552b6f45af1ebcc27bcd0bb6e9L146-R149)
* [`docs/source/get-started/installation.md`](diffhunk://#diff-6b6e86452632bfddc31d943f941596afb53dc8fd7edf65b8329bf393b6ada2d5L40-R40): Updated documentation to include `DEEPSEEK_API_KEY`.
* [`notebooks/README.md`](diffhunk://#diff-0b18820a10e63a624bb476291cc951a93d52c0e28f1aa032c0fd8331002e2716L24-R24): Updated Jupyter notebook setup instructions to include `DEEPSEEK_API_KEY`.

### Configuration and Template Changes:
* [`lumigator/backend/backend/config_templates.py`](diffhunk://#diff-a98345e6d8e842834c9bccbc055082d02d4122d8a7529edee4257151fedd283cL153-R159): Added Deepseek templates to the configuration.
* [`lumigator/backend/backend/services/jobs.py`](diffhunk://#diff-59885862ae7b5c8dd1f02df9d48befc70d96c68b03e5308d24a45990f0e5bb22L309-L319): Removed `_set_model_type` method and updated job parameters to use `request.model_url` directly. [[1]](diffhunk://#diff-59885862ae7b5c8dd1f02df9d48befc70d96c68b03e5308d24a45990f0e5bb22L309-L319) [[2]](diffhunk://#diff-59885862ae7b5c8dd1f02df9d48befc70d96c68b03e5308d24a45990f0e5bb22L353-R342) [[3]](diffhunk://#diff-59885862ae7b5c8dd1f02df9d48befc70d96c68b03e5308d24a45990f0e5bb22L362-R351)

### Model Client Refactoring:
* [`lumigator/jobs/inference/inference.py`](diffhunk://#diff-152ef274b64fdf879f65ae105525731fdae593d3b98e9446e55f36d64b3e5eaeL12-R12): Replaced `MistralModelClient` and `OpenAIModelClient` with `LiteLLMModelClient`. [[1]](diffhunk://#diff-152ef274b64fdf879f65ae105525731fdae593d3b98e9446e55f36d64b3e5eaeL12-R12) [[2]](diffhunk://#diff-152ef274b64fdf879f65ae105525731fdae593d3b98e9446e55f36d64b3e5eaeL94-R90)
* [`lumigator/jobs/inference/model_clients.py`](diffhunk://#diff-26ce53c706eccca82798fff14bfc531b57789138468d24dee14552943c640574L1-L9): Removed `APIModelClient`, `OpenAIModelClient`, and `MistralModelClient`; added `LiteLLMModelClient`. [[1]](diffhunk://#diff-26ce53c706eccca82798fff14bfc531b57789138468d24dee14552943c640574L1-L9) [[2]](diffhunk://#diff-26ce53c706eccca82798fff14bfc531b57789138468d24dee14552943c640574L37-R34) [[3]](diffhunk://#diff-26ce53c706eccca82798fff14bfc531b57789138468d24dee14552943c640574L97-R71)

### Dependency Updates:
* [`lumigator/backend/pyproject.toml`](diffhunk://#diff-47c868ee8e29ef8780043c9ce28c397dda50d4716123d50f88e45b479eebb4a5L28-R29): Added `litellm` dependency.
* [`lumigator/jobs/inference/requirements.txt`](diffhunk://#diff-3e89c3eef8e3d8ae6a40498ca64658b9e704705f2dad919268fb0b0e43cdc8a9R3-L5): Added `litellm` and removed `mistralai` and `openai`.
* [`lumigator/jobs/inference/requirements_cpu.txt`](diffhunk://#diff-820eccf4dbf180048693959869ac3a9a1437b6999297ae00aca77256099f4e1dR4-L6): Added `litellm` and removed `mistralai` and `openai`.

### Miscellaneous:
* [`lumigator/schemas/lumigator_schemas/models.py`](diffhunk://#diff-d91203260477ff0f7f717be003d3a5fd6bd0dc98830ac1a193e4f9afbc844d9aL12-R12): Updated `ModelRequirement` to include `DEEPSEEK_API_KEY`.
* [`lumigator/sdk/lumigator_sdk/client.py`](diffhunk://#diff-07da969efd8fe789742d9dc219151f941622a922241e10f460a4531098fa4098L89-R89): Changed default value of `verbose` parameter to `False` in `get_response` method.
* [`lumigator/backend/backend/tests/unit/services/test_job_service.py`](diffhunk://#diff-8892524da6ba446b52a5cca523877c96ce850ff8e11459852208181ec88d6246L64-L67): Removed tests related to `_set_model_type` method. [[1]](diffhunk://#diff-8892524da6ba446b52a5cca523877c96ce850ff8e11459852208181ec88d6246L64-L67) [[2]](diffhunk://#diff-8892524da6ba446b52a5cca523877c96ce850ff8e11459852208181ec88d6246L78-R74)

Refs #475 
# How to test it

Run the notebook I added! You can play around with a few different models, but they need to exist in config_templates.py in the `templates[jobType.INFERENCE]` dict

# I already...

### DIFF
--- a/lumigator/backend/backend/services/jobs.py
+++ b/lumigator/backend/backend/services/jobs.py
@@ -201,7 +201,6 @@ class JobService:
         with s3.open(f"{settings.S3_BUCKET}/{result_key}", "r") as f:
             # Validate that the output file adheres to the expected inference output schema
             results_json = json.loads(f.read())
-            loguru.logger.critical(f"Got results: {results_json}")
             # TODO Move into job-specific territory
             # Specifically, this works only for inferences!
             results = InferenceJobOutput.model_validate(results_json)
@@ -434,6 +433,8 @@ class JobService:
         job_type = request.job_config.job_type
 
         # Create a db record for the job
+        # To find the experiment that a job belongs to,
+        # we'd use https://mlflow.org/docs/latest/python_api/mlflow.client.html#mlflow.client.MlflowClient.search_runs
         record = self.job_repo.create(
             name=request.name,
             description=request.description,
--- a/lumigator/backend/backend/services/workflows.py
+++ b/lumigator/backend/backend/services/workflows.py
@@ -80,7 +80,6 @@ class WorkflowService:
         # submit inference job first
         inference_job = self._job_service.create_job(
             job_infer_create,
-            experiment_id=request.experiment_id,
         )
         # workflow has now started!
         self._tracking_client.update_workflow_status(workflow.id, WorkflowStatus.RUNNING)
@@ -126,8 +125,6 @@ class WorkflowService:
         # submit the job
         evaluation_job = self._job_service.create_job(
             job_eval_create,
-            self._background_tasks,
-            experiment_id=request.experiment_id,
         )
 
         # wait for the evaluation job to complete


lumigator/backend/backend/config_templates.py
```diff
@@ -150,13 +149,14 @@
...
     JobType.INFERENCE: {
         "default": default_infer_template,
-        "oai://gpt-4o-mini": oai_infer_template,
-        "oai://gpt-4o": oai_eval_template,
-        "mistral://open-mistral-7b": oai_infer_template,
+        "gpt-4o-mini": oai_infer_template,
+        "gpt-4o": oai_infer_template,
+        "mistral/open-mistral-7b": oai_infer_template,
+        "deepseek/deepseek-reasoner": oai_infer_template,
+        "deepseek/deepseek-chat": oai_infer_template,
```
njbrake, line 159: Note that I didn't yet update the models.yaml file, because the frontend doesn't use this logic, and the frontend doesn't yet support LiteLLM. this means that the "get_suggested_models()" route won't work for the SDK, because that route is going to give back the model list of models.yaml, which is only applicable for JobType.EVALUATION, not JobType.INFERENCE.

lumigator/jobs/inference/model_clients.py
```diff
@@ -34,54 +31,7 @@ def predict(self, prompt: str) -> str:
...
-    def predict(self, prompt):
-        response = self._get_response_with_retries(self._config, prompt)
-
-        return response.choices[0].message.content
-
-
-class OpenAIModelClient(APIModelClient):
+class LiteLLMModelClient(BaseModelClient):
     """Model client for models served via openai-compatible API.
     For OpenAI models:
```
HareeshBahuleyan, line 36: Could we update the doc strings? `base_url` isnt being explicitly used as an argument

lumigator/jobs/inference/model_clients.py
```diff
@@ -94,58 +44,31 @@ class OpenAIModelClient(APIModelClient):
...
+        response = completion(
+            model=self.config.inference_server.base_url,
+            messages=[
+                {"role": "system", "content": self.system},
+                {"role": "user", "content": prompt},
+            ],
+            max_tokens=self.config.params.max_tokens,
+            frequency_penalty=self.config.params.frequency_penalty,
+            temperature=self.config.params.temperature,
+            top_p=self.config.params.top_p,
```
HareeshBahuleyan, line 65:
```suggestion
            top_p=self.config.params.top_p,
            drop_params=True
```

HareeshBahuleyan: When testing with `mistral/open-mistral-7b`, I get the error:
```
litellm.exceptions.UnsupportedParamsError: litellm.UnsupportedParamsError: mistral does not support parameters: {'frequency_penalty': 0.0}, for model=open-mistral-7b. To drop these, set `litellm.drop_params=True`:
```

lumigator/backend/backend/config_templates.py
```diff
@@ -151,12 +150,26 @@
...
+]
+
 templates = {
     JobType.INFERENCE: {
         "default": default_infer_template,
-        "oai://gpt-4o-mini": oai_infer_template,
-        "oai://gpt-4o": oai_eval_template,
-        "mistral://open-mistral-7b": oai_infer_template,
+        **{endpoint: oai_infer_template for endpoint in supported_litellm_providers_models},
         "llamafile://mistralai/Mistral-7B-Instruct-v0.2": oai_infer_template,
```
HareeshBahuleyan, line 173: I am not sure if it now works with Llamafile. It gave me an error and I had to make the following changes to `completion()`
Assuming Llamafile is running on `http://localhost:8080`:
```
response = completion(
            model=self.config.inference_server.base_url, # this is http://localhost:8080 but not really used (?) when api_base argument is present
            messages=[
                {"role": "system", "content": self.system},
                {"role": "user", "content": prompt},
            ],
            max_tokens=self.config.params.max_tokens,
            frequency_penalty=self.config.params.frequency_penalty,
            temperature=self.config.params.temperature,
            top_p=self.config.params.top_p,
            drop_params=True,
            api_base="http://localhost:8080/v1", # the localhost URL is needed here
            custom_llm_provider="openai" # this arg needs to be set
        )
```
In that case, I think we will need both model_name and model_uri (i.e., base_url) to be available here: 
https://github.com/mozilla-ai/lumigator/blob/acdc4cbefe668cdb542877078431629e6f3a587d/lumigator/jobs/inference/model_clients.py#L56
Being able to set `api_base` would also support the situation if users are standing up their own models on some remote machine.

lumigator/backend/backend/config_templates.py
```diff
@@ -151,12 +150,26 @@
 }}"""
 
 
+# Theoretically many more models and providers could be supported here
```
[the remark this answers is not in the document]
njbrake: I don't think these options are exclusive to each other. The user could use option 1 (use the litellm completion api), and then standup their own litellm proxy, and direct the client from option 1 to make inference requests to the proxy. Does that actually work? I'm not very familiar with the proxy server to know if that is a valid configuration 😅

lumigator/backend/backend/config_templates.py
```diff
@@ -151,12 +150,26 @@
...
+# See https://docs.litellm.ai/docs/providers for a list of supported providers
+supported_litellm_providers_models = [
+    # GPT Models
+    "gpt-4o-mini",
```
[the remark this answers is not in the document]
njbrake: This is actually liteLLM lingo. Since openAI doesn't have the option to host a openai model locally, LiteLLM doesn't use a prefix for their models.

lumigator/backend/backend/config_templates.py
njbrake: What error did it give you? looks like you also had to set drop_params=True so it's not clear which error motivated the changes you made.

lumigator/jobs/inference/model_clients.py
njbrake: Agree! We should update the doc strings and actually I think we should make some changes to the API, once the JobType.EVALUATION configuration is removed (hopefully this will happen next week). I can't merge this PR till that cleanup is done anyways, since the `get_suggested_models` route is incompatible with the litellm changes.

lumigator/jobs/inference/model_clients.py
njbrake: Ah I think this answers one of my earlier questions too! Yeah I guess we need to make the decision about whether we should let LiteLLM ignore unsuported parameters, or whether we should handle them ourselves. I lean towards thinking we should handle them ourselves, but having litellm ignore them for now does seem like a reasonable start (so long as we remember to re-visit this later)

notebooks/pyproject.toml
```diff
@@ -0,0 +1,12 @@
+[project]
+name = "notebooks"
+version = "0.1.0"
+description = "Add your description here"
+readme = "README.md"
+requires-python = ">=3.11"
+dependencies = []
+
+[dependency-groups]
+dev = [
```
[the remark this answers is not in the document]
njbrake: I was managing the notebook with UV, and UV made this pyproject file. You're totally right, I need to figure this part out more and update the documentation appropriately.

njbrake (commented):
Thank you for the review, @HareeshBahuleyan !

lumigator/backend/backend/config_templates.py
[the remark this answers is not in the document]
HareeshBahuleyan: Exactly, we can support both. For the "popular" APIs we have option-1, but users can bring any other model that they would like to test (by standing up via the LiteLLM proxy server). To support both cases, our implementation would be fairly straightforward, with some changes to `LiteLLMModelClient` I think

lumigator/backend/backend/config_templates.py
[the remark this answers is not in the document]
HareeshBahuleyan: Understood that, what I mean is that maybe we can have something internally to differentiate option-1 and option-2 above.
A default or no prefix for option-1.
And a different prefix such as [`litellm_proxy`](https://docs.litellm.ai/docs/providers/litellm_proxy) and llamafile for models that users stand up.

lumigator/backend/backend/config_templates.py
HareeshBahuleyan: The error comes from the fact that LiteLLM doesn't know that the `api_base` for llamafile is. 

In other words, the `completion()` call in LiteLLMModelClient currently supports option-1 (where api_base is known to liteLLM from their configuration, according to the `model` name provided. E.g. gpt-4o -> https://platform.openai.com/ and so on). So `api_base` was not passed to `completion()` here:

https://github.com/mozilla-ai/lumigator/blob/af58a8b9556a4bb5024fb5eb5da03cacea1c1e2a/lumigator/jobs/inference/model_clients.py#L56

I think we should have some way for the user to pass both model (e.g. `llamafile://mistralai/Mistral-7B-Instruct-v0.2`) and model_uri (e.g. `http://localhost:8080/v1`) to completion (to be able to also support llamafile/standing up their own models via option-2).

notebooks/deepseek.ipynb
```diff
@@ -0,0 +1,387 @@
...
+    "   I'm a domain expert in this because I've watched most episodes a few times, so I can judge how good the summary is, in addition to the automatic metrics we get back."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Pre-Requisites\n",
+    "\n",
+    "Before running this notebook, you need to have Lumigator running.You will need to ensure that you have both `OPENAI_API_KEY` and `DEEPSEEK_API_KEY` set in your environment variables. Then, run `make local-up` in order to build and have Lumigator listening. Now Lumigator should be ready for our experiment and be able to make requests to OpenAI and DeepSeek!"
```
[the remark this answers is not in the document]
njbrake: Good idea 💯

notebooks/deepseek.ipynb
```diff
@@ -0,0 +1,387 @@
...
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "from datasets import load_dataset\n",
+    "\n",
+    "# First step, let's prepare the dataset!\n",
+    "\n",
+    "# First, grab the dataset off huggingface: https://huggingface.co/datasets/YuanPJ/summ_screen\n",
```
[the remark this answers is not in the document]
njbrake: Not that I'm aware of. It's a crawl of a few public sites. https://github.com/mingdachen/SummScreen

lumigator/backend/backend/config_templates.py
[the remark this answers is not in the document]
njbrake: Thanks for the comment: can you provide an example? I don't think I totally understand your comment 😅

lumigator/backend/backend/config_templates.py
njbrake: Ah cool. So you're saying that we need to have a conditional check because in some cases, configurations like llamafile/ollama/vllm need an api_base?

HareeshBahuleyan: Exactly 👍

lumigator/backend/backend/config_templates.py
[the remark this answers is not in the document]
HareeshBahuleyan: Its in line with what you mentioned:
>So you're saying that we need to have a conditional check because in some cases, configurations like llamafile/ollama/vllm need an api_base?

And so I am suggesting that this "conditional check" can be based on the prefix.
Like for local models, we can have their respective prefix followed by the model name llamafile://<model_name> or ollama://<model_name>
And for the API providers, it will be just the default model name without prefix. WDYT?

PR #819, opened 2025-02-06
Remove all the completions route code

njbrake:
# What's changing

Since the inference job was created which has support for running the OpenAI and Mistral completion service, the /completions route is duplicating effort and increasing the complexity of the codebase. Lumigator is now designed to be an evaluation platform, not a proxy to other LLMs.

With this in mind, this PR deletes the completion route: inference with OpenAI and Mistral is still supported via running the inference job, so this PR does not remove functionilty, but rather consolidates our LLM inference logic into a single location (the inference job)

# How to test it

The SDK had a connection to the completions/ route but the frontend didn't so the functionality appears to be unused. Unit tests and integration tests passing seem like enough, as long as we all agree on removing it.

# I already...


ividal (approved):
Thanks! This makes sense. As discussed, we may bring back live interacting with a model later on, but if/when that happens we can see if inference as it is already covers the need.

PR #821, opened 2025-02-06
Persist state in named volumes, logs in db

### LINKED ISSUE #816: Persist all states in Lumigator
Currently there are several sources of information that are not reliably persisted:

* SQL database
* S3 store
* Ray information (GCS and logs)

They should all be stored in places that can be picked up when Lumigator resumes.

javiermtorres:
# What's changing

The SQLite embedded DB for the personal deployment is now stored in a volume, so information about datasets and experiments is preserved across Lumigator runs. Additionally, logs are stored in a specific database field in the `jobs` table so they are also persisted across runs, as Ray will not store them in GCS.

Closes #816 

# How to test it

1. Stop Lumigator and remove all containers and volumes from Docker
2. Start Lumigator with `make local-up` or `make start-lumigator-build`
3. Check the Lumigator UI, load a dataset, and start an experiment
4. Close Lumigator with Ctrl-C or `make stop-lumigator`
5. Start Lumigator again
6. Previously available datasets and experiments should still be visible

N/A

# I already...


agpituk (approved):
looks good to me! I still have issues with the ray nodes after start/stopping, but that should go into a different issue!

PR #825, opened 2025-02-07
Reduce ray job cold time

### LINKED ISSUE #694: [FEATURE]: Reduce Ray job cold-start time
### Motivation

Image

The first time we spin up a new inference job on the cluster, it takes over a minute because Ray has to hit HuggingFace, load the model into cache, instantiate the artifact, and then start inference. Model instantiation happens here: https://github.com/mozilla-ai/lumigator/blob/09799ee6f035ea9e00aa0139c3b03dcbcb56f7fe/lumigator/python/mzai/jobs/inference/inference.py#L105 and more specifically the model is loaded here: https://github.com/mozilla-ai/lumigator/blob/09799ee6f035ea9e00aa0139c3b03dcbcb56f7fe/lumigator/python/mzai/jobs/evaluator/evaluator/jobs/model_clients.py#L78

At the cluster level, we can mitigate this by pre-populating the model cache: 

https://huggingface.co/docs/datasets/en/cache

Given that we don't know which model the user will select or their machine specs, we can try a couple strategies here depending on our product outcomes: 

1. Pre-populate a small model and point the user to try out that model locally, cache that model 
2. Load a small model as soon as the user instantiates and loads the UI and give an error message

### Alternatives

_No response_

### Contribution

Happy to work on this

agpituk:
# What's changing
As described in #694 currently Ray first job takes a while in the local environment. What I've tried to do in this PR is to add a new container that builds a cache for Ray. Right now this container is build locally, but if we think is better, I can have the option to get this container in a registry where we can pull (improving also start time)

Refs #694 
Closes #694 

# How to test it

Steps to test the changes:

1. In `main` branch: Ensure you clean all containers AND volumes (`docker volume prune -a`). Then start lumigator locally and record how long it takes to add ground truth to a dataset (I've tested with the one in the repo)
2. Repeat the process in this PR branch (`agpituk/694-reduce-ray-job-cold-time`)
3. hopefully it improves the timing!

This removes the local container that we have cached as a developers, so it's a downside we need to consider if it's worth it.

# I already...


agpituk:
I am aware that a couple of tests are failing, I'm trying to figure them out as locally I can't reproduce that error.

ividal:
Completely forgot the first time around: knowing that we are now pre-downloading models, it is the perfect time to add a warning to users that they _need_ to make sure they have enough disk+docker space available for them.

WDYT of, [here in the README.md](https://github.com/mozilla-ai/lumigator/blob/main/README.md?plain=1#L48), something like:

"""## Get started

The simplest way (...)

- The system Python (version managers such as uv should be deactivated)
**- At least X GB available on disk and allocated for docker, since some small language models will be pre-downloaded.**

You can run and develop Lumigator locally (...)
"""

docker-compose.yaml
```diff
@@ -2,6 +2,16 @@ name: lumigator
 
 services:
 
+  inference-model:
+    build:
+      context: .
+      dockerfile: cache/Dockerfile.model-inference
+    platform: linux/arm64
```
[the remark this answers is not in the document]
agpituk: Done!

cache/Dockerfile.model-inference
```diff
@@ -0,0 +1,22 @@
+# Dockerfile.huggingface-cache
+FROM --platform=linux/arm64 python:3.11-slim
```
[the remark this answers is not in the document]
agpituk: Done!

ividal (approved):
Thanks for the changes! Tried it again on ubuntu + mac - LGTM (after rebasing+tests pass)!

agpituk:
> Thanks for the changes! Tried it again on ubuntu + mac - LGTM (after rebasing+tests pass)!

Thanks Irina for all the comments! I am trying to figure out what's wrong with the tests as they don't look related to my PR. I think it's an issue with the persistence we added recently. I've just triggered another run to triple check and if not I'll follow up tomorrow morning

README.md
```diff
@@ -45,6 +45,7 @@ need to have the following prerequisites installed on your machine:
     - On Linux, you need to follow the
       [post-installation steps](https://docs.docker.com/engine/install/linux-postinstall/).
 - The system Python (version managers such as uv should be deactivated)
+- At least 10 GB available on disk and allocated for docker, since some small language models will be pre downloaded
```
aittalam, line 48: I think it'd be great if we added in the docs (1) what models are downloaded (right now it's bart alone, right? I'd suggest roberta-large too for the bertscore metric), (2) their exact size (bart+roberta are less than 3GB), and (3) how this can be disabled if e.g. someone has no intention of ever running bart. WDYT?

docker-compose.yaml
```diff
@@ -68,6 +78,8 @@ services:
     depends_on:
       redis:
         condition: service_healthy
+      inference-model:
+        condition: service_completed_successfully
```
aittalam, line 82: As this will take a while, what are we planning to do with the other services in the meantime? Options:

- make all of them depend on ray (might not be needed, we can e.g. still upload datasets or directly check previous experiment results)
- have something to prevent us from running ray-dependent workflows until ray is up
- none of the above, but clearly communicate to the users that they'll have to wait a bit before running anything that requires ray (not ideal for beginners IMO)

docker-compose.yaml
```diff
@@ -83,18 +95,18 @@ services:
...
           # ... || true allows this to fail (-e is set)
-          sudo chmod -R 777 /tmp/ray_pip_cache/ || true
-          RAY_JOB_ALLOW_DRIVER_ON_WORKER_NODES=1 RAY_REDIS_ADDRESS=redis:6379 ray start --head --dashboard-port=${RAY_DASHBOARD_PORT} --port=6379 --dashboard-host=0.0.0.0 --ray-client-server-port 10001
           mkdir -p /tmp/ray/session_latest/runtime_resources/pip
           rmdir /tmp/ray/session_latest/runtime_resources/pip/ && ln -s /tmp/ray_pip_cache /tmp/ray/session_latest/runtime_resources/pip
           sleep infinity
     shm_size: 2g
     volumes:
-      - ${HOME}/.cache/huggingface:/home/ray/.cache/huggingface
+      - huggingface_cache_vol:/home/ray/.cache/huggingface
```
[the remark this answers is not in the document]
agpituk: This is definitely one of the problems this PR may introduce (on top of slower CI times)  I moved this to a volume because we create that volume before, so we ensure the bart image is there (reducing first time to experiment). Without it being a volume, I'm not sure if I know how I can add this to ray cache.

README.md
agpituk: Bart has to run right now to generate GT, that's why I added it as a kind of _mandatory_ model. As it is, we can't disable it (apart from manually removing the service from docker-compose, which is not very user friendly I'd say. 
We could maybe add a list of variables that is the models you want to predownload into Ray's cache. Would that work?

cache/Dockerfile.model-inference
```diff
@@ -0,0 +1,19 @@
...
+RUN mkdir -p /home/ray/.cache/huggingface/hub
+
+# Use the huggingface_hub API to download the model exactly as the Hub does.
+# This will create a folder with the proper structure (e.g. blobs, refs, snapshots).
+RUN python -c "\
+from huggingface_hub import snapshot_download; \
+model_path = snapshot_download('facebook/bart-large-cnn', cache_dir='/home/ray/.cache/huggingface/hub'); \
+print('Model downloaded to:', model_path)\
+"
+
```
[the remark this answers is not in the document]
agpituk: Definitely happy with the addition to have more than 1 model (look at my comment above). Not sure I follow about the path. 
In the docker-compose, in this line
`      - huggingface_cache_vol:/home/ray/.cache/huggingface`
we use the same path inside Ray (I did a few tests around this to get it right)

agpituk:
> > > Thanks for the changes! Tried it again on ubuntu + mac - LGTM (after rebasing+tests pass)!
> > 
> > 
> > Thanks Irina for all the comments! I am trying to figure out what's wrong with the tests as they don't look related to my PR. I think it's an issue with the persistence we added recently. I've just triggered another run to triple check and if not I'll follow up tomorrow morning
> 
> Re: persistence, if that's the HF one there's a [new issue open ](https://github.com/mozilla-ai/lumigator/issues/876)as it looks like we lost it in one of the most recent updates. Hope this helps!

Thanks for the comments Davide!

The new issue [876](https://github.com/mozilla-ai/lumigator/issues/876) technically would be fixed by this one, but I say technically because we're removing that cache, but I'd like everyone to agree with this approach

README.md
aittalam: Yes, I think that'd be great! For instance right now we are using roberta-large for bertscore evaluations so the models are already two, and having a list we could point users to makes it easy for them to customise it. Thank you!

cache/Dockerfile.model-inference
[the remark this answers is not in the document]
aittalam: My bad, sorry, I did not explain it properly!
What I meant is that we are printing "Model downloaded to <path>" with that python code, and the <path> will be the directory inside the container... Which makes no sense to the user because it is not where they will look for it if they need it (that is, the volume or the local path, not the container one).
As an example, let's say that we are storing this in the classical HF_HOME path. The user will see a message "Model blahblah downloaded to /home/ray/.cache/huggingface", but that is the folder _in the container_, not on their host.

docker-compose.yaml
```diff
@@ -2,6 +2,18 @@ name: lumigator
...
 services:
 
+  inference-model:
+    build:
+      context: .
+      dockerfile: cache/Dockerfile.model-inference
+    platform: linux/${ARCH}
+    command: /bin/true
+    volumes:
+      - ${HOME}/.cache/huggingface:/home/ray/.cache/huggingface
```
aittalam, line 12:
```suggestion
      - ${HF_HOME}:/home/ray/.cache/huggingface
```

We have just fixed the HF cache issue and refactored the code a bit.
The main idea now is that, consistently with [HuggingFace's definition](https://huggingface.co/docs/huggingface_hub/main/en/package_reference/environment_variables#hfhome), we map the HF cache dir in the container to a HF_HOME dir whose default is ${HOME}/.cache/huggingface, but that can be customised if the user is already using a different place for that (see https://github.com/mozilla-ai/lumigator/pull/935).
Sorry for the change under the hood!

agpituk: updated!

aittalam (approved):
Many thanks Ale, having this plus the ability to turn the feature off and documentation explaining that is IMO the best of all possibilities!
I pre-approve the PR subject to merging with the latest code in main. Thank you!

docker-compose.yaml
aittalam: I created [this issue](https://github.com/mozilla-ai/lumigator/issues/1087) to track this, but I think what you did for the current PR is ok for the time being. Thanks!

Issue #812, opened 2025-02-05
Simplify installation: Centralise YAML configuration for Docker


peteski22:
As described in the RFC, in order to reduce problems with day 1 installation of Lumigator this issue should take the configuration that appears in various places (`.env`, `.env.template`, `.docker-compose.yaml`, etc.) and provide a centralized way for config to be generated.

**Expected outcomes:**

* Documentation for settings
* Default YAML settings
* User supplied YAML settings
* `.env` file generated from the above YAML files and used (temporarily) for deployment
* VS Code users should not see issues attempting to use the built in terminal for deployments

PR #836, opened 2025-02-10
Installation: Day 1 improvements to settings

peteski22:
# What's changing

Improvements to Lumigator's day 1 installation story (settings).

* Remove use of .env file directly (.env.template not needed)
* VS Code terminal issue around default vars is now moot
* Updated docs
* Use `.default.conf` and optional `user.conf` for configuration of app
* Allow default settings and user supplied settings to be combined, preferring user settings

Closes https://github.com/mozilla-ai/lumigator/issues/812

# How to test it

Steps to test the changes:

Defaults:

```bash
make start-lumigator
# No user configuration found, default will be used: '.default.conf'
cat .default.conf | grep DEPLOYMENT_TYPE # should show 'local'
curl -s 'http://localhost:8000/api/v1/health/' | jq '.deployment_type' # should show 'local'
cat build/.env # Should show defaults
make stop-lumigator
```

Then add a setting via the user config override:

```bash
rm user.conf # Sanity check
echo DEPLOYMENT_TYPE=development > user.conf
make start-lumigator
# "Found user configuration: 'user.conf', overrides will be applied"
cat .default.conf | grep DEPLOYMENT_TYPE # should show 'local'
cat user.conf | grep DEPLOYMENT_TYPE # should show 'development'
curl -s 'http://localhost:8000/api/v1/health/' | jq '.deployment_type' # should show 'development'
cat build/.env # Should show aggregated settings
make stop-lumigator
```

# I already...


config.default.yaml
```diff
@@ -0,0 +1,49 @@
...
+    deployment_type: local
+    database_url: sqlite:///local.db
+    # LUMI_API_CORS_ALLOWED_ORIGINS:
+    # Array of origins (See: https://developer.mozilla.org/en-US/docs/Glossary/Origin)
+    # that should be allowed to make Cross-Domain (CORS) API requests to the Lumigator backend API.
+    # The expected format of each is: scheme + domain + port (if no port is specified then 80 is assumed).
+    # e.g. "http://localhost:3000,http://lumigator.mydomain.com"
+    # To allow CORS requests from anywhere specify "*" as any, or the only value.
+    # e.g.  - "*"
+    lumi_api_cors_allowed_origins:
```
agpituk, line 41: can we change lumi in this one to lumigator now?  I know it's part of a different thing but it would be nice to change it at some point (feel free to ignore this comment)

Makefile
```diff
@@ -87,62 +100,51 @@ LOCAL_DOCKERCOMPOSE_FILE:= docker-compose.yaml
...
-	echo -e "\033[1;31m====================================================================\033[0m"; \
-	echo -e "\033[1;31mWARNING: .env and .env.template are out of sync. Please review changes\033[0m"; \
-	echo -e "\033[1;31m====================================================================\033[0m"; \
-	fi
-
-	# Update diff file for next comparison
-	@mv .env.diff.new .env.diff 2>/dev/null || true
+define remove_config_dir
+	@echo "Cleaning up config build directory: '$(CONFIG_BUILD_DIR)'..."
+	@rm -rf $(CONFIG_BUILD_DIR)
```
agpituk, line 105: I tried to pull this branch and stop my containers and it failed because the .env wasn't there. 

Maybe we can add a `|| true` at the end of the rm line?

peteski22: I think it's because the down/stop targets expect the config to be there (from the up/start ones).

I've added the config generation to the stop targets so they generate config before stopping... then clean it up.

Makefile
```diff
@@ -1,11 +1,26 @@
-.PHONY: local-up local-down local-logs clean-docker-buildcache clean-docker-images clean-docker-containers start-lumigator-external-services start-lumigator start-lumigator-postgres stop-lumigator test-sdk-unit test-sdk-integration test-sdk-integration-containers test-sdk test-backend-unit test-backend-integration test-backend-integration-containers test-backend test-jobs-evaluation-unit test-jobs-inference-unit test-jobs test-all check-dot-env
+.PHONY: local-up local-down local-logs clean-docker-buildcache clean-docker-images clean-docker-containers start-lumigator-external-services start-lumigator start-lumigator-postgres stop-lumigator test-sdk-unit test-sdk-integration test-sdk-integration-containers test-sdk test-backend-unit test-backend-integration test-backend-integration-containers test-backend test-jobs-evaluation-unit test-jobs-inference-unit test-jobs test-all config-clean config-generate-env
 
 SHELL:=/bin/bash
 UNAME:= $(shell uname -o)
+
+# Required binaries in order to correctly run the makefile, if any cannot be found the script will fail.
+# uv is only required for local-up (dev).
+REQUIRED_BINARIES := git docker yq
```
agpituk, line 8: we don't need yq anymore!

Makefile
```diff
@@ -1,11 +1,26 @@
...
 
-KEEP_CONTAINERS_UP := $(shell grep -E '^KEEP_CONTAINERS_UP=' .env | cut -d'=' -f2 | tr -d '"' || echo "FALSE")
+KEEP_CONTAINERS_UP ?= "FALSE"
+
+# Configuration to identify the input and output config files
+# NOTE: Changing CONFIG_BUILD_DIR will require review of .gitignore
+CONFIG_BUILD_DIR=build
+# Default config prefixed with dot to hide from user
+CONFIG_DEFAULT=.default.env
+# User editable config file (will be generated if missing)
```
agpituk, line 22: if we generate the config only if missing (the .env file) we still going to end up facing the same problems we had before, with the .env not being in the repo, right? so when someone adds a new variable, we're going to face an issue again (which has been usually the main problem people have run into)

Makefile
peteski22, line 8:
```suggestion
REQUIRED_BINARIES := git docker
```

Makefile
peteski22: Not exactly.

If a new variable is added (to the currently named `.default.env` file) then it would be picked up and used to generate the `build/.env` file used to start the Docker containers.

The only issue we'd still be susceptible to would be if we change an existing default value, as the user would have a copy of the settings containing an older version, and user settings supercede the default ones.

We could resolve this too by just not creating a copy for the user, but it would mean if they wanted to override settings they would have to create the file manually and add the relevant settings. 🤔

peteski22: I've updated the PR to mean we never create the user conf file now, it's a manual step if they want to override one or more settings to do this.

agpituk (approved):
thanks for all the changes, great work Peter!

peteski22:
> thanks for all the changes, great work Peter!

No worries, it was a team effort with the conversations and contributions. Thanks to @macaab26 too! 🥳

PR #844, opened 2025-02-11
Fix name of Ray cluster env var

javiermtorres:
Quickfix for https://github.com/ray-project/ray/issues/32167
The name of the env var is not correctly set in the Ray container entrypoint.


agpituk (approved):
awesome!thanks for the quick turnaround

Issue #846, opened 2025-02-11
[FEATURE]: Improvements in Dataset provision


jularase:
### Motivation

**User Story**
As a developer or technical user evaluating models in Lumigator, I want ready-to-use example datasets and clear guidance on preparing my own data so that I can start evaluating models quickly without getting stuck.

**Problem Statement**
Users who are new to Lumigator often don’t have their own datasets and struggle to understand what a good dataset looks like. This creates a major onboarding hurdle and slows down their ability to test models. By providing preloaded example datasets and making dataset preparation easier, we can significantly reduce friction and improve adoption.

Key Capabilities
- Preload example datasets in Lumigator, so users can run evaluations immediately without needing to upload their own data.
- Provide clear dataset preparation guidelines in the UI, helping users understand the ideal format, structure, and quality of datasets.
- Improve dataset discoverability by adding a dedicated section in documentation and GitHub with discussion around dataset examples.

### Have you searched for similar issues before submitting this one?

- [x] Yes, I have searched for similar issues

jbeliao:
adding here ideas discussed with @agpituk and @ividal  this morning about:
**Enhancing Dataset Selection for Users Without Pre-existing Data**

Not all users uploading datasets for evaluation will have their own high-quality dataset ready. We want to ensure that they can still interact with Lumigator meaningfully, explore evaluation workflows, and test models in realistic conditions. To achieve this, I think we should:
- Offer a selection of preloaded datasets so users can experiment without uploading their own.
- Ensure multilingual support by including diverse datasets.
- Explore synthetic data generation and transcription-based datasets to cover more use cases.

Suggested Datasets to offer:
- The Pile v2 (EleutherAI) – A large-scale dataset optimized for language modeling, useful for evaluating general LLM performance.
- PLEIAS (Multilingual Dataset) – High-quality multilingual dataset covering diverse languages and use cases.
- Common Voice Transcription Data – Leveraging Mozilla’s Common Voice project for evaluating models on transcribed speech-to-text data.
- WMT Translation Datasets – Standard benchmark datasets used for evaluating machine translation systems.
- XSum or CNN/DailyMail – Classic summarization datasets for benchmarking.

This will reduce friction for users, allowing them to test model evaluation workflows without having to source their own datasets first, BUT should be coupled with https://github.com/mozilla-ai/lumigator/issues/834 as it's always better for user to evaluate models against their own use case data

njbrake:
[...]

PR #848, opened 2025-02-11
Expose sample datasets link on datasets page

hasangzl:
# What's changing

Sample datasets currently available in Lumigator’s repository are not exposed to users in the UI. We’ve received feedback from multiple users asking for easier access to sample datasets so they can start exploring Lumigator right away.

Until a more integrated solution is implemented, this PR simply adds sample datasets link to the UI, making it easier to find and use.

Additionally, the `Learn More` documentation link in the empty state will now direct users to the specific section about `Datasets` rather than the general documentation page.

# How to test it
1. Click the `Sample Datasets` link on the `Datasets` page (in the empty state). You should be directed to the folder in Lumigator’s repository where sample datasets are available to use.
2. Click the `Learn More` link at the bottom of the empty state. You should be directed to the relevant section in Lumigator’s documentation.


lumigator/frontend/src/components/molecules/LDatasetEmpty.vue
```diff
@@ -1,10 +1,15 @@
...
-    <h2 class="l-dataset-empty__instructions">Submit a dataset to start using Lumigator.</h2>
+    <h2 class="l-dataset-empty__instructions">Use a dataset as the basis for your evaluation. It includes data for the model you'd like to
+      evaluate and possibly a ground truth "answer".</h2>
     <p class="l-dataset-empty__instructions-text">
       <span>
-        Use a dataset as the basis for your evaluation. It includes data for the model you'd like to
-        evaluate and possibly a ground truth "answer".
+        If you don't have your dataset yet, you can use our
+        <a
+        href="https://github.com/mozilla-ai/lumigator/tree/7b3112a40d44c671810453961a8e8f6fbb0b18c3/lumigator/sample_data"
```
javiermtorres, line 9: Would this be better?
```suggestion
        href="https://github.com/mozilla-ai/lumigator/tree/main/lumigator/sample_data"
```

hasangzl: ah yes!

lumigator/frontend/src/components/molecules/LDatasetEmpty.vue
```diff
@@ -1,10 +1,15 @@
...
     <p class="l-dataset-empty__instructions-text">
       <span>
-        Use a dataset as the basis for your evaluation. It includes data for the model you'd like to
-        evaluate and possibly a ground truth "answer".
+        If you don't have your dataset yet, you can use our
+        <a
+        href="https://github.com/mozilla-ai/lumigator/tree/main/lumigator/sample_data"
+        target="_blank"
+        style="background-color: transparent"
+        > <span class="underline">sample datasets</span> <span class="pi pi-arrow-up-right" /></a>
```
khaledosman, line 12: nit: `<span>` tags are not self-closing in html unlike `<a>` tags

```suggestion
        > <span class="underline">sample datasets</span> <span class="pi pi-arrow-up-right"></span></a>
```

khaledosman (approved):
LGTM

lumigator/frontend/src/styles/_typography.scss
```diff
@@ -12,3 +12,7 @@ h1 {
 .capitalize {
   text-transform: capitalize;
 }
+
+.underline {
+  text-decoration: underline;
+}
```
khaledosman, line 18: instead of adding this class as a global style, you can only add it in the `LDatasetEmpty` component's `<style>` tag which have `<style scoped>`, this makes sure that the css doesn't bleed onto other pages or libraries and that each component truly defines its own styles

hasangzl: I added as a global style because I anticipate we will use this underline in other components/pages. Do you still think I should add it within the component's style tag?

khaledosman: I think so yes, because global styles hurt the reusability of the component. IMO components should be self-containing and reusable enough that they can be plugged in a new project and work the same as-in containing all their dependencies from other code or css. it also helps the bundler make better decisions when tree-shaking or importing the components to import only the code and css used by that component.

But its not a big deal since there's similar stuff I'd like to refactor anyway, so feel free to merge it as is

hasangzl: Gotcha!

hasangzl:
This PR, in part, addresses the improvements mentioned in this [issue](https://github.com/mozilla-ai/lumigator/issues/846) #

Issue #882, opened 2025-02-17
Derive job config from model specs


javiermtorres:
Currently the job configurations are generated using hardcoded, specific tags. This should be made more flexible to allow e.g. an arbitrary llama server with user-supplied server config (host and port, for example) from the models file.

PR #883, opened 2025-02-17
Match examples with current job config, allow external models

### LINKED ISSUE #878: [BUG]: Broken support for local LLMs
### Description

An API change broke support for local LLMs as documented in https://mozilla-ai.github.io/lumigator/user-guides/local-models.html

### Reproduction

- pull lumigator, checkout 1035cc1b554c538e36e7d8c3fd05238cb2e96ce5
- follow through the tutorial -> everything works
- checkout to 4cf82e2df69a7decc6ba735d711b7eaeb2b9b3f7 or later
- re-run the same scripts -> things break

### Relevant log output

```shell
{"detail":[{"type":"missing","loc":["body","job_config"],"msg":"Field required","input":{"name":"Llamafile mistral-7b-instruct-v0.2","description":"Test inference with mistral-7b-instruct-v0.2","model":"llamafile://mistral-7b-instruct-v0.2","dataset":"cea40594-9dc8-4bdf-a817-ea38332cd478","max_samples":"10","model_url":"http://localhost:8080/v1","system_prompt":"You are a helpful assistant, expert in text summarization. For every prompt you receive, provide a summary of its contents in at most two sentences."}}]}
```

#### API change

Before:

Image

After:

Image

### Expected behavior

Docs should be aligned with the API. I think something like the following should work for the llamafile example
(note the new `job_config` section where `model`, `mode_url`, and `system_prompt` are moved). The ollama
example should work if updated accordingly.

```
#!/bin/bash
source common_variables.sh

INFERENCE_NAME="Llamafile mistral-7b-instruct-v0.2"
INFERENCE_DESC="Test inference with mistral-7b-instruct-v0.2"
INFERENCE_MODEL="llamafile://mistral-7b-instruct-v0.2" # Format llamafile://<model_name>
INFERENCE_MODEL_URL="http://localhost:8080/v1" # Llamafile runs on port 8080

curl -s "$BACKEND_URL/api/v1/jobs/inference/" \
-d '{
   "name": "'"$INFERENCE_NAME"'",
   "description": "'"$INFERENCE_DESC"'",
   "dataset": "'"$INFERENCE_DATASET_ID"'",
   "max_samples": "'"$INFERENCE_MAX_SAMPLES"'",
   "job_config" : {
      "model": "'"$INFERENCE_MODEL"'",
      "model_url": "'"$INFERENCE_MODEL_URL"'",
      "system_prompt": "'"$INFERENCE_SYSTEM_PROMPT"'"
   }
}'
```

### System Info

- Lumigator version: anything after 4cf82e2df69a7decc6ba735d711b7eaeb2b9b3f7

### LINKED ISSUE #879: [BUG]: Custom local models are not allowed in inference jobs anymore
### Description

Up until 1035cc1b554c538e36e7d8c3fd05238cb2e96ce5 configuration templates allowed [some flexibility in choosing custom models](https://github.com/mozilla-ai/lumigator/blob/1035cc1b554c538e36e7d8c3fd05238cb2e96ce5/lumigator/backend/backend/config_templates.py#L194).
Right now, this is not possible anymore because [a small list of allowed model is hardcoded into services/jobs.py](https://github.com/mozilla-ai/lumigator/blob/main/lumigator/backend/backend/services/jobs.py#L391).

### Reproduction

Try running lumigator with a llamafile or ollama model as described in [our own guide](https://mozilla-ai.github.io/lumigator/user-guides/local-models.html#) (after [this other bug is fixed](https://github.com/mozilla-ai/lumigator/issues/878)).

### Relevant log output

```shell
The inference job on ray returns the following:


pydantic_core._pydantic_core.ValidationError: 1 validation error for InferenceJobConfig
hf_pipeline.model_uri
  Value error, Unable to resolve asset path from llamafile://mistral-7b-instruct-v0.2. Allowed prefixes: {'hf://'} [type=value_error, input_value='llamafile://mistral-7b-instruct-v0.2', input_type=str]
    For further information visit https://errors.pydantic.dev/2.10/v/value_error


This is due to the fact that any non-identified model is [matched with a huggingface model](https://github.com/mozilla-ai/lumigator/blob/main/lumigator/backend/backend/services/jobs.py#L432).
```

### Expected behavior

- the code should implement the same features previously available in configuration templates (in particular, for oai-api compatible models there should be freedom in specifying the model name)
- ideally, we should not require users to edit the main JobService file if they want to add a new model or change a configuration template (I am 100% ok with pydantic configs, but I'd factor them out from the main code)

### System Info

- Lumigator version: 4cf82e2df69a7decc6ba735d711b7eaeb2b9b3f7

javiermtorres:
# What's changing

The local LLM example has been modified to use the new job configuration. Also, the right model URL is used when generating the job configuration.

Closes #878
Closes #879 

# How to test it

Steps to test the changes:

1. Start Lumigator e.g. by using `make local-up`
2. Follow the steps in the `user-guides/local-models.md` guide
3. The job should run ok, with no validation errors

N/A

# I already...

  - The DB was not updated


ividal (changes requested):
Pre-approved after making the inference dataset change, since otherwise this would break.

Thanks for the quick fix @javiermtorres !

docs/source/user-guides/local-models.md
```diff
@@ -58,20 +58,26 @@ You have a choice of choosing one among the below-mentioned local LLM tools. We
 
    INFERENCE_NAME="Llamafile mistral-7b-instruct-v0.2"
    INFERENCE_DESC="Test inference with mistral-7b-instruct-v0.2"
-   INFERENCE_MODEL="llamafile://mistral-7b-instruct-v0.2" # Format llamafile://<model_name>
+   INFERENCE_MODEL="llamafile://mistralai/mistral-7b-instruct-v0.2" # Format llamafile://<model_name>
    INFERENCE_MODEL_URL="http://localhost:8080/v1" # Llamafile runs on port 8080
 
+   # Fill in with your own UUID
```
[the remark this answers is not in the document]
javiermtorres: I will remove the dataset ID line. I didn't notice it was being set before.

docs/source/user-guides/local-models.md
```diff
@@ -19,7 +19,7 @@ Before installation and setup, here are some recommended system requirements:
       user@host:~/lumigator$ export OPENAI_API_KEY="abc123" # pragma: allowlist secret
       ```
 
-2. A dataset for experimentation: you can upload the {{ '[sample dataset](https://github.com/mozilla-ai/lumigator/blob/{}/lumigator/sample_data/dialogsum_exc.csv)'.format(commit_id) }} provided in the [Lumigator repository](https://github.com/mozilla-ai/lumigator) or upload your own dataset through the [Lumigator UI](../get-started/ui-guide.md#upload-a-dataset).
+2. A dataset for experimentation: you can upload the {{ '[sample dataset](https://github.com/mozilla-ai/lumigator/blob/{}/lumigator/sample_data/dialogsum_exc.csv)'.format(commit_id) }} provided in the [Lumigator repository](https://github.com/mozilla-ai/lumigator) or upload your own dataset through the [Lumigator UI](../get-started/ui-guide.md#upload-a-dataset). You will need the dataset ID as assigned by Lumigator.
```
aittalam, line 22: Where do you need the dataset id in the following?
I agree it would be needed if the scripts required you to use one, but as you can see from the examples they pick up the id from the latest uploaded dataset.
If you want to add this sentence, I think there should be at least one example where the dataset id is actually used explicitly

docs/source/user-guides/local-models.md
```diff
@@ -58,7 +58,7 @@ You have a choice of choosing one among the below-mentioned local LLM tools. We
 
    INFERENCE_NAME="Llamafile mistral-7b-instruct-v0.2"
    INFERENCE_DESC="Test inference with mistral-7b-instruct-v0.2"
-   INFERENCE_MODEL="llamafile://mistral-7b-instruct-v0.2" # Format llamafile://<model_name>
+   INFERENCE_MODEL="llamafile://mistralai/mistral-7b-instruct-v0.2" # Format llamafile://<model_name>
```
aittalam, line 61: This is fine, but it does not address https://github.com/mozilla-ai/lumigator/issues/879 (it will work just because that path is hardcoded in JobService at the moment

docs/source/user-guides/local-models.md
```diff
@@ -67,11 +67,14 @@ You have a choice of choosing one among the below-mentioned local LLM tools. We
    -d '{
       "name": "'"$INFERENCE_NAME"'",
       "description": "'"$INFERENCE_DESC"'",
-      "model": "'"$INFERENCE_MODEL"'",
       "dataset": "'"$INFERENCE_DATASET_ID"'",
       "max_samples": "'"$INFERENCE_MAX_SAMPLES"'",
-      "model_url": "'"$INFERENCE_MODEL_URL"'",
-      "system_prompt": "'"$INFERENCE_SYSTEM_PROMPT"'"
+      "job_config": {
```
aittalam, line 72: TY, this works!
There's another piece of code later (the ollama one) that should be updated accordingly, as described in the [issue](https://github.com/mozilla-ai/lumigator/issues/878)

lumigator/backend/backend/services/jobs.py
```diff
@@ -416,7 +416,8 @@ def generate_inference_job_config(self, request: JobCreate, record_id: UUID, dat
                     temperature=request.job_config.temperature,
                     top_p=request.job_config.top_p,
                 )
-            case "llamafile://mistralai/Mistral-7B-Instruct-v0.2":
+            # case "llamafile://mistralai/Mistral-7b-Instruct-v0.2":
+            case "llamafile://mistralai/mistral-7b-instruct-v0.2":
```
aittalam, line 420: We should drop exact string matching for oai-api compatible models. This does not close https://github.com/mozilla-ai/lumigator/issues/879

docs/source/user-guides/local-models.md
javiermtorres: No, I missed the dataset being set in the common_variables.sh script. Already removed.

docs/source/user-guides/local-models.md
javiermtorres: I know, I know. The external interfaces are currently hardcoded because we are only allowing a set of them in the UI. We can open an issue to allow arbitrary servers. Note that using a specific model_url here is not really helpful because anything one sets up at 8080 will be contacted, so an alternate spec (e.g. `llamafile://server:port`) would be more appropriate. https://github.com/mozilla-ai/lumigator/issues/882 tries to address this at least partially.

lumigator/backend/backend/services/jobs.py
javiermtorres: Working on it. I will probably test it by setting up the llamafile server on the ollama port, to illustrate my previous comment :D

docs/source/user-guides/local-models.md
aittalam: We have two different fields for this kind of models:
- `model`, which is a URI, used to _identify_ the model
- `model_url`, which is, well, a URL :-), used to _hit_ the model

We need both of them because you don't want to put an IP address to specify what model generated some output, and we need a URL to hit the model. The only case where you don't need both is when the IP is a given (e.g. oai:// and mistral:// have their own fixed URLs).

In the old config_templates, we _already_ manage oai-api-compatible models dynamically. I think the best way to do this would be incremental (i.e. first pick up the code in config_templates and use it to restore functionality, then think if there's a better engineered version of this we could work on later and capture it in #882 )

lumigator/backend/backend/services/jobs.py
aittalam: Here's my 2 cents and a little bit of history to better explain our context for URI naming.

URIs are used in lumigator to identify a model (see https://github.com/mozilla-ai/lumigator/pull/883#discussion_r1958042940). In evaluator (previously lm-buddy) we were quite strict with those because every single model was validated as being an actual model on HF. This had some advantages (we could catch an unexisting model very early on), but also its drawbacks (e.g. your arbitrary conversion to lowercase where the model name on HF is `mistralai/Mistral-7B-Instruct-v0.2` could have raised issues).

In particular, we realised that this would not work for local models, because:
1. there is no way to validate them as HF models
2. different technologies might support difererent models (e.g. we usually used vLLM to serve full HF models)
3. different tools might refer to the same models with different strings
4. there's a plethora of variants, e.g. you might want to compare 4 different quantization levels for the same model or even the same quant with different versions of the software, and want to make this explicit in your URI so you can replicate your results

For this reason, we found it helpful to have a catch-all in our config_templates where OAI-api-compatible models can basically be defined the same way regardless of their name. Or, more simply, while previous templates were defined _by model name_ these are defined _by URI prefix_.

This is it, I hope it helps clarifying the reasons of our previous choice and helping the future ones!

javiermtorres: > your arbitrary conversion to lowercase where the model name on HF is mistralai/Mistral-7B-Instruct-v0.2 could have raised issues

IIRC I copied this from the template. I did not intentionally changed the case, but retrieved the info from the models file IIRC.

Since the UI specified a fixed set of models (from the models file), I just allowed those. I have included a quick fix that will use a server config if the model_url is a HTTP url, but no further changes can come in the PR; I would not like to add a whole revamping of model URLs. I'll open an issue about model URL renaming, which we need nonetheless.

docs/source/user-guides/local-models.md
javiermtorres: The idea of this change is exactly to expose the config in the jobs as it is, instead of resorting to error-prone intermediate templates. Currently the issue is in the default mapping for models that don't appear in the UI. AFAICT those were mapped to HF models; I've tweaked this to use the server config instead (fully available to the API caller, so no need for a template) if there's a model_url starting with HTTP. But further changes will require a separate PR.

docs/source/user-guides/local-models.md
HareeshBahuleyan, line 22:
```suggestion
2. A dataset for experimentation: you can upload the {{ '[sample dataset](https://github.com/mozilla-ai/lumigator/blob/{}/lumigator/sample_data/dialogsum_exc.csv)'.format(commit_id) }} provided in the [Lumigator repository](https://github.com/mozilla-ai/lumigator) or upload your own dataset through the [Lumigator UI](../get-started/ui-guide.md#upload-a-dataset).
```

HareeshBahuleyan: +1 regarding mention of dataset ID. To reduce friction for the user, we simply use the latest uploaded dataset ID for the demo walkthough

docs/source/user-guides/local-models.md
HareeshBahuleyan: For `llamafile://` models, what comes after the `//` is not technically being used. 
However for ollama and vLLM (below), the `model` name is necessary.

docs/source/user-guides/local-models.md
```diff
@@ -116,11 +119,14 @@ You can then download the results following the steps described [below](#downloa
...
```
HareeshBahuleyan, line 124: Thanks a lot @javiermtorres for updating the scripts for all the 3 local LLM approaches 🙏

docs/source/user-guides/local-models.md
aittalam: Sorry, I don't understand: wasn't this [removed already](https://github.com/mozilla-ai/lumigator/pull/883/files#r1958021749) after you marked our conversation as resolved?

lumigator/backend/backend/services/jobs.py
aittalam: > Are mistral models also oai-api compatible?

If you refer to the mistral API, there's currently a difference that requires us to have a separate class.
But if you think about a self-served mistral model, that is OAI-API compatible as long as you serve it e.g. with vLLM, ollama, or llamafile

> IIRC I copied this from the template. I did not intentionally changed the case, but retrieved the info from the models file IIRC.
No worries! What I meant is that if we continue allowing for custom-defined URIs for e.g. llamafiles a change like that has no bad consequences.

ividal (approved):

Issue #877, opened 2025-02-14
[BUG]: a new workflow always generates two new datasets


aittalam:
### Description

When one inference + eval workflow is run, two datasets are generated by the same inference job instead of just one.

### Reproduction

- Run a new evaluation workflow given a dataset which already has GT
- Check the list of datasets at the end of the workflow:  two new datasets with predictions will be generated by the same model

### Relevant log output

```shell

```

### Expected behavior

Only one dataset should be generated.
I already pinpointed the issue which is that dataset generation has been enabled automatically both at job level and at workflow level.
I have a possible solution in mind, will self-assign this issue and push a PR soon for it.

### System Info

- OS: Mac
- Lumigator version: https://github.com/mozilla-ai/lumigator/commit/1bb8af995c6df16a0e19653b22de465e0bb457dc

### Have you searched for similar issues before submitting this one?

- [x] Yes, I have searched for similar issues

PR #884, opened 2025-02-17
Refactored dataset saving for inference jobs

aittalam:
# What's changing

Dataset saving is triggered inside the JobService only on inference jobs which explicitly provide a True `store_to_dataset` config variable.

Within an inference + evaluation workflow, inference jobs are called with an explicit `store_to_dataset`=False and dataset saving is immediately enqueued afterwards to guarantee the dataset is present before evaluation.

Closes https://github.com/mozilla-ai/lumigator/issues/877

# How to test it

Steps to test the changes:

1. run a full workflow on a dataset with GT
2. verify that there's exactly one dataset generated by the inference job 
3. run an annotation job on a dataset without GT
4. verify that there's exactly one dataset generated by the inference job 

# I already...

- [x] Added some tests for any new functionality: there are no new functionalities, it's just a bug fix.
- [x] Updated the documentation: no need for updated prod documentation, but I updated comments in code
- [x] Checked if a (backend) DB migration step was required and included it if required: none required

### DIFF
--- a/lumigator/backend/backend/services/jobs.py
+++ b/lumigator/backend/backend/services/jobs.py
@@ -540,8 +540,13 @@ class JobService:
         loguru.logger.info("Submitting {job_type} Ray job...")
         submit_ray_job(self.ray_client, entrypoint)
 
-        # TODO Defer to specific job
-        if job_type == JobType.INFERENCE:
+        # NOTE: Only inference jobs can store results in a dataset atm. Among them:
+        # - prediction jobs are run in a workflow before evaluations => they trigger dataset saving
+        #   at workflow level so it is prepended to the eval job
+        # - annotation jobs do not run in workflows => they trigger dataset saving here at job level
+        # As JobType.ANNOTATION is not used uniformly throughout our code yet, we rely on the already
+        # existing `store_to_dataset` parameter to explicitly trigger this in the annotation case
+        if job_type == JobType.INFERENCE and request.job_config.store_to_dataset:
             self.add_background_task(self._background_tasks, self.handle_inference_job, record.id, request)
 
         loguru.logger.info("Getting response...")
--- a/lumigator/backend/backend/services/workflows.py
+++ b/lumigator/backend/backend/services/workflows.py
@@ -68,7 +68,8 @@ class WorkflowService:
             model_url=request.model_url,
             output_field=request.inference_output_field,
             system_prompt=request.system_prompt,
-            store_to_dataset=True,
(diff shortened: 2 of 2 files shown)


lumigator/backend/backend/services/jobs.py
```diff
@@ -555,8 +555,13 @@ def create_job(
         loguru.logger.info("Submitting {job_type} Ray job...")
         submit_ray_job(self.ray_client, entrypoint)
 
-        # TODO Defer to specific job
-        if job_type == JobType.INFERENCE:
+        # NOTE: Only inference jobs can store results in a dataset atm. Among them:
+        # - prediction jobs are run in a workflow before evaluations => they trigger dataset saving
+        #   at workflow level so it is prepended to the eval job
+        # - annotation jobs do not run in workflows => they trigger dataset saving here at job level
+        # As JobType.ANNOTATION is not used uniformly throughout our code yet, we rely on the already
```
HareeshBahuleyan, line 562: Should we create an issue to use this in future? 

Also the new dataset from the inference + eval workflow is saved as `-annotated.csv` - which should also be probably be changed?

HareeshBahuleyan (approved):
Thanks @aittalam for the fix! I tested the dataset creation and it works as per the steps 👍

lumigator/backend/backend/services/jobs.py
aittalam: Yes, good idea! I'll create a new issue 🙏 
Re: the "annotated" prefix, even if "annotated" can be general enough to refer to generic annotations, what happens when we run predictions on an already annotated dataset? I think your point of having a different prefix might help understanding the iterations that occurred on a dataset from the filename itself.
Also, I realised while using the UI that a dataset with human GT will result as "AI generated" after predictions are run on it. It's true that there is some AI generated content but I am afraid this could be misleading wrt what the synthetic content refers to.
I'll try to capture all of these points in a "consolidate annotation / inference workflow" issue (we can always choose to split that in different subtasks if we see any of them becomes too large)

aittalam: aaand there it is: https://github.com/mozilla-ai/lumigator/issues/889

ividal: At least one of those is an unrelated bug. Namely, the fact that **any** prediction seems to be setting an "AI-annotated" flag, instead of that being set only if the ground-truth column was AI-generated.

So that one should be a separate issue.

As for 1-job workflows, let's discuss value.

Issue #880, opened 2025-02-17
Implement BLEU score as metric for translation use case


HareeshBahuleyan:
[BLEU Score](https://en.wikipedia.org/wiki/BLEU) is the most commonly used metric for assessing the quality of machine translation output.

PR #886, opened 2025-02-17
Add support for BLEU score in evalutator job

HareeshBahuleyan:
# What's changing

Refs #880 

# How to test it

Steps to test the changes:

1. Run an inference job 
2. Once completed run eval with the following params (i.e., adding BLEU to the list of metrics)
```
{
  "name": "new metrics",
  "dataset": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "max_samples": -1,
  "job_config": {
    "job_type": "eval_lite",
    "metrics": [
      "meteor",
      "rouge",
      "bertscore",
      "bleu"
    ]
  }
}
```
3. Download results.json and check if BLEU is being reported along with other metrics

At the moment, this is being computed on a summarization inference job, but it will be useful for translation as a part of #922 

# I already...


njbrake:
The description of this issue should be updated to change eval_lite -> evaluator, since eval_lite no longer exists 👍

lumigator/jobs/evaluator/schemas.py
```diff
@@ -17,7 +17,7 @@ class ModelConfig(BaseModel):
 
 
 class EvaluationConfig(BaseModel):
-    metrics: list[str] = Field(default=["rouge", "meteor", "bertscore"])
+    metrics: list[str] = Field(default=["rouge", "meteor", "bertscore", "bleu"])
```
njbrake, line 20: We should at some point think about how we should decide what metrics to use by default or whether we should always run all them by default, but that's a question for another day :)

njbrake (approved):
LGTM!

PR #895, opened 2025-02-17
[BUG] Fix error when uploading dataset using Swagger with empty run_id

### LINKED ISSUE #618: [BUG]: Uploading dataset using Swagger causes an error
### Description

When trying to upload a dataset without (I used one without ground truth), a `422 - Unprocessable Entity` is returned indicating there is a problem with the `run_id`.

The problem appears to be the 'send empty value' checkbox that Swagger is adding, as it causes the parameter to appear in the request, but with no value.

### Reproduction

1. `git checkout main && git pull`
2. `make local-up`
3. Browse to [dataset upload API docs](http://localhost:8000/docs#/datasets/upload_dataset_api_v1_datasets__post)
4. Attempt to upload a CSV setting `run_id` to 'send empty value' 

```bash
curl -X 'POST' \
  'http://localhost:8000/api/v1/datasets/' \
  -F 'dataset=@thunderbird.csv;type=text/csv' \
  -F 'format=job' \
  -F 'run_id='
```

Image

### Relevant log output

```shell
{
  "detail": [
    {
      "type": "uuid_parsing",
      "loc": [
        "body",
        "run_id"
      ],
      "msg": "Input should be a valid UUID, invalid length: expected length 32 for simple format, found 0",
      "input": "",
      "ctx": {
        "error": "invalid length: expected length 32 for simple format, found 0"
      }
    }
  ]
}
```

### Expected behavior

If checking this box causes validation to occur and fail on the server side, we should either update the server code to handle this, or look at removing the checkbox if it serves no purpose.

```bash
curl -X 'POST' \
  'http://localhost:8000/api/v1/datasets/' \
  -F 'dataset=@thunderbird.csv;type=text/csv' \
  -F 'format=job'
```

e.g. this works fine (see curl command above):

Image

### System Info

- OS Version: macOS Sequoia
- Lumigator commit SHA `87b6a611efcd61083834df502a7218e5084015b0`

HareeshBahuleyan:
# What's changing

Handles cases where run_id id an empty string (occurs when user checks the "Send empty value" checkbox on Swagger UI). We treat empty string cases as null.

402886192-918cb55f-c3de-41b0-8bbb-a15ca40b71ca

Closes #618 

# How to test it

Steps to test the changes:

From the Swagger UI, try uploading a dataset with 
1. run_id field with valid uuid
2. empty run_id field (Send empty value UN-checked) 
3. empty run_id field (Send empty value checked) 
^all these should work without errors

And the following case should return an error.
1. set run_id = "random_string"

# I already...

### DIFF
--- a/lumigator/backend/backend/api/routes/datasets.py
+++ b/lumigator/backend/backend/api/routes/datasets.py
@@ -2,7 +2,7 @@ from http import HTTPStatus
 from typing import Annotated
 from uuid import UUID
 
-from fastapi import APIRouter, Form, Query, UploadFile, status
+from fastapi import APIRouter, Depends, Form, HTTPException, Query, UploadFile, status
 from lumigator_schemas.datasets import DatasetDownloadResponse, DatasetFormat, DatasetResponse
 from lumigator_schemas.extras import ListingResponse
 from starlette.requests import Request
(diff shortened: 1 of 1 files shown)


HareeshBahuleyan:
As pointed out by @javiermtorres actually, the `run_id` should not be visible through the swagger interface. A user cannot say "I created this dataset from a job", only lumigator itself should say that.  
Similar for the other fields `generated` and `generated_by`.

Hence, closing this issue.

PR #896, opened 2025-02-17
API: Allow redaction on entrypoint config for a Job

peteski22:
# What's changing

* Allow redaction 'middleware' for models (first attempt)
* Restructure Job via API to include config (dataset, max samples, model, name)
* UI updates to fix some things

Closes: https://github.com/mozilla-ai/lumigator/issues/933

# I already...


lumigator/frontend/src/types/JobDetails.ts
```diff
@@ -5,7 +5,17 @@ export type JobDetails = {
...
+  config: {
+    name: string
+    dataset?: {
+      id: string
+      name: string
+    }
+    max_samples?: number
+    model?: {
+      path: string
+    }
```
khaledosman, line 17: when can these be `undefined` ? I'm guessing we typically show them in the UI, so ideally they should be there.

Is this one of the cases where the data structure for "annotation jobs" are different?

peteski22: I've updated this, only dataset might be missing (e.g. for evaluation jobs)

lumigator/frontend/src/stores/datasetsStore.ts
```diff
@@ -56,6 +55,9 @@ export const useDatasetStore = defineStore('datasets', () => {
     }
     inferenceJobs.value = allJobs
       .filter((job) => job.metadata.job_type === 'inference')
+      // NOTE: 'temporary fix' to prevent showing inference jobs that weren't created by the UI
```
javiermtorres, line 58: maybe filter jobs that don't have a set experiment ref (ref eq to null)?

lumigator/schemas/lumigator_schemas/jobs.py
```diff
@@ -66,6 +74,170 @@ class JobSubmissionResponse(BaseModel):
     driver_node_id: str | None = None
     driver_exit_code: int | None = None
 
+    # Key used to identify the entrypoint in data sent to Ray.
```
javiermtorres, line 77: This should be placed imho somewhere else. The schemas are meant to be declarative schemas, and hold as little impl as possible.

lumigator/frontend/src/stores/datasetsStore.ts
peteski22: This was just a temporary fix for now but was part of: https://github.com/mozilla-ai/lumigator/pull/986

Issue #573, opened 2025-01-07
Deprecate old experiments in favor of new code


aittalam:
- rename eval_lite to evaluation
- switch experiment_new to experiment
- remove old evaluation code
- tests (all CI + manual tests with UI)

njbrake:
I'm thinking that I'll split this into three PRs: 

1. Delete the old evaluator code
2. Rename eval_lite to evaluator since the old evaluator is gone
3. Update the API to delete the "new" experiment endpoints and make them the main experiment endpoints.

PR #903, opened 2025-02-18
Remove unused Schema objects

njbrake:
# What's changing

Now that the experiments endpoint has been updated, a few schema objects are no longer used. Removing them.

Refs #573 

# How to test it

CI testing

# I already...


lumigator/schemas/lumigator_schemas/experiments.py
```diff
@@ -7,19 +7,6 @@
 
 
 class ExperimentCreate(BaseModel):
-    name: str
-    description: str = ""
-    model: str
-    task: str | None = "summarization"
-    dataset: UUID
-    max_samples: int = -1  # set to all samples by default
```
javiermtorres, line 15: Check with Irina if the samples are set by experiment. IIRC it had to be consistent across workflows (but I may be wrong).

javiermtorres (approved):
Only check the max_samples thing, please. Thanks!

lumigator/schemas/lumigator_schemas/experiments.py
njbrake: Checked and confirmed that it does belong in the experiments object. Thanks for the comment, was good to double check 🚀

PR #904, opened 2025-02-18
Use Protocols to define interfaces over inheritance

peteski22:
# What's changing

* Use Protocols to define interfaces for `TrackingClient` and `TrackingManager`
* No shared logic in base classes, no implementation in base classes
* Removes subclassing requirement, reducing tight coupling

# I already...


njbrake (approved):
On the technical side I'm not super familiar with protocols and the pro/cons vs ABC. However, the changes are passing unit tests and I don't see anything wrong with the changes, so I'll approve it 👍  Thanks for the work, @peteski22 !

javiermtorres:
Basically AFAICT this means moving to duck typing (similar to Go's interfaces). In the job/backend interface, it actually has advantages because making a base class available to jobs packages to import may be a bit complicated at this point.

javiermtorres (approved):

Issue #704, opened 2025-01-21
[BUG]: Unused tokenizer parameter `mod_max_length`


HareeshBahuleyan:
### Description

We have a tokenizer parameter `mod_max_length` in the code for HF models, which is used in some of the config_templates
https://github.com/mozilla-ai/lumigator/blob/ebca149116daec157db91622d056dc67335d232c/lumigator/python/mzai/jobs/evaluator/evaluator/configs/huggingface.py#L34
https://github.com/mozilla-ai/lumigator/blob/ebca149116daec157db91622d056dc67335d232c/lumigator/python/mzai/jobs/evaluator/evaluator/jobs/asset_loader.py#L166

Based on some research of the HF docs, `mod_max_length` is not a parameter. And the actual parameter name seems to be `model_max_length`
https://huggingface.co/docs/transformers/en/main_classes/tokenizer#transformers.PreTrainedTokenizer

So most likely the max_length parameter is not being used. Which means that model_max_length by default is `VERY_LARGE_INTEGER (int(1e30))`. This could be related to the Issue https://github.com/mozilla-ai/lumigator/issues/667 @agpituk  

### Reproduction

`mod_max_length` does not influence the model inference.

### Relevant log output

```shell

```

### Expected behavior

Use the correct parameter name `model_max_length` 

### System Info

- macOS
- Lumigator version: `1262577e5c5aac8a9b49f3190422105d69af2b51`

### Have you searched for similar issues before submitting this one?

- [x] Yes, I have searched for similar issues

agpituk:
@HareeshBahuleyan is this still an issue after merging #667 ?

HareeshBahuleyan:
@agpituk `mod_max_length` was mainly being used in `evaluator` which has now been removed thanks to #899 
Also got rid of the last few occurrences in some mock data with https://github.com/mozilla-ai/lumigator/pull/910
So, we are good to close this issue 👍

PR #910, opened 2025-02-18
chore: Update json data for pytest fixtures

HareeshBahuleyan:
# What's changing

* Remove `evaluator` usage from mock data json (use inference.py instead)
* Remove usage of `mod_max_length` which is a typo and not an actual parameter for huggingface transformers (Refs #704)

# How to test it

Steps to test the changes:

1.
2.
3.

# I already...


njbrake (approved):
Thanks! What uses these files, and why didn't the issue get caught in a unit/integration test?

HareeshBahuleyan:
@njbrake Its used in the unit test, but doesn't really check the exact contents but only the number of items expected, for example:

https://github.com/mozilla-ai/lumigator/blob/890d29cf1d041f8bd6710ff0d6651000cc5f68f5/lumigator/sdk/tests/unit/test_jobs.py#L70-L80

njbrake:
> @njbrake Its used in the unit tests, but doesn't really check the exact contents, for example here:
> 
> https://github.com/mozilla-ai/lumigator/blob/890d29cf1d041f8bd6710ff0d6651000cc5f68f5/lumigator/sdk/tests/unit/test_jobs.py#L70-L80

Did this break something you were using, or did you happen upon it while doing some other work? I'm trying to figure out if there is something else I should add to my normal testing to avoid making a mistake like this again 😬

HareeshBahuleyan:
No no, it didnt break anything, but encountered it while working on https://github.com/mozilla-ai/lumigator/issues/704 where I was searching for the occurences of `mod_max_length` - so found this coincidentally 😄

Issue #906, opened 2025-02-18
Test Lumigator support for external OpenAI-compatible APIs on Grok


ividal:
The request we received was to evaluate Grok (via its API) for a summarization task.

Following the experience with DeepSeek, this involves:
* access to the model via API: testing LiteLLM integration
* data: starting with the Thunderbird dataset
* evaluation: using currently supported ones
* nice to haves: 
    * using instead (or in addition) a non-OS dataset or one that has been recently released and may not have been used in the model's training
    * adding llm-as-a-judge metrics as they become available

njbrake:
https://github.com/mozilla-ai/lumigator/pull/911 has been merged, so this issue is ready to be worked on

PR #911, opened 2025-02-18
LiteLLM Support (Also DeepSeek in the UI and Llamafile in the UI)

njbrake:
# What's changing

Instead of manually making classes for each provider, use LiteLLM allow that library to manage all the complexities of provider support.

This will help give us support for additional models like Grok as well as self-hosted options like vLLM and Ollama.

WIth this PR, now the inference job uses LiteLLM.

I also renamed "engine" to "model" and "model_url" to be "base_url" to make it more clear that "model" is the model you're running while "base_url" has to do with where that model is running. Providers like OpenAI and Mistral will have their Base_urls handled by LiteLLM, while for things like Llamafile we handle the base_url to make sure it points to the right thing.

Refs 
#475 
#906 

# How to test it

If you follow the docs for running llamafile https://mozilla-ai.github.io/lumigator/user-guides/local-models.html, then you can use the UI and now llamafile works.

I think this may need a little more work on the frontend to correctly toggle llamafile avaiilability, not sure what the best way to do that will be: maybe only show the option in the UI if llamafile appears to be running?

# I already...


docs/source/conceptual-guides/new-endpoint.md
njbrake, line 1: I deleted this file because the guide was all about how to add a new route to the API which already had a variety of info that was out of date, and it also seemed like a really intense guide for adding a new route while we would want users to focus on making their own new jobs, not new routes.

lumigator/backend/backend/services/jobs.py
```diff
@@ -498,7 +432,7 @@ def create_job(
 
         # ...and use directly Job*Config(request.job.config.model_dump_json())
         job_config_args = {
-            "--config": job_config.model_dump_json(exclude_unset=True, exclude_none=True),
```
njbrake, line 501: I deleted these configs because this was stripping out parameters that were set to None, which then conflicts with the schema that is used during config creation as well as loading it again inside the inference job. I see these lines were added via

https://github.com/mozilla-ai/lumigator/pull/740/files#diff-59885862ae7b5c8dd1f02df9d48befc70d96c68b03e5308d24a45990f0e5bb22R501

@javiermtorres  would you be able to weigh in here about why this code was added in PR-740?

lumigator/backend/backend/services/workflows.py
```diff
@@ -126,7 +126,7 @@ async def _run_inference_eval_pipeline(
 
         # wait for the evaluation job to complete
         status = await self._job_service.wait_for_job_complete(evaluation_job.id, max_wait_time_sec=60 * 10)
-        self._job_service._validate_results(evaluation_job.id, self._dataset_service.s3_filesystem)
```
njbrake, line 129: Not sure how this worked because _validate_results expects three args, not two...

lumigator/backend/backend/services/jobs.py
javiermtorres: IIRC, including the None values did cause issues to me. If None values are meant as optional and have None values as default in the schema, this should not be an issue. Usually None values can be left out because, in serialization, we don't normally distinguish from a missing JSON key and an explicit `null` for said key; and we can get the list of allowed keys from the schema itself.

Let me know which values cause problems, and we can work it out in the schema. We'll also add a check for those in the backend unit tests.

lumigator/backend/backend/services/workflows.py
javiermtorres: AFAICT in the base code it was called with 2 args... Maybe it didn't get checked in at some point?

lumigator/backend/backend/models.yaml
```diff
@@ -33,7 +33,7 @@
         num_beams: 4
 
 - name: gpt-4o-mini
-  uri: oai://gpt-4o-mini
+  uri: openai/gpt-4o-mini
```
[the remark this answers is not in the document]
ividal: +1 to keeping some `schema:` and it seems the common denominator is OpenAI-compatible, so `oaiapi:`?

That for the scope of this PR. 

At some point, though, we have to see how to support users sending models that we do not have on this yaml file - or else it'll be playing catch up.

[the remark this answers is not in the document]
javiermtorres: oaiapi seems fine to me; as I said, if it's a uri we need a schema, but I don't have a strong preference. But maybe this is _not_ a uri after all, let's check.

lumigator/backend/backend/services/jobs.py
```diff
@@ -229,7 +229,7 @@ def _add_dataset_to_db(self, job_id: UUID, request: JobCreate, s3: S3FileSystem)
 
         loguru.logger.info(f"Dataset '{dataset_filename}' with ID '{dataset_record.id}' added to the database.")
 
-    def _validate_results(self, job_id: UUID, s3: S3FileSystem) -> JobResultObject:
+    def _validate_results(self, job_id: UUID, request: JobCreate, s3: S3FileSystem):
```
javiermtorres, line 232: AFAICT the request is not used within this method?

dpoulopoulos: And this also introduces a bug cause it's called like this in 194:

```
results = self._validate_results(job_id, s3)
```

dpoulopoulos (changes requested):

lumigator/jobs/inference/inference_config.py
```diff
@@ -100,7 +100,7 @@ class JobConfig(BaseJobConfig):
     max_samples: int = -1  # set to all samples by default
     storage_path: str
     output_field: str = "predictions"
-    enable_tqdm: bool = True
+    enable_tqdm: bool | None = True
```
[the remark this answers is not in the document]
njbrake: Good point. Yeah let me tweak this. Thanks! I'll keep the default as True though, since tqdm is the "progress bar" for python, which generally we like to have in the logs.

lumigator/backend/backend/models.yaml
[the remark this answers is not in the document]
njbrake: Interesting. So this model.yaml file goes through a few jumps of names. In models.yaml it's called "uri" but then when a request is made to the createWorkflow endpoint, it's renamed to 'model', and then when we pass it into the inference config, we rename it to "model_uri" if it's a HF:// model (HFPipelineConfig), or keep it named as "model" if it's a InferenceServerConfig.

So: how about we do a slight variation Javier says in bullet one of his recent message in slack:

`the models.yaml file will replace the "uri" field with values for "model" and "model_url" that can be directly set in the inference config to use those models`

But instead of calling it `model_url` we call it `provider`? 
This would mean that 

```suggestion
 model: gpt-40-mini
 provider: openai
```

The only slight complexity here is that we need some way to differentiate between huggingface as a hosting provieder https://docs.litellm.ai/docs/providers/huggingface vs huggingface as running locally in our inference job. I'll try to draft something up.

lumigator/frontend/src/stores/experimentsStore.ts
```diff
@@ -129,6 +129,7 @@ export const useExperimentStore = defineStore('experiments', () => {
           ...experimentData,
           experiment_id: experimentId,
           model: model.uri,
+          base_url: model.base_url,
```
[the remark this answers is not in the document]
njbrake: Currently yes, we could keep this all in the backend. However, then we would have to do this change as soon as we would want the user to be able to configure their own base_url (for instance, if they want to point to a custom vLLM server location). This change is made to support that use case that I see being requested in the near future :)

dpoulopoulos (approved):
I tested this PR by running an evaluation of `xai/grok`. A few issues I found have been addressed, so it looks good to me!

lumigator/backend/backend/models.yaml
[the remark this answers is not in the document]
javiermtorres: whatever we do, I'd vote to keep the api consistent (so if we rename it in models.yaml, let's rename it in the schema too)

HareeshBahuleyan (approved):
Thanks @njbrake for adding support for local/vLLM self-hosted models via LiteLLM 👏

PR #912, opened 2025-02-19
Backend: Handle workflow not found from mlflow client

peteski22:
# What's changing

Added error handling when attempting to get a workflow/run that doesn't exist in MLFlow. This was causing `500` internal server errors instead of returning `404`s.

In a few cases it seems the MLFlow API throws exceptions `RESOURCE_DOES_NOT_EXIST` when a resource cannot be found (that map to the HTTP `404` status code).

Note: `get_workflow` already returned `WorkflowDetailsResponse | None` but the signature didn't reflect that.

# How to test it

Steps to test the changes:

1. `make local-up`
2.  Hit the API endpoint for get workflow using any valid UUID (none will exist in mlflow)
3. Check the logs in the backend to ensure that we're now returning `404` (WorkflowNotFoundError) instead of `500`.

This doesn't seem to be well documented in MLFlow

Example: 

Declares the exception to be raised:

* https://mlflow.org/docs/latest/rest-api.html#get-experiment-by-name
* https://mlflow.org/docs/latest/rest-api.html#restore-experiment
* https://mlflow.org/docs/latest/python_api/mlflow.client.html#mlflow.client.MlflowClient.log_batch

No mention of exceptions:

* https://mlflow.org/docs/latest/python_api/mlflow.client.html#mlflow.client.MlflowClient.get_run
* https://mlflow.org/docs/latest/rest-api.html#get-run

😞 maybe time for a contribution to improve their docs.

# I already...


njbrake (approved):
Good catch. Coincidentally I found the same bug in the get_experiments code yesterday, but I forgot to apply the same patch to get_workflow. You can see my implementation of the handling in the mlflow:get_experiment function, might be worth unifying them to use the same logic. I think your logic looks better so probably the get_experiment mlflow function should be updated to use your code.

peteski22:
> Good catch. Coincidentally I found the same bug in the get_experiments code yesterday, but I forgot to apply the same patch to get_workflow. You can see my implementation of the handling in the mlflow:get_experiment function, might be worth unifying them to use the same logic. I think your logic looks better so probably the get_experiment mlflow function should be updated to use your code.

Thanks for the heads up @njbrake, I've pushed a commit to match them both up. 😄

Issue #921, opened 2025-02-19
Users can specify the task as ‘translation’ when creating an experiment


HareeshBahuleyan:
Include the following checks:
- Source (X) and target (Y) language needs to be provided when task = translation
- We auto-create a basic prompt “translate from X to Y:”, but user can override this

Language codes:
https://quickref.me/iso-639-1.html

rena-ertam028:
AI tried to solve the issue:

Added validation for translation tasks to ensure source and target languages are provided and properly formatted. Also added default prompt generation for translation tasks while preserving ability for users to override with custom prompts.

You can review changes in this commit: https://github.com/rena-ertam028/mozilla-ai-lumigator-921/commit/c162460803754fd29f138014ba7c900ef072de53.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

Latta AI seeks to solve problems in open source projects as part of its mission to support developers around the world. Learn more about our mission at https://latta.ai/ourmission . If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

PR #888, opened 2025-02-17
Initial job interface to the backend

javiermtorres:
# What's changing

This PR prepares a job interface between the backend and the jobs proper. The idea is to end up moving the subclasses to each of the job definitions, and possibly move some parameters across the job-backend boundary .

Refs #413 

# How to test it

CI should run without issues.

# Additional notes for reviewers

N/A

# I already...

- [ ] Tested the changes in a working environment to ensure they work as expected
- [ ] Added some tests for any new functionality
- [ ] Updated the documentation (both comments in code and [product documentation](https://mozilla-ai.github.io/lumigator) under `/docs`)
- [ ] Checked if a (backend) DB migration step was required and included it if required

PR #936, opened 2025-02-19
Implement task-specific validation and setting default prompts if none provided

HareeshBahuleyan:
# What's changing

- Treat task as Enum TaskType instead of str
- Validate task <-> source_language and target_language (no language pair should be specified for summarization, but both source_language and target_language required for translation)
- Setting default prompt at JobInferenceConfig: for summarization use DEFAULT_SUMMARIZER_PROMPT, for translation set default prompt based on language pair
- ^Fix current hardcoded system prompt

Refs #921 

# How to test it

(currently HF models are not fully supported, you have to test with LiteLLM models, i.e., causal LLM API models (e.g. OpenAI) 

Steps to test the changes:
1. make local-up
2. Upload sample dataset - use [this for translation en-de](https://github.com/mozilla-ai/lumigator/blob/4233c3c74b5106faed772941a7dfd8ca823cc199/lumigator/sample_data/sample_translation_en_de.csv) and dialog_sum for summarization
3. Run inference job (try different combinations of task and source_language and target_language). 

For example with summarization:
```
curl -X 'POST' \
  'http://localhost:8000/api/v1/jobs/inference/' \
  -d '{
  "name": "summarization",
  "description": "test",
  "dataset": "7e63bc7c-b03e-438d-a48e-f695d82ec25b",
  "max_samples": -1,
  "job_config": {
    "job_type": "inference",
    "model": "gpt-4o-mini",
    "provider": "openai",
    "task_definition": {
      "task": "summarization"
    }
  }
}'
```
This would run an inference job with the default prompt:
```
python inference.py --config '{"name":"summarization/de3c7b45-93f0-431b-b653-43d04d546751","dataset":{"path":"s3://lumigator-storage/datasets/7e63bc7c-b03e-438d-a48e-f695d82ec25b/sample_translation_en_de.csv"},"job":{"max_samples":-1,"storage_path":"s3://lumigator-storage/jobs/results/","output_field":"predictions","enable_tqdm":true},"system_prompt":"You are a helpful assistant, expert in text summarization. For every prompt you receive, provide a summary of its contents in at most two sentences.","inference_server":{"base_url":null,"model":"gpt-4o-mini","provider":"openai","system_prompt":"You are a helpful assistant, expert in text summarization. For every prompt you receive, provide a summary of its contents in at most two sentences.","max_retries":3},"params":{"max_tokens":1024,"frequency_penalty":0.0,"temperature":1.0,"top_p":1.0},"hf_pipeline":null}'
```
 
For example with translation:
```
curl -X 'POST' \
  'http://localhost:8000/api/v1/jobs/inference/' \
  -d '{
  "name": "translation",
  "description": "test",
  "dataset": "7e63bc7c-b03e-438d-a48e-f695d82ec25b",
  "max_samples": -1,
  "job_config": {
    "job_type": "inference",
    "model": "gpt-4o-mini",
    "provider": "openai",
    "task_definition": {
      "task": "translation",
       "source_language": "en",
       "target_language": "de"
    }
  }
}'
```
This would run an inference job with the default prompt with source and target languages:
```
python inference.py --config '{"name":"translation/6c4be683-69a3-471a-9d99-b8f8a3b51863","dataset":{"path":"s3://lumigator-storage/datasets/7e63bc7c-b03e-438d-a48e-f695d82ec25b/sample_translation_en_de.csv"},"job":{"max_samples":-1,"storage_path":"s3://lumigator-storage/jobs/results/","output_field":"predictions","enable_tqdm":true},"system_prompt":"translate en to de:","inference_server":{"base_url":null,"model":"gpt-4o-mini","provider":"openai","system_prompt":"translate en to de:","max_retries":3},"params":{"max_tokens":1024,"frequency_penalty":0.0,"temperature":1.0,"top_p":1.0},"hf_pipeline":null}'
```

# I already...


lumigator/schemas/lumigator_schemas/tasks.py
njbrake, line 1: Thinking here, I don't think I have an opinion yet, but should this code be in the jobs.py schema file instead of a new file? The structure of lumigator_schemas thus far is a schema matched to a route. A new file would be a departure from that design

lumigator/schemas/lumigator_schemas/tasks.py
```diff
@@ -0,0 +1,61 @@
...
+            raise ValueError("system_prompt must be provided for text generation tasks.")
+
+    def set_default_prompt(self, config):
+        # No default prompt is set, as the user is expected to provide one
+        # according to the task they want to perform.
+        pass
+
+
+class SummarizationValidator(TaskValidator):
+    DEFAULT_PROMPT: str = "You are a helpful assistant, expert in text summarization. For every prompt you receive, provide a summary of its contents in at most two sentences."  # noqa: E501
```
njbrake, line 34: Idk, if we're moving this someplace, should this validation go into the inference job schema instead of lumigator_schema? Might make more sense to move this closer to the logic that is running the model.

njbrake (commented):
Nice re-organization! I have a few questions mostly about where to move the logic.

lumigator/schemas/lumigator_schemas/tasks.py
```diff
@@ -0,0 +1,61 @@
+from abc import ABC, abstractmethod
+from enum import Enum
+
+
+class TaskType(str, Enum):
```
aittalam, line 5: I think it's great that we have a schema specifically for tasks with ad-hoc validators, thank you!
WDYT about being explicit here (e.g. with a comment) saying that the task names matches the HF ones, so if a HF model is called directly we'll use the right pipeline? Just as a reference for people who might update this code later

lumigator/schemas/lumigator_schemas/tasks.py
```diff
@@ -0,0 +1,61 @@
...
+        if config.source_language or config.target_language:
+            raise ValueError(
+                f"Fields source_language and target_language should not be provided when task={TaskType.SUMMARIZATION}"
+                f"but got source_language={config.source_language} and target_language={config.target_language}"
+            )
+
+    def set_default_prompt(self, config):
+        # We set the default prompt only if the user has not provided one
+        if config.system_prompt is None:
+            config.system_prompt = self.DEFAULT_PROMPT
```
aittalam, line 46: What happens with HF seq2seq models which do not require a prompt?

lumigator/schemas/lumigator_schemas/tasks.py
HareeshBahuleyan: Yes that would be closer to the logic running the model. However. it might be good to catch and throw the error early on - rather than having to wait till reaching the jobs.

lumigator/schemas/lumigator_schemas/tasks.py
HareeshBahuleyan: Good idea, will add a comment.

lumigator/schemas/lumigator_schemas/tasks.py
HareeshBahuleyan: You are right, current it isn't being used. Also this iteration won't support HF seq2seq models, but only causal LLM API models.
But in future, we will need the language pairs to be used as a "prefix" to every input sequence for seq2seq models. So in a later iteration, we could try to use construct a prefix out of it for seq2seq models and as prompt for causal models.

From https://huggingface.co/docs/transformers/en/tasks/translation:
image

HareeshBahuleyan: Created an issue to track it: https://github.com/mozilla-ai/lumigator/issues/976

lumigator/schemas/lumigator_schemas/tasks.py
aittalam: I kinda agree with both here :-)
On the one hand, a job which does not validate its own inputs opens itself to possible issues.
On the other hand, our inference job is kinda agnostic of its purpose and having task-specific validation here means bringing much more logic than is needed to "just run inference".
Also, I like the principle of breaking early rather than within the job: this allows us to get higher-quality errors directly in the client via the API.
For the above reasons I'd lean more towards keeping stuff where it is now rather than with jobs. WDYT?

lumigator/schemas/lumigator_schemas/tasks.py
aittalam: Cool, I think as long as this does not break current s2s models doing summarization this is ok (I guess they will just be assigned a prompt which is not used by the summarization pipeline, right?)

lumigator/schemas/lumigator_schemas/experiments.py
```diff
@@ -11,15 +12,15 @@ class ExperimentCreate(BaseModel):
     description: str = ""
     dataset: UUID
     max_samples: int = -1  # set to all samples by default
-    task: str | None = "summarization"
+    task: TaskType = Field(default=TaskType.SUMMARIZATION)
```
javiermtorres, line 15: If the default task type is summarization, shouldn't the default prompt also be the one for summarization?
Also, maybe we can use per-task defaults instead of a single default (we can take a look at a way to encode that in pydantic)

lumigator/schemas/lumigator_schemas/jobs.py
```diff
@@ -76,7 +78,9 @@ class JobInferenceConfig(BaseModel):
     job_type: Literal[JobType.INFERENCE] = JobType.INFERENCE
     model: str
     provider: str
-    task: str | None = "summarization"
+    task: TaskType = Field(default=TaskType.SUMMARIZATION)
```
javiermtorres, line 81: I feel like the inference task is being overloaded :-/ Annotations, summarizations and translations are all inferences. But they all handle different parameters, that in turn create the _same_ set of parameters for the inference task. How can we model this? Maybe with a specific `type` param that holds ann/summ/trans specific subparams?

lumigator/schemas/lumigator_schemas/tasks.py
javiermtorres: The less translation across dataclasses, the better ^_^

lumigator/schemas/lumigator_schemas/tasks.py
javiermtorres: > rather than having to wait till reaching the jobs.
the jobs schemas should be reachable from the backend. Please wait for https://github.com/mozilla-ai/lumigator/pull/888
@aittalam the generate_config is implemented by each job, so the backend is kept agnostic while allowing per-job checks at the same time 👍

lumigator/schemas/lumigator_schemas/jobs.py
HareeshBahuleyan: Sorry yes we probably need to break it down!
Some of this will go away from jobs in a [follow-up PR](https://github.com/mozilla-ai/lumigator/pull/971/files#diff-851bf090de2849773e9c7486bebd855b586c8a24f6b20cdd2eb10d65139dd1c2) that I am working on. 

Anyways I have created an issue to track this #977

lumigator/schemas/lumigator_schemas/experiments.py
HareeshBahuleyan: Yes, default prompt setting per task happens in jobs [here](https://github.com/mozilla-ai/lumigator/blob/0e44e43d281374dc2d36ccc04325743bdc65d1ad/lumigator/schemas/lumigator_schemas/jobs.py#L115) since prompt is available at the job level and not at the experiment level.

lumigator/schemas/lumigator_schemas/tasks.py
HareeshBahuleyan: Added to docstring.

lumigator/schemas/lumigator_schemas/tasks.py
ividal: > as long as this does not break current s2s models doing summarization
💯 If I am understanding this correctly, SummarizationValidator is assuming a default, non-empty, prompt and, when none is passed with the config, sets it. Wouldn't this fail with Falconsai/text_summarization and other seq2sew?

lumigator/schemas/lumigator_schemas/jobs.py
```diff
@@ -76,7 +78,9 @@ class JobInferenceConfig(BaseModel):
...
+    source_language: str | None = Field(None, description="Source language for translation", examples=["en", "English"])
+    target_language: str | None = Field(None, description="Target language for translation", examples=["de", "German"])
```
javiermtorres, line 83: I think you can do away with custom validators and keep summarization and translation apart with these models instead:
```python
class JobInferenceConfig(BaseModel):
[...]
    task_config: SummarizationTask | TranslationTask = Field(discriminator='task_type')
[...]
class SummarizationTask
    task_type: TaskType = Field(default=TaskType.SUMMARIZATION)
    system_prompt: str = SUMMARIZATION_DEFAULT_PROMPT
class TranslationTask
    task_type: TaskType = Field(default=TaskType.TRANSLATION)
    system_prompt: str = TRANSLATION_DEFAULT_PROMPT
    source_language: str = Field(description="Source language for translation", examples=["en", "English"])
    target_language: str = Field(description="Target language for translation", examples=["de", "German"])

```

javiermtorres: There would be one thing missing, setting the default task as summarization, which has all default values itself:
```
class JobInferenceConfig(BaseModel):
[...]
    task_config: SummarizationTask | TranslationTask = Field(SummarizationTask(), discriminator='task_type')
```

lumigator/schemas/lumigator_schemas/tasks.py
```diff
@@ -0,0 +1,52 @@
...
+        default=None, description="System prompt for translation tasks", examples=["translate English to German"]
+    )
+    source_language: str = Field(examples=["en", "English"])
+    target_language: str = Field(examples=["de", "German"])
+    model_config = ConfigDict(extra="forbid")
+
+    @model_validator(mode="after")
+    def set_translation_prompt(self):
+        if not self.system_prompt:
+            self.system_prompt = f"translate {self.source_language} to {self.target_language}:"
```
javiermtorres, line 40: I'd like to have a more declarative default instead of the validator, but we can think about that in a next PR

javiermtorres (approved):
LGTM

lumigator/schemas/lumigator_schemas/tasks.py
HareeshBahuleyan: Thanks all - will move the some of the validation at the experiment create stage (since the language pair could be relevant to have at the experiment level and to be available to both inference and eval) in the follow-up PR: https://github.com/mozilla-ai/lumigator/pull/971

lumigator/schemas/lumigator_schemas/tasks.py
HareeshBahuleyan: ^no it doesn't break s2s summarization model. You are right the default prompt is set when none is passed. But HF s2s pipeline will not use it, the model runs inference as usual without it. Tested with the following:
```
curl -X 'POST' \
  'http://localhost:8000/api/v1/jobs/inference/' \
  -d '{
  "name": "summarization",
  "description": "test",
  "dataset": "7e63bc7c-b03e-438d-a48e-f695d82ec25b",
  "max_samples": -1,
  "job_config": {
    "job_type": "inference",
    "provider": "hf",
    "task_definition": {
      "task": "summarization"
    }
  }
}'
```

We will circle back to this scenario on how to handle the prompt as prefix in iteration-2 when we try to introduce support for translation with HF models - issue #976

lumigator/schemas/lumigator_schemas/tasks.py
HareeshBahuleyan: Got it, will keep it in mind for the follow-up PR

lumigator/schemas/lumigator_schemas/jobs.py
HareeshBahuleyan: Thanks @javiermtorres for the suggestion ❤️  It indeed achieves the purpose with Pydantic without having to use additional custom methods.

PR #937, opened 2025-02-19
Add a timeout to the workflow API

### LINKED ISSUE #934: [BUG]: Long-running jobs time out after 600 seconds.
### Description

When I tried to evaluate Grok on [`dialogsum`](https://github.com/mozilla-ai/lumigator-datasets/tree/main/dialogsum) (#906), the inference job took more than 10 minutes to complete. However, there's an arbitrary timeout set in the code that interrupts the workflow after 600 seconds (10 minutes): https://github.com/mozilla-ai/lumigator/blob/3528f9c24cbd955be8a7e330fccdc74039a03167/lumigator/backend/backend/services/workflows.py#L128

As a result, the evaluation job never executed.

### Reproduction

To reproduce this behavior, start a long-running job, something that would take more than 600 seconds.

### Expected behavior

Since the inference job completed successfully on Ray, I expected the workflow to run the evaluation job next.

### System Info

- OS-independent
- Commit `3528f9c24cbd955be8a7e330fccdc74039a03167`

javiermtorres:
# What's changing

Jobs in a workflow currently expire at a fixed timeout of 600 seconds. Also, timed out jobs are not stopped. This PR provides a timeout field in the workflow that sets the waiting time for all jobs in the workflow, with the expected default value of 600 seconds. It should also stop jobs that have timed out.

Closes #934 

# How to test it

Steps to test the changes:

1. Post a new job with a very different timeout value than 600 (1 or 90000 for example)
2. Check that the job is set to the right status after the set time

N/A

# I already...

  - A new migration adding a new job status `STOPPED`to the job status enum is provided


lumigator/schemas/lumigator_schemas/workflows.py
```diff
@@ -28,6 +28,7 @@ class WorkflowCreateRequest(BaseModel):
     system_prompt: str | None = None
     inference_output_field: str = "predictions"
     config_template: str | None = None
+    job_timeout: NonNegativeInt = 60 * 10
```
njbrake, line 31: this parameter needs to have a unit given it. Maybe `job_timeout_sec`?

javiermtorres: LGTM

lumigator/backend/backend/services/workflows.py
```diff
@@ -53,6 +60,17 @@ def __init__(
...
+    # Maybe move to the job service?
+    def _stop_job(self, job_id: UUID):
+        resp = requests.post(urljoin(settings.RAY_JOBS_URL, f"{job_id}/stop"), timeout=5)  # 5 seconds
+        if resp.status_code == HTTPStatus.NOT_FOUND:
+            raise JobUpstreamError("ray", "job_id not found when retrieving logs") from None
+        elif resp.status_code != HTTPStatus.OK:
+            raise JobUpstreamError(
+                "ray",
+                f"Unexpected status code getting job logs: {resp.status_code}, error: {resp.text or ''}",
+            ) from None
```
njbrake, line 72: Hmm. Yeah this does seem like something that should be in the job service.

lumigator/backend/backend/services/workflows.py
```diff
@@ -87,11 +90,18 @@ async def _run_inference_eval_pipeline(
...
         if status != JobStatus.SUCCEEDED:
             loguru.logger.error(f"Inference job {inference_job.id} failed")
+            try:
+                self._job_service._stop_job(inference_job.id)
+            except JobUpstreamError:
+                loguru.logger.error(f"Failed to stop infer job {inference_job.id}, continuing")
             self._tracking_client.update_workflow_status(workflow.id, WorkflowStatus.FAILED)
-            raise Exception(f"Inference job {inference_job.id} failed")
+            # raise Exception(f"Inference job {inference_job.id} failed")
+            return
```
njbrake, line 104: Is this commented out by accident?

lumigator/schemas/lumigator_schemas/jobs.py
```diff
@@ -29,6 +29,7 @@ class JobStatus(LowercaseEnum):
     RUNNING = "running"
     FAILED = "failed"
     SUCCEEDED = "succeeded"
+    STOPPED = "stopped"
```
njbrake, line 32: My only question with this part, is does this impact any other logic? Like if logic is checking whether a job is either failed or succeeeded, they would now need to also check if it was stopped, since that's another valid end state.

njbrake (approved):
My only concern is whether there is any other logic that needs to handle the new STOPPED state that was added with the JobStatus. We may have other logic that is checking for stopped states of SUCCEED and FAIL, and may need to now also check for STOPPED.

I'll approve to keep things moving but please double check that part before the merge :) Thanks!

lumigator/schemas/lumigator_schemas/jobs.py
javiermtorres: Yes, this is also a terminal state. I'll double check where do we need to check this, because iirc there _is_ a list of terminal states that is used for some decision.

lumigator/backend/backend/services/workflows.py
javiermtorres: No. AFAICT the state is reported (as failed), and the requester can act on the state. There's no need to raise an exception at this point.

Issue #413, opened 2024-11-21
Add a new job type to clarify the job interface


javiermtorres:
**Is your feature request related to a problem? Please describe.**
Currently, adding a job means modifying several modules in the code without clear guidelines. It would be very helpful to have an interface definition (in an abstract sense) of what a job is, and how it is added to Lumigator minimizing the burden on the developer of a job. A new job could be developed from scratch, so all the necessary integration points could be noted down as an input for a possible job interface.

**Describe the solution you'd like**
A new job written from scratch and a list of all the new and modified files that were needed to provide it.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

njbrake:
I think this is related to https://github.com/mozilla-ai/lumigator/issues/706

PR #939, opened 2025-02-19
Make a clean interface for jobs

javiermtorres:
# What's changing

Work in a cleaner job/backend interface mechanism

Refs #413 

# How to test it

CI tests should work properly.

N/A

# I already...

  - Documentation about how to write a new job will be included in a follow up PR
  - No changes in DB


docker-compose.yaml
```diff
@@ -173,6 +173,7 @@ services:
       - RAY_WORKER_GPUS_FRACTION
       - LUMIGATOR_API_CORS_ALLOWED_ORIGINS
       - MLFLOW_TRACKING_URI
+      - PYTHONPATH=/mzai/lumigator/backend:/mzai/lumigator/jobs
```
njbrake, line 176: Seems like this should go into the dockerfile? This is required for the thing to work

njbrake (commented):
I think there was a bad merge because there are still some changes in here that we've discussed previously

docker-compose.yaml
javiermtorres: I don't like hard coding these paths into the docker file. We may need more flexibility. Let's talk about it on Monday.

javiermtorres:
@njbrake yes, I'll try to go through the usual suspects (mlflow, services/jobs)

docker-compose.yaml
javiermtorres: The idea behind this is to expose the jobs paths so another path can be mounted from outside the container, for user specified jobs. I'd actually remove the baked-in env var in the Dockerfile, but I decided not to do it in this PR just in case. We may need to come up with an alternative entrypoint; I'd also rather use a shell script instead of running uv directly.

javiermtorres: OTOH, since such flexibility is not really important for this specific PR at the moment, we can make a new issue about it and keep it hardcoded for the moment.

lumigator/backend/backend/tests/conftest.py
```diff
@@ -8,6 +8,8 @@
 from uuid import UUID
 
 import boto3
+import evaluator
+import evaluator.definition
```
njbrake, line 12: This is maybe a slightly nonstandard way to import stuff in python? Idk. I would have done something like `from evaluator import definition` or `from evaluator.definition import JOB_DEFINITION` but probably just my preference 🤷

lumigator/jobs/evaluator/definition.py
```diff
@@ -0,0 +1,42 @@
...
+
+# Eval job details
+# FIXME tweak paths in the backend
+EVALUATOR_WORK_DIR = "../jobs/evaluator"
+# FIXME maybe we can read the requirements file and tweak it in the backend
+# otherwise, we make another method in the job interface
+EVALUATOR_PIP_REQS = "../jobs/evaluator/requirements.txt"
+EVALUATOR_COMMAND = str = "python evaluator.py"
+
+JOB_DEFINITION: JobDefinition = JobInterfaceEvalLite(
```
njbrake, line 36:
```suggestion
JOB_DEFINITION: JobDefinition = JobInterfaceEvaluation(
```

lumigator/jobs/evaluator/definition.py
```diff
@@ -0,0 +1,42 @@
+from uuid import UUID
+
+from evaluator.schemas import DatasetConfig, EvalJobConfig, EvaluationConfig
+from lumigator_schemas.jobs import JobCreate, JobType
+
+from backend.services.job_interface import JobDefinition
+
+
+class JobInterfaceEvalLite(JobDefinition):
```
njbrake, line 9:
```suggestion
class JobInterfaceEvaluation(JobDefinition):
```

lumigator/jobs/inference/definition.py
```diff
@@ -0,0 +1,84 @@
...
+)
+from lumigator_schemas.jobs import JobCreate, JobType
+
+from backend.services.exceptions.job_exceptions import JobValidationError
+from backend.services.job_interface import JobDefinition
+
+# Served models
+OAI_API_URL: str = "https://api.openai.com/v1"
+MISTRAL_API_URL: str = "https://api.mistral.ai/v1"
+DEEPSEEK_API_URL: str = "https://api.deepseek.com/v1"
```
njbrake, line 19: We don't need this anymore. Some of this code looks like it's pre LiteLLM and some of Hareesh's hanges, may be worth double checking merges 😅 
```suggestion
```

lumigator/jobs/inference/definition.py
```diff
@@ -0,0 +1,84 @@
...
+            # Custom logic: if provider is hf, we run the hf model inside the ray job
+            job_config.hf_pipeline = HuggingFacePipelineConfig(
+                model_name_or_path=request.job_config.model,
+                task=request.job_config.task_definition.task,
+                accelerator=request.job_config.accelerator,
+                revision=request.job_config.revision,
+                use_fast=request.job_config.use_fast,
+                trust_remote_code=request.job_config.trust_remote_code,
+                torch_dtype=request.job_config.torch_dtype,
+                max_new_tokens=500,
```
njbrake, line 49: This should be gone, maybe a missing merge?

njbrake (commented):
the definition.py files seem like they might be out of date with what's in main, I think that needs to be double checked.

lumigator/backend/backend/tests/conftest.py
javiermtorres: We need to have a qualified JOB_DEFINITION since we need to use several, but otherwise I can probably take out the `{inference,evaluator}.definition` imports.

lumigator/jobs/evaluator/definition.py
javiermtorres: Argh, thanks a lot :-S

lumigator/jobs/inference/definition.py
javiermtorres: It's complicated to keep track of things that get moved _and_ modified :-/ I really don't know how to best handle these situations. Thanks a lot for the extra pair of eyes checking these issues.

javiermtorres: I'll check the definitions according to the latest main. If you spot something else, please let me know.

njbrake: With the codebase being relatively new and small, things are changing super fast, so I think the best way to avoid this is by having small PRs that get opened and merged quickly. The longer they sit open, the more they risk becoming out of date. I think this will be less of a problem as the codebase grows and the structure stabilizes.

lumigator/jobs/inference/definition.py
javiermtorres: Nah, as I mentioned, moved the code, so git doesn't realize the situation :-/
Double checking 👍

lumigator/jobs/inference/definition.py
javiermtorres: Furthermore, I'll wait for @HareeshBahuleyan 's PR on translation and then sync the job definitions.

docker-compose.yaml
javiermtorres: Removed

lumigator/backend/backend/tests/conftest.py
javiermtorres: I checked this specific use, and for clarity purposes I'd rather leave the qualified import, but just the `import evaluator` clause.

lumigator/jobs/inference/definition.py
javiermtorres: Ok, diff'ed with main and manually ported from main. It should be consistent but please give it a quick glance.

lumigator/jobs/inference/definition.py
javiermtorres: Manually checked as mentioned. Should be ok now.

lumigator/backend/backend/services/jobs.py
```diff
@@ -516,9 +361,9 @@ def create_job(
         # This includes both the command that is going to be executed and its
         # arguments defined in eval_config_args
         try:
-            job_settings = self.job_settings[job_type]
-        except KeyError as e:
-            raise JobTypeUnsupportedError("Unknown job type") from e
+            job_settings = job_settings_map[job_type]
+        except KeyError:
+            raise JobTypeUnsupportedError("Unknown job type") from None
```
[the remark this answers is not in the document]
javiermtorres: I'd say because in this particular case we have enough information about the error. Linking the keyerror does not provide additional info. We originate the exception, and it's not super complicated.

lumigator/jobs/inference/definition.py
```diff
@@ -0,0 +1,86 @@
...
+from inference.schemas import (
+    DatasetConfig,
+    GenerationConfig,
+    HuggingFacePipelineConfig,
+    InferenceJobConfig,
+    InferenceServerConfig,
+    JobConfig,
+)
+from lumigator_schemas.jobs import JobCreate, JobType
+from lumigator_schemas.tasks import TaskDefinition, get_default_system_prompt, validate_system_prompt
```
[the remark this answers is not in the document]
javiermtorres: We have two conflicting forces:

* Group together all the job-related info
* Group together all code running in the same place

I'm not sure the grouping we have right now is the right one. We could have `jobs/inference/plugin`and `jobs/inference/ray`, for example. I think we can decide in a followup PR. I'm not sure we'd want to put this in the backend folders, to avoid a user adding a new job being distracted with the backend code.

And actually, the code can be loaded by the job ;) just as the schemas are. I'm just not so sure that the requirements are so clear as in the case of schemas, but it should be fairly possible. Schemas work both ways because both envs have the same pydantic version.

Issue #501, opened 2024-12-13
[FEATURE]: Delete an experiment


ividal:
### Motivation

Currently, when experiments are listed on the UI, there is no way to select one and delete it, the way we can with datasets.

### Have you searched for similar issues before submitting this one?

- [x] Yes, I have searched for similar issues

javiermtorres:
If an experiment would be deleted, what would happen to jobs that were dependent on that experiment? Would they be orphaned (their parent would be `NULL`) or would they be deleted as well?

PR #953, opened 2025-02-20
feat: delete workflow/experiment

khaledosman:
# What's changing

Closes https://github.com/mozilla-ai/lumigator/issues/501

Should be merged after https://github.com/mozilla-ai/lumigator/pull/947

- Implements new options in `ExperimentDetails`, `JobDetails`, `JobTable` and `ExperimentTable` to delete workflow and/or delete experiment depending on selected item.
- Adds "view results" option in `JobTable` and `ExperimentTable` to view experiment results thats only available for successful runs
- Updates results page to show a relevant title for the results being displayed
- Fixes the `model` not being shown when clicking on a workflow, and rename "job id" to "model run id"

# How to test it

Steps to test the changes:

1. Click on a workflow
2. Click on the delete icon from the side drawer
3. Click on an experiment
4. Click on the delete icon from the side drawer
5. Click on the "..." for an experiment
6. Try to delete the experiment from there
7. Click on the "..." for a workflow
8. Try to delete the workflow from there (if its the only workflow left, the experiment should be deleted)
9. Click on "View results" for a workflow
10. Click on "View results" for an experiment

image
image
image
image
image

# I already...


hasangzl:
@khaledosman Just tested it the branch. All looks good to me.

Minor comment: was it not easy to get rid of the X in the pop-up? Otherwise we have two cancel buttons for the same action in the same pop-up (Applies to Deleting Datasets).

hasangzl (approved):

khaledosman:
> Minor comment: was it not easy to get rid of the X in the pop-up? Otherwise we have two cancel buttons for the same action in the same pop-up (Applies to Deleting Datasets).

The "X" button should be removed now in that [latest commit](https://github.com/mozilla-ai/lumigator/pull/953/commits/f056bfaab2f26ca3d854698c91f56f2aa3c06c51)

Although personally it doesn't bother me as much, I think its common to have the "X" button in these modals even though there's another "cancel" option, for example here's the modal from github 
image

PR #938, opened 2025-02-19
Environment creation/dependency sync

peteski22:
# What's changing

Provide a simple way to do the following via `make setup`:

* Install `uv` if it's not available
* Create `.venv`s for each `uv` managed project (with a `pyproject.toml`)
* Sync all dependencies for each created venv

Updated docs to include info on running `make setup`

# How to test it

Steps to test the changes:

1. Clone the repo
2. Remove all existing .venvs with: `find . -type d -name ".venv" -exec rm -rf {} +`
3. Run `make setup`

# Additional notes for reviewers

Optionally remove `uv` from your system entirely before testing:

```
uv cache clean
rm -r "$(uv python dir)"
rm -r "$(uv tool dir)"
rm ~/.local/bin/uv ~/.local/bin/uvx
rm ~/.cargo/bin/uv ~/.cargo/bin/uvx
```

# I already...

- [x] Tested the changes in a working environment to ensure they work as expected
- [ ] Added some tests for any new functionality
- [x] Updated the documentation (both comments in code and [product documentation](https://mozilla-ai.github.io/lumigator) under `/docs`)
- [x] Checked if a (backend) DB migration step was required and included it if required

PR #954, opened 2025-02-20
Environment creation: improve `make setup`

peteski22:
# What's changing

TL;DR: more robust checking for path + `uv`install...

* Checks if expected installation directory is part of `PATH` and adds it (temporarily if not)
* Gives a heads up about adding it permanently so users can use `uv` directly outside of `make setup` 
* Installs UV if it cannot be found (now includes checking the expected install location in `PATH`)
* Passes `PATH` to the script

This should remove any potential errors where re-running sometimes causes warnings about shadowing path locations, and explains why `uv` doesn't appear to exist in terminal (after install) unless it's added to `PATH` in user's profile

Refs: https://github.com/mozilla-ai/lumigator/pull/938

# How to test it

Steps to test the changes:

1. See test steps for https://github.com/mozilla-ai/lumigator/pull/938
2. Repeat with uv installed and not installed (and installed by the script)

# I already...


njbrake (approved):

Makefile
```diff
@@ -160,12 +160,25 @@ clean-docker-all: clean-docker-containers clean-docker-buildcache clean-docker-i
...
+		echo ""; \
+	fi && \
+	command -v uv >/dev/null 2>&1 || { \
 		echo "uv not found. Installing..."; \
 		curl -LsSf https://astral.sh/uv/install.sh | sh; \
-		export PATH="$(HOME)/.local/bin:$(PATH)"; \
-	}
-	@scripts/setup_venvs.sh;
+	} && \
+	PATH="$$LOCAL_BIN:$$PATH" scripts/setup_venvs.sh;
```
njbrake, line 181: Can this logic go into the shell script?

scripts/setup_uv.sh
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env bash
+
+set -e  # Exit on error
+
+#!/bin/bash
+
```
njbrake, line 6: I haven't seen something like this before: why #!/usr/bin/env bash and then #!/bin/bash?

Issue #956, opened 2025-02-20
[BUG]: Documentation: examples in the running local models section need to be updated after LiteLLM new syntax


aittalam:
### Description

After the introduction of LiteLLM, a new  required `provider` parameter has been introduced that is not present in the examples shown [here](https://github.com/mozilla-ai/lumigator/blob/main/docs/source/user-guides/local-models.md) so the examples break at the moment.

### Reproduction

Try to run one of the examples on https://github.com/mozilla-ai/lumigator/commit/b56a47a8344600ace7f6298b9715afa3db18bdcb. This results in the following error:

```
{
  "detail": [
    {
      "type": "missing",
      "loc": [
        "body",
        "job_config",
        "provider"
      ],
      "msg": "Field required",
      "input": {
        "model": "llamafile://mistralai/Mistral-7B-Instruct-v0.2",
        "model_url": "http://localhost:8080/v1",
        "system_prompt": "You are a helpful assistant, expert in text summarization. For every prompt you receive, provide a summary of its contents in at most two sentences.",
        "store_to_dataset": true
      }
    }
  ]
}
```

### Expected behavior

The example should be updated to include the new parameter (and any other extra change introduced by LiteLLM).

Adding an example of how I could make this work:

```
EVAL_NAME="test_inference_llamafile"
EVAL_DESC="Test inference with llamafile's Mistral"
EVAL_MODEL="openai/llamafile/Mistral-7B-Instruct-v0.2"
EVAL_MODEL_URL="http://localhost:8080/v1"
EVAL_SYSTEM_PROMPT="You are a helpful assistant, expert in text summarization. For every prompt you receive, provide a summary of its contents in at most two sentences."
EVAL_DATASET=$DATASET_ID
EVAL_MAX_SAMPLES="10"

JSON_STRING=$(jq -n \
                --arg name "$EVAL_NAME" \
                --arg desc "$EVAL_DESC" \
                --arg model "$EVAL_MODEL" \
                --arg dataset_id "$EVAL_DATASET" \
                --arg max_samples "$EVAL_MAX_SAMPLES" \
                --arg model_url "$EVAL_MODEL_URL" \
                --arg system_prompt "$EVAL_SYSTEM_PROMPT" \
                '''{name: $name,
                description: $desc,
                dataset: $dataset_id,
                max_samples: $max_samples,
                job_config: {
                  model: $model,
                  provider: "openai",
                  base_url: $model_url,
                  system_prompt: $system_prompt,
[...]

PR #957, opened 2025-02-20
LiteLLM should use the provider key

njbrake:
# What's changing

@aittalam mentioned some docs updates needed for LiteLLM and while answering that question I realized there is a bug in the code where we're not using the provider key in LiteLLM. this doesn't cause any issues when the provider is OpenAI compatible but does raise issues if using the deepseek api, mistral, etc.

Refs #956 

# How to test it

# I already...


aittalam (approved):
This is great, thanks @njbrake !!!

PR #963, opened 2025-02-20
Add basic keys infrastructure

### LINKED ISSUE #925: [FEATURE]: API: Settings endpoint for API keys
### Motivation

In order to allow users of Lumigator (via UI, SDK or REST API) to update their API keys for use in upstream services, we need to expose a new API endpoint.

* New endpoint: `PUT` `/settings/secrets/{key_name}`
* Only allow users to submit their key, not to retrieve values via the API
* Must be documented
* Must be unit tested

### Alternatives

_No response_

### Contribution

_No response_

javiermtorres:
# What's changing

This PR provides basic support for key handling:

* New table holding keys in plain
* New internal service for keys
* Endpoint for `PUT`ting keys (but not `GET`ting)
* One unit test case including DB checks

Closes #925 

# How to test it

A unit test testing across all layers is included.

N/A

# I already...

  - New revision added


Makefile
```diff
@@ -204,7 +204,7 @@ test-backend-unit:
 	SQLALCHEMY_DATABASE_URL=sqlite:////tmp/local.db \
 	MLFLOW_TRACKING_URI=http://localhost:8001 \
 	PYTHONPATH=../jobs:$$PYTHONPATH \
-	LUMIGATOR_SECRET_KEY=abcdefghijklmnopqrstuvwxyz123456 \
+	LUMIGATOR_SECRET_KEY=7yz2E+qwV3TCg4xHTlvXcYIO3PdifFkd1urv2F/u/5o= \
```
javiermtorres, line 207: maybe put the secret also in plain, if we need it?

lumigator/backend/backend/api/router.py
```diff
@@ -12,3 +12,4 @@
 api_router.include_router(experiments.router, prefix="/experiments", tags=[Tags.EXPERIMENTS])
 api_router.include_router(models.router, prefix="/models", tags=[Tags.MODELS])
 api_router.include_router(workflows.router, prefix="/workflows", tags=[Tags.WORKFLOWS])
+api_router.include_router(secrets.router, prefix="/settings/secrets", tags=[Tags.SETTINGS])
```
javiermtorres, line 15: Maybe......
```suggestion
api_router.include_router(secrets.router, prefix="/settings/secrets", tags=[Tags.SECRETS])
```
This is only meant for swagger docs grouping, so the current `SETTINGS` would also be fine to me.

peteski22: I wonder if we should leave it as `SETTINGS` for now as it covers the higher-level grouping and then as you say in Swagger docs it means if we add other settings related endpoints they'll just appear together?

peteski22 (approved):
👍🏼

lumigator/backend/backend/api/routes/secrets.py
```diff
@@ -0,0 +1,46 @@
...
+        status.HTTP_204_NO_CONTENT: {"description": "Secret successfully updated"},
+    },
+)
+def upload_secret(
+    service: SecretServiceDep,
+    secret_upload_request: SecretUploadRequest,
+    request: Request,
+    response: Response,
+    secret_name: str,
+) -> None:
```
khaledosman, line 32: is this doing an `upsert`? it might be weird to get a `201 CREATED` or a `204 NO CONTENTz` when doing an update

I would expect this to return the updated/created item instead, and ideally have separate endpoints with different http methods for create vs update to be more RESTful

Is this a real `PUT` where the consumer needs to pass everything including `name` & `description` for example when changing a secret or do we want to allow more fine-grained partial updates via a `PATCH`?

javiermtorres: Being a secret, iirc this would be a hard put. You'll need to specify everything, plus won't get the value since it's, er, secret. Yes, it's an upsert, so If the value exists, it's overwritten, hence the 201/204.

Issue #810, opened 2025-02-05
Re-design how we pair releases with the docker-compose


njbrake:
The below thread introduces an issue that think will re-occur in the future if we don't change anything. 

tl;dr our docker image releases are attached to specific versions of the docker-compose file. So if you run released version 1 of lumigator docker image using the docker-compose file from version 2 (aka main branch latest commit) they are not guaranteed to be compatible.

> Hello there! I want to recap and ensure we understand why this happened.
> Currently, the command `start-lumigator ` doesn't point to the `latest` tag but to the MVP tag. The docker image of the backend was created before the folder reorganization was made #717 . In #717 we changed docker-compose to use the new paths, but the MVP backend image still had the folders organized in the old way, therefore, making it impossible to find the evaluator when launching an experiment in ray.
> 
> The fix itself, was to override the old docker image tagged for the MVP with the latest version in main (we couldn't bump versions or the old image was going to always be broken).
> 
> This poses a question around if we should use the latest tag in lumigator or not, but this bug, as it is, is resolved now!
> @njbrake and @khaledosman , if you are happy with this, feel free to close the bug (I'm leaving it open in case you've got any more questions) 

 _Originally posted by @agpituk in [#795](https://github.com/mozilla-ai/lumigator/issues/795#issuecomment-2635161197)_

What would be a good way to restructure our release process to account for this?

One option I can think of: Main branch of lumigator docker-compose file always points to the `latest` tags of lumigator, and every release gets its own git branch/tag where the values in the docker-compose are set to be the release version. This would ensure that every release is run with the docker-compose file that existed when the release was made.

WDYT @agpituk ?

agpituk:
It's tough! Part of me thinks that we need a way to point to a _stable_ release (which the other day wasn't that stable), other part of me says, because we're moving really fast, you should get the latest version always or you'll be missing out. 

My thinking now is that if we keep things as they are, we need a way that in every commit to main, we test the stable containers, not only main code. In that sense, it would be way easier to just run tests on main (to avoid the overkill of testing things twice) and always point to latest. 
I also like the idea of getting containers with a git tag. 


My approach would be, to keep things as they are now (because they should be stable) but as soon as we add E2E tests (that @khaledosman already started working on and I know @javiermtorres wants to help with) we move the tag to latest. By doing this, we'll keep always the latest version of lumigator (images) aligned to what we're testing (main branch) and we'll have enough security that the product works properly.

njbrake:
[...]

PR #975, opened 2025-02-21
Changed containers in docker-compose to use latest tag

agpituk:
We're changing how we run start-lumigator to ensure we always pick the latest version of the containers. As it happened twice, we've made important changes to the environment variables in our docker-compose for ray. The tag for the backend and frontend container have code that need the variables in an older version (the same git tag as the containers).

This prs ensures we always point to the latest images in the containers

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -128,7 +128,7 @@ services:
   backend:
     labels:
       ai.mozilla.product_name: lumigator
-    image: mzdotai/lumigator:v0.1.1-alpha
+    image: mzdotai/lumigator:latest
     pull_policy: always
     build:
       context: .
@@ -188,7 +188,7 @@ services:
     labels:
       ai.mozilla.product_name: lumigator
     pull_policy: always
-    image: mzdotai/lumigator-frontend:v0.1.1-alpha
+    image: mzdotai/lumigator-frontend:latest
     build:
       context: .
       dockerfile: "./lumigator/frontend/Dockerfile"


macaab26 (approved):

njbrake (approved):
tl;dr: this will fix the problem we saw today, but it will introduce a problem next release, unless we change the procedure for making a release.

LGTM in terms of fixing the mismatch.

However, keep in mind this will break checking out a release now, unless you make sure that when you make the release, the docker-compose.yaml in that release git commit is specifying the image version, and not the latest tag.

njbrake:
This PR relates to https://github.com/mozilla-ai/lumigator/issues/810#issue-2832964689:

>One option I can think of: Main branch of lumigator docker-compose file always points to the latest tags of lumigator, and >every release gets its own git branch/tag where the values in the docker-compose are set to be the release version. This >would ensure that every release is run with the docker-compose file that existed when the release was made.

If we merge this, we need to make sure that we implement the full suggestion of that comment

agpituk:
> tl;dr: this will fix the problem we saw today, but it will introduce a problem next release, unless we change the procedure for making a release.
> 
> LGTM in terms of fixing the mismatch.
> 
> However, keep in mind this will break checking out a release now, unless you make sure that when you make the release, the docker-compose.yaml in that release git commit is specifying the image version, and not the latest tag.

100%, release process will have to change to avoid this from happening again. Let's draft a new one based on this next week before the next release (but, I think, in Main, from now on we should always point to the latest tag)

PR #980, opened 2025-02-21
Settings: Support generation/usage of encryption key

### LINKED ISSUE #948: [FEATURE]: Supply or generate+store AES symmetric key for Lumigator on startup
### Motivation

In order to encrypt sensitive API key information which we plan to store in Lumigator's DB we need a symmetric key which will be used to encrypt API keys, and decrypt them for the internal app to send to Ray.

Makefile:

* Has path to key file location configured
* Checks for existence of key on `local-up` or `start-*` targets
* If key doesn't exist generate and store it in the key file location
* Supply key as an env var to Lumigator's backend
* Output information on key file location 
* Document the process!

NOTE: Consider if the key value path should be set in .default.conf

### Alternatives

_No response_

### Contribution

_No response_

peteski22:
# What's changing

On starting Lumigator via `make local-up` or `make start-lumigator*` we now check for the existence of a key file in the Lumigator 'dot folder' in the user's home folder (`~/.lumigator/lumigator.key`). 

If the key isn't found we generate it and store it.

The key is used to create the temp `.env` file supplied to Docker Compose for deployment, which sets the `LUMIGATOR_SECRET_KEY` on the Backend container (and the backend app pulls it into it's settings).

Closes #948 

# How to test it

Steps to test the changes:

* Start Lumigator and check the terminal output:
```bash
make local-up
```
* Check the generated key value (to compare with the rest):
```bash
cat ~/.lumigator/lumigator.key
```
* Check the `build/.env` file: 
```bash
cat build/.env | grep LUMIGATOR_SECRET_KEY
```
* Check the docker compose output: 
```bash
docker-compose --log-level ERROR -f docker-compose.yaml -f .devcontainer/docker-compose.override.yaml --profile local --env-file build/.env config | yq '.services.backend.environment' | grep LUMIGATOR_SECRET_KEY
```
* Check the backend container's env var:
```bash
docker exec -it lumigator-backend-1 env | grep LUMIGATOR_SECRET_KEY
```
* Stop Lumigator and check the temp build file containing the key is gone (but not the user key):
```
make local-down
cat build/.env
cat ~/.lumigator/lumigator.key
```

# I already...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,15 +16,15 @@ KEEP_CONTAINERS_UP ?= "FALSE"
 
 # Configuration to identify the input and output config files
 # NOTE: Changing CONFIG_BUILD_DIR will require review of .gitignore
-CONFIG_BUILD_DIR="$(shell pwd)/build"
+CONFIG_BUILD_DIR=$(shell pwd)/build
 # Path to default config prefixed with dot to hide from user
-CONFIG_DEFAULT_FILE:="$(shell pwd)/.default.conf"
+CONFIG_DEFAULT_FILE:=$(shell pwd)/.default.conf
 # Directory for user specific configuration and keys
-CONFIG_USER_DIR:="${HOME}/.lumigator"
+CONFIG_USER_DIR:=${HOME}/.lumigator
 # Path to user editable config file (will be generated if missing)
-CONFIG_OVERRIDE_FILE:="$(CONFIG_USER_DIR)/user.conf"
+CONFIG_OVERRIDE_FILE:=$(CONFIG_USER_DIR)/user.conf
 # Location of the key to use for encryption.
-CONFIG_USER_KEY_FILE:="$(CONFIG_USER_DIR)/lumigator.key"
+CONFIG_USER_KEY_FILE:=$(CONFIG_USER_DIR)/lumigator.key
 # Env var name to hold encryption key
 CONFIG_USER_KEY_ENV_VAR=LUMIGATOR_SECRET_KEY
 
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -1,11 +1,11 @@
...
+.PHONY: local-up local-down local-logs clean-docker-buildcache clean-docker-images clean-docker-containers start-lumigator-external-services start-lumigator start-lumigator-postgres stop-lumigator test-sdk-unit test-sdk-integration test-sdk-integration-containers test-sdk test-backend-unit test-backend-integration test-backend-integration-containers test-backend test-jobs-evaluation-unit test-jobs-inference-unit test-jobs test-all config-clean config-generate-env setup config-generate-key
 
 SHELL:=/bin/bash
 UNAME:= $(shell uname -o)
 
 # Required binaries in order to correctly run the makefile, if any cannot be found the script will fail.
 # uv is only required for local-up (dev).
-REQUIRED_BINARIES := git docker
+REQUIRED_BINARIES := git docker openssl
```
njbrake, line 8: Any version requirements on this? I remember there being some pretty big changes between v1/v2/v3, not sure if that's important to consider.

njbrake (approved):
Only one comment about openssl versions.

Using openssl seems like it's going to tie us a lot tighter into mac/linux, so this may be something to re-evaluate if we try to support windows. Idk

Makefile
peteski22: I think for `rand` it should be fine across versions 🤞🏼

peteski22:
> Only one comment about openssl versions.
> 
> Using openssl seems like it's going to tie us a lot tighter into mac/linux, so this may be something to re-evaluate if we try to support windows. Idk

Yeah, I think we might need to re-visit the whole Windows support in general at some point soon.

PR #982, opened 2025-02-21
[DOCS] Write some descriptions for the apis

njbrake:
# What's changing

Woo! I think this is the last chunk to get our docs up to date; this PR has the high level descriptions about what each set of routes do. It serves as an intro to the OpenAPI docs that are automatically generated.

# How to test it

Reading

# I already...


docs/source/api/experiments.rst
```diff
@@ -1,6 +1,15 @@
 Experiments API
 ===============
 
+The Experiments API allows you to manage experiments within the system.
```
davidmanzanoai, line 4: I think we should introduce what we understand by "experiment" which the execution of different models (each of them will a job) with the same dataset prepared for a concrete use case (or similar definition) and whatever we add into the experiment. If we are following an internal structure to hold the experiment information it is good to include or add a link to it (but we shouldn't overcomplicate the documentation)

docs/source/api/models.rst
```diff
@@ -1,6 +1,12 @@
 Models API
 ============
 
+Lumigator provides a list of suggested models for each task. By calling the model routes you can retrieve suggested
```
davidmanzanoai, line 4: I would add after "each task" (for instance, text summarization).

docs/source/api/experiments.rst
njbrake: Sounds good! I updated the description, let me know what you think :)

Issue #862, opened 2025-02-13
Extend our eval library / workflow code to support LLM-as-judge


aittalam:
See rationale [here](https://github.com/mozilla-ai/lumigator/issues/802#issuecomment-2656463449).

The goal of this task is to have all the code in place to run a pipeline which provides LLM-as-judge evaluation of a given dataset (following the design we aligned on and extending eval library with some aggregate method to have e.g. mean score from models)

aittalam:
Following up from [this](https://github.com/mozilla-ai/lumigator/issues/802?issue=mozilla-ai%7Clumigator%7C861), I have investigated [deepeval's implementation of g-eval](https://docs.confident-ai.com/docs/metrics-llm-evals). Here are a few notes:

- we can use oai models out of the box by passing any model listed in [here](https://github.com/confident-ai/deepeval/blob/main/deepeval/models/gpt_model.py#L29). This could be good enough to test our workflow, but we'll need to make sure our feature comes with a good OSS model out-of-the-box
- oai-api-compatible models are supported through the "local model" approach, which is not exactly straightforward as it relies on creating a `.deepeval` json file with the parameters one needs to connect to a local model. As we are already relying on LiteLLM to transparently hit different kind of models in our inference job, we might want to try it here too
- evaluation steps and criteria are not provided by default, only a [general template is available](https://github.com/confident-ai/deepeval/blob/main/deepeval/metrics/g_eval/template.py). This is not a big problem, as we can start with [these prompts for g-eval](https://github.com/microsoft/promptflow/tree/main/examples/flows/evaluation/eval-summarization/prompts)
- footprint: python libs take about 500GB when installed in isolation, it might be much less in our existing ML eval environment

Some code to test this approach is available [here](https://github.com/mozilla-ai/lumigator/tree/862-add-geval-metric-to-eval-job), will soon send a PR with all four g-eval metrics so we can start testing

Issue #922, opened 2025-02-19
When the task is translation, eval job computes only translation-specific metrics


HareeshBahuleyan:
Similarly, when the task is summarization, eval job computes only summarization-specific metrics

Summarization specific metrics: ROUGE, METEOR, Bertscore

Translation specific metrics: BLEU, TER, Bertscore

PR #995, opened 2025-02-24
862 add geval metric to eval job

aittalam:
# What's changing

* Added a first implementation of [G-Eval](https://github.com/nlpyang/geval), relying on the implementation provided by [deepeval](https://docs.confident-ai.com/docs/metrics-llm-evals)
* Introduced a dictionary specifying, for each metric, which fields are required (as some will use the model inputs too, others won't need GT)
* Provided prompts for the summarization task as a json file which is easily generalizable to other tasks
* Defined schema for results

Refs #862 

# How to test it

Steps to test the changes:

1. Upload a dataset
2. Run an inference job to generate predictions and save a dataset
3. Run an evaluation job with the added `g_eval_summarization` metric, e.g.

```
fi

EVAL_NAME="test_evaluation_geval"
EVAL_DESC="Test evaluation with the geval metric"
EVAL_DATASET=$DATASET_ID
EVAL_MAX_SAMPLES="10"

JSON_STRING=$(jq -n \
                --arg name "$EVAL_NAME" \
                --arg desc "$EVAL_DESC" \
                '''{name: $name,
                description: $desc,
                dataset: $dataset_id,
                max_samples: $max_samples,
                job_config: {
                    "job_type": "evaluator",
                    "metrics": [
                      "meteor",
                      "rouge",
                      "bertscore",
                      "g_eval_summarization"
                    ]
                }}''' )

curl -s $BACKEND_URL/api/v1/jobs/evaluator/ \
  -d "$JSON_STRING" | jq
```

### What this PR does not address

This PR focuses on the new metric rather than how that is included in the workflow. The main reason is that we want this to work first, then test it extensively (especially with OSS models), and only at that point introduce it as a feature.

For this reason, I have not added yet:
- integration tests
- notebook / documentation updates
- an `EvaluationConfig` update in the API to expose the new metric

# I already...


lumigator/jobs/evaluator/evaluator.py
```diff
@@ -51,9 +51,11 @@ def save_outputs(config: EvalJobConfig, eval_results: JobOutput) -> Path:
 
 
 @timer
-def run_eval_metrics(predictions: list, ground_truth: list, evaluation_metrics: list) -> EvalJobMetrics:
+def run_eval_metrics(
+    llm_inputs: list, predictions: list, ground_truth: list, evaluation_metrics: list
```
HareeshBahuleyan, line 55: Curious as to why we name the variable `llm_inputs`?
How about using the same terminology from datasets, i.e., `examples`? And if we provide a doc-string describing the function arguments, that would be helpful,

lumigator/jobs/evaluator/eval_metrics.py
```diff
@@ -94,10 +122,85 @@ def _bertscore(self, pred: list, ref: list):
...
+            # iterate on all samples
+            for p, r, e in zip(pred, ref, llm_inputs):
+                test_case = LLMTestCase(input=e, expected_output=r, actual_output=p)
+
+                try:
+                    result = self._g_eval_measure_with_retry(metric, test_case, max_retries=MEASURE_RETRIES)
+                    evals_for_metric.append(result)
+                except ValueError as e:
+                    # Handle the failure case
+                    logger.error(f"G-Eval measurement failed after {MEASURE_RETRIES} attempts: {e}")
```
HareeshBahuleyan, line 189: Incase the failure is because the user has not set the OPENAI_API_KEY, would that be clear from the error message. If not, it would be good to pinpoint the error cause and maybe point the user to the guide on how to set it within lumigator: https://mozilla-ai.github.io/lumigator/operations-guide/configuration.html#settings

lumigator/jobs/evaluator/eval_metrics.py
```diff
@@ -1,17 +1,45 @@
...
-            "bertscore": self._bertscore,
-            "bleu": self._bleu,
+            "rouge": {"method": self._rouge, "requires": [EvaluationFields.GROUND_TRUTH]},
+            "meteor": {"method": self._meteor, "requires": [EvaluationFields.GROUND_TRUTH]},
+            "bertscore": {"method": self._bertscore, "requires": [EvaluationFields.GROUND_TRUTH]},
+            "bleu": {"method": self._bleu, "requires": [EvaluationFields.GROUND_TRUTH]},
+            "g_eval_summarization": {
+                # the available tasks in g_eval are the ones we have defined
+                # criteria / evaluation steps for inside `g_eval_prompts.json`
+                "method": functools.partial(self._g_eval, task="summarization"),
```
HareeshBahuleyan, line 40: I see the task is hard-coded which works for this iteration of LLM-as-a-judge, something to think about is how to get it from upstream, like from the backend.
https://github.com/mozilla-ai/lumigator/blob/93145b243e59f2e1ff5acb6ea11959154bbe8f46/lumigator/schemas/lumigator_schemas/tasks.py#L7-L16

lumigator/jobs/evaluator/eval_metrics.py
```diff
@@ -94,10 +122,85 @@ def _bertscore(self, pred: list, ref: list):
...
+
+        See: https://github.com/nlpyang/geval (G-Eval)
+        See: https://docs.confident-ai.com/docs/metrics-llm-evals (deepeval)
+        """
+        # Load task-dependent criteria / evaluation steps
+        with Path(G_EVAL_PROMPTS).open() as f:
+            prompt_templates = json.load(f)
+
+        evals = {}
+        for metric_name in prompt_templates[task]:
```
HareeshBahuleyan, line 166: Maybe we can have a check with suitable error message if task is not found in prompt_templates?

HareeshBahuleyan (commented):
Thanks @aittalam! Overall looks good, just a few minor comments.

lumigator/jobs/evaluator/evaluator.py
aittalam: Good catch, I should have added a docstring to explain my rationale here!
While predictions and ground_truth are kinda self-explanatory, to me examples has always been a bit vague and I wanted to be explicit here about the fact that these are not the samples we evaluate (which are the predictions), but the original inputs to the llm that created the predictions.
Given this is just a local variable (i.e. nothing changes in the data the field is still going to be called "examples") I think a proper explanation in the docstring might be enough, but please lmk if you think I should stick to the examples variable name for clarity and I'll just use it

lumigator/jobs/evaluator/eval_metrics.py
aittalam: +1! Do you know if we already have something similar in the inference job? In that case I'll use the same wording for consistency, otherwise I'll add something there too

lumigator/jobs/evaluator/eval_metrics.py
aittalam: Good point. I think we can split the problem in two:

- how we can run custom g-eval tasks
- how we can automatically map a task from TaskType

If I got it right, right now you are more interested in the latter, right? I agree with you and I think it'd be generally useful to map a set of metrics to a given task, e.g. in the UI we can expose (and let users choose among) a different subset of metrics depending on which task we are evaluating for.

At the job level, I chose to have a general `_g_eval` method that is parametric wrt the task, but also a named metric. The reason is that, at least for now, I'd like to be able to enumerate different metrics rather than having a general name in the interface and additional parameters to pass, as this can easily explode IMO.

This also links to the former sub-problem, i.e. running custom g-evals. If we decide to allow more input parameters to be set to the evaluation job, we can consider having a `g_eval_custom` task which just gets evaluation steps from parameters. As an alternative (or as a way to get there iteratively), we can start documenting how to add the new metric to the json file and call it in the eval code. WDYT?

lumigator/jobs/evaluator/eval_metrics.py
aittalam: Thank you! I'll add it 🙏

lumigator/jobs/evaluator/eval_metrics.py
HareeshBahuleyan: > If I got it right, right now you are more interested in the latter, right?

Yes, correct. 
Regarding the "different subset of metrics". there is a related ticket for task-specific metrics, so maybe this can be done as a part of that:
https://github.com/mozilla-ai/lumigator/issues/922

If I understand correctly, the "task" the g-eval code actually refers to the keys in the g_eval_prompts.json and not necessarily to the TaskType enum that I linked to. So one can actually have, for example, "summarization_formal", "summarization_informal" . Being able to pass this parameter as user input would be great (probably for the next iterations of g-eval, for now we can go with the hard-coded one)

lumigator/jobs/evaluator/eval_metrics.py
HareeshBahuleyan: Ahh, I am afraid that we don't have it in inference either! Usually I get notified of this error from the ray job logs. Probably it is already a clear error message in those logs, in which case, please feel free to ignore this comment.

HareeshBahuleyan (approved):

lumigator/jobs/evaluator/evaluator.py
HareeshBahuleyan: Sounds good, docstring update would work!

HareeshBahuleyan:
Thanks @aittalam  Pre-approved - all good, maybe only requires some additional error messages and updates to doc-strings.

njbrake (approved):
Wooo! I'm so excited for this! This is going to be a very nice addition to the evaluation features of Lumi. I had a few editorial comments but nothing serious. Thanks for the work on this 💯 🎉

lumigator/jobs/evaluator/evaluator.py
aittalam: Thanks Nathan! We don't have an issue for that yet, if you want to open one I'll name this examples (and add the docstring) for the time being so we'll update everything consistently afterwards

lumigator/jobs/evaluator/eval_metrics.py
aittalam: Ok, I'll note it down to check and open an issue if we see the message is not good enough (also, now we are using deepeval's client which is different from the one we use in inference but if we choose to migrate everything to litellm we can take advantage of its error handling)

lumigator/jobs/evaluator/eval_metrics.py
aittalam: Exactly! I also saw approaches (e.g. [openevals](https://github.com/langchain-ai/openevals)) that implement a zero-GT LLM-as-judge evaluation, so we could have something like supervised vs unsupervised summarization eval metrics that both rely on LAJ.

For translation specifically, we can have a new `translation` key in the json file with prompts which are adequate for the task, and specify a `g_eval_translation` metric that calls g-eval with task==`translation`. Whatever method we use upstream to map a set of metrics with the task will be compatible with this.

PR #1006, opened 2025-02-25
Unify naming to use `max_new_tokens`

njbrake:
# What's changing

This change unifies how we handle the max tokens. We will specify the `max_new_tokens` to control how many new tokens are generated by either Seq2Seq model. That parameter will not control how many tokens we allow as input (existing logic already handles setting the tokenizer max length configuration).

The PR that comes after this one will adjust the logic to handle and set the tokenizer.model_max_length parameter based on the model type (seq2seq/CausalLM) and the max_new_tokens param.

# I already...


lumigator/jobs/inference/inference_config.py
```diff
@@ -167,6 +166,6 @@ class InferenceJobConfig(BaseInferenceJobConfig):
     job: JobConfig
     system_prompt: str | None = Field(title="System Prompt", default=None, exclude=True)
     inference_server: InferenceServerConfig | None = None
-    params: GenerationConfig | None = None
+    generation_config: GenerationConfig | None = None
```
dpoulopoulos, line 169: Is this something that you missed in the previous PR?

lumigator/backend/backend/tests/conftest.py
```diff
@@ -463,8 +463,7 @@ def simple_infer_template():
             "revision": "{revision}",
             "use_fast": "{use_fast}",
             "trust_remote_code": "{trust_remote_code}",
-            "torch_dtype": "{torch_dtype}",
-            "max_new_tokens": 500
+            "torch_dtype": "{torch_dtype}"
```
dpoulopoulos, line 466: Does removing it from this config break anything? Why was `max_new_tokens` here?

lumigator/jobs/inference/inference_config.py
njbrake: yes 🙈 . But it didn't make a difference because the InferenceJobConfig in inference/schema.py did have `generation_config` set already, and since this class inherits that class, this object did end up having the parameter name as expected.

So, the byproduct of me missing this was that the InferenceJobConfig object used in the inference job had both a generation_config and a params member, but the params member was unused.

lumigator/backend/backend/tests/conftest.py
njbrake: It appears to have been an accidental artifact left from https://github.com/mozilla-ai/lumigator/pull/762/files#diff-26ce53c706eccca82798fff14bfc531b57789138468d24dee14552943c640574?

max_new_tokens was added but then didn't appear to be used anywhere in the final version of the code that was merged.

PR #1020, opened 2025-02-25
[BUG] Summarization System prompt was being ignored, a few other parameters are silently ignored

njbrake:
# What's changing

When running a workflow, we were accidentally ignoring the system prompt that the user passes in.
This change:
* Removes the duplicate system_prompt parameter
* Makes the workflow service make use of the provided system prompt
* Disallows for extra config parameters in our pydantic models, which revealed a few other bugs:
    * in several cases, we were passing parameters into Pydantic models that were being ignored.

# I already...


lumigator/backend/backend/tests/conftest.py
```diff
@@ -81,7 +81,7 @@ def wait_for_job(client, job_id: UUID) -> bool:
     for _ in range(1, MAX_POLLS):
         get_job_response = client.get(f"/jobs/{job_id}")
```
njbrake, line 82: Calling get /jobs/{job_id} returns a Job, not a JobResponse.

njbrake:
Sorry for adding you all and then closing the PR: as happened with the last PR, my small bug fix sort of exploded into a bigger unwieldy PR (can anyone say scope creep 😂). I'm going to go back and make these a bit smaller and more manageable PRs to review.

PR #971, opened 2025-02-21
Create experiment and run workflow with task as translation

HareeshBahuleyan:
# What's changing

This is a follow-up of PR #936 
* Previously, the task_definition and validation/checks were happening at the job level.
* In this PR, it is moved to experiment level

ExperimentCreate schema change:
- Replace task with task_definition (to be able to support task specific parameters within a dictionary structure)
- For translation: these would be the source and target language

System Prompt being validated and set within JobDefinitionInference

> If this PR is related to an issue or closes one, please link it here.

Closes https://github.com/mozilla-ai/lumigator/issues/964

# How to test it

Use dataset: https://github.com/mozilla-ai/lumigator/blob/main/lumigator/sample_data/sample_translation_en_de.csv

Run an translation experiment with a causal LLM supported by LiteLLM 

1. make local-up
2. Upload dataset
3. Create Experiment
```
curl -X 'POST' \
  'http://localhost:8000/api/v1/experiments/' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "name": "translation-pr971",
  "description": "pr971",
  "dataset": "4ebc30ad-8abf-477b-9df2-f1a5f4f8eefc",
  "max_samples": -1,
  "task_definition": {
    "task": "translation",
    "source_language": "en",
    "target_language": "de"
  }
}'
```
4. Create Workflow
```
curl -X 'POST' \
  'http://localhost:8000/api/v1/workflows/' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "name": "translation-pr971",
  "description": "pr971",
  "experiment_id": "1",
  "task_definition": {
    "task": "translation",
    "source_language": "en",
    "target_language": "de"
  },
  "model": "gpt-4o-mini",
  "provider": "openai",
  "dataset": "4ebc30ad-8abf-477b-9df2-f1a5f4f8eefc"
}'
```
Check the UI to see the completed experiment:
image

# Additional notes for reviewers

Anything you'd like to add to help the reviewer understand the changes you're proposing.

# I already...

- [x] Tested the changes in a working environment to ensure they work as expected
- [x] Added some tests for any new functionality
- [ ] Updated the documentation (both comments in code and [product documentation](https://mozilla-ai.github.io/lumigator) under `/docs`)
- [ ] Checked if a (backend) DB migration step was required and included it if required

PR #1045, opened 2025-02-26
[BUG] System prompt unused

njbrake:
# What's changing

This PR addressed the issue where the system_prompt parameter of the workflow create request was not being used during the inference request.

# I already...

### DIFF
--- a/lumigator/backend/backend/services/jobs.py
+++ b/lumigator/backend/backend/services/jobs.py
@@ -163,7 +163,6 @@ class JobDefinitionInference(JobDefinition):
                 base_url=request.job_config.base_url if request.job_config.base_url else None,
                 model=request.job_config.model,
                 provider=request.job_config.provider,
-                system_prompt=request.job_config.task_definition.system_prompt,
                 max_retries=3,
             )
         job_config.generation_config = GenerationConfig(
--- a/lumigator/backend/backend/services/workflows.py
+++ b/lumigator/backend/backend/services/workflows.py
@@ -11,6 +11,7 @@ from lumigator_schemas.jobs import (
     JobResultObject,
     JobStatus,
 )
+from lumigator_schemas.tasks import SummarizationTaskDefinition, TaskType
 from lumigator_schemas.workflows import (
     WorkflowCreateRequest,
     WorkflowDetailsResponse,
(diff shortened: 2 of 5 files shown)


lumigator/backend/backend/services/jobs.py
```diff
@@ -163,7 +163,6 @@ def generate_config(self, request: JobCreate, record_id: UUID, dataset_path: str
                 base_url=request.job_config.base_url if request.job_config.base_url else None,
                 model=request.job_config.model,
                 provider=request.job_config.provider,
-                system_prompt=request.job_config.task_definition.system_prompt,
```
njbrake, line 166: The system prompt is passed in to the `InferenceJobConfig` up in line 138, so the system_prompt is still accessible tot he job.

HareeshBahuleyan:
Hey @njbrake , I think some of this system_prompt changes will be fixed in https://github.com/mozilla-ai/lumigator/pull/971/files#diff-59885862ae7b5c8dd1f02df9d48befc70d96c68b03e5308d24a45990f0e5bb22

Also, with that PR^ I plan to move system_prompt to be a field under `JobInferenceConfig` (rather than under `TaskDefinition` 

Could you check that PR to see if that is the case? Or if that does not address what you plan to do, please let me know.

lumigator/backend/backend/services/workflows.py
```diff
@@ -63,12 +64,21 @@ async def _run_inference_eval_pipeline(
         """
         # input is WorkflowCreateRequest, we need to split the configs and generate one
         # JobInferenceCreate and one JobEvalCreate
+        if not request.system_prompt:
```
HareeshBahuleyan, line 67: Would this if-else be an issue if I want to create an experiment with translation as the task? In both conditions the task_definition is SummarizationTaskDefinition()

njbrake: ooh! Yes, it would be an issue. However, the current code in main already has this issue because we weren't passing in task_definition so it was already defaulting to Summarization. It's a bug, but I think you may already be addressing this in a PR for translation support?

HareeshBahuleyan: Yeah, valid point - currently translation is supported only at the job level. 
It is being fixed to be at experiment level in #971

HareeshBahuleyan (approved):

PR #1084, opened 2025-02-27
Added the script for OSX and Linux

davidmanzanoai:
# What's changing

The script allows a easy way to set up the local environment for starting lumigator. It takes care of installing all the required software (it it is not present in your environment). It supports:
- OSX, including brew based installed or a direct DMG installation.
- Linux (debian and ubuntu), supporting sudo based installation or rootless (here there are some dependencies the user needs to secure they are installed prior to run the script (related to Docker rootless installation). In any case, they will need to have available in the user environment: curl, zip, make. 

The script, by default, uses the latest stable release (at this moment alpha-0.1.0) unless it is executed with flag "-m" which points to main. If the source code already exists in the selected directory (using -d or otherwise it will start from current directory)  it is possible to force overwritting the folder with a fresh installation using "-o"

Refs #...
Closes #...

# How to test it

Steps to test the changes:

1. sh install_lumigator.sh -m (if you want to try main content)
2.
3.

# I already...


javiermtorres:
LGTM. Note that if `/tmp/local.db` exists, it seems to be used by the backend, but imho it's not an issue of the installer itself and we'd need to check thoroughly to see if this is really the case.

javiermtorres:
> LGTM. Note that if `/tmp/local.db` exists, it seems to be used by the backend, but imho it's not an issue of the installer itself and we'd need to check thoroughly to see if this is really the case.

False alarm. It was another issue in my host.

javiermtorres (approved):

Kostis-S-Z:
@davidmanzanoai I didn't manage to get it fully running, it got stuck here:

image


Some other notes:
- Before merging this PR, the docs should be updated to reflect the new way of installing, both in README and the mkdocs.
- The script shouldnt say:
> This script will install the latest Docker and Docker Compose, then set up Lumigator.

If docker is already installed and the installation will be skipped, cause that might cause confusion to the user if they already have Docker installed. They might think that Lumigator might try to overwrite an existing installation, or install it a different version in parallel which will cause issues. Better phrasing could be:
> This script will try to find an existing Docker & Docker compose installation. If none is found, then it will install the latest Docker and Docker Compose available. Finally, the script will set up Lumigator and start running it.

PR #1090, opened 2025-02-28
Use key from the secrets table in the env var list in Ray jobs

### LINKED ISSUE #930: [FEATURE]: Jobs: update creation to include API keys when required
### Motivation

When we create jobs in the `JobService`, if a job (e.g. inference) requires contact with an external service via API (e.g. OpenAI, Mistral, HuggingFace) we should supply the relevant keys as part of the job submission.

* Supply the relevant key for a job (HuggingFace, Mistral, OpenAI, DeepSeek)
* ~Prefer a value stored in the DB over an environment variable (configured during deployment) - fallback to using env var~
* Must be documented
* Must be tested
 
Depends on: https://github.com/mozilla-ai/lumigator/issues/950
Depends on: https://github.com/mozilla-ai/lumigator/issues/933 (Redaction must be in place to prevent data leaking via `/jobs/{id}` etc.)
Depends on https://github.com/mozilla-ai/lumigator/issues/925

**NOTE:**

The intention in the short term is to adjust the job configuration in Ray so that we do not rely on implicit environment variables being present in the Ray container, instead we will explicitly use the supplied API key.

In future iterations the `JobService` may be adjusted to interact with 3rd party APIs (OpenAI/Mistral) for inference jobs directly, removing the dependency (and additional risk of another component having access to API keys) on Ray. Note, Ray may still require the `HF_TOKEN` equivalent if access to gated models is required.

### Alternatives

_No response_

### Contribution

_No response_

### LINKED ISSUE #933: [FEATURE]: API: Redact sensitive API keys from Ray
### Motivation

When users of the API call `/jobs/{id}` they get an object which combines data from Lumigator's database and also from the upstream Ray service. 

... (also with `/jobs/` which lists all jobs which appear in both Lumigator and Ray).

Lumigator should redact any values for keys that are configured in the system.

* Allow regex patterns for sensitive keys to be described in Lumigator's backend settings
* Allow redaction value to be supplied (with sensible default)
* Redact all API responses which contain data from Ray that includes the sensitive patterns

Depends on https://github.com/mozilla-ai/lumigator/issues/930

### Alternatives

_No response_

### Contribution

_No response_

javiermtorres:
# What's changing

A `JobCreate` request can now include a list of api keys to be retrieved from the secrets table and set as env vars in the Ray job. Missing keys will appear as warnings but will not stop the job from being run.

Closes #930

# How to test it

Please take into account that this test case need to use a real api key for the moment, so it is not included in CI nor any real key is included in this test.

1. Remove the trailing `_` in the test case `backend/tests/integration/api/routes/test_api_workflows.py::_test_launch_job_with_secret
2. Fill in an incorrect key where it says `<WRONG SECRET HERE>` and the right API key where it says `<CORRECT SECRET HERE>`
3. Run integration tests as usual

A unit test isolating the Ray environment will likely be developed, but due to the interface with Ray it is advisable to manually run the integration tests locally if this part is modified.

N/A

# I already...


lumigator/backend/backend/repositories/secrets.py
```diff
@@ -27,6 +28,31 @@ def get_secret_by_name(self, name: str) -> SecretRecord | None:
...
+            return None
+
+        for key, value in secret_data.items():
+            setattr(secret, key, value)
+        self.session.commit()
+        loguru.logger.critical(
+            f"DB _update_secret --> {[(r.name, r._name, r.value, r.description) for r in self.list()]}"
+        )
+
+        return secret
```
peteski22, line 55:
```suggestion
```

This is duplicated from below (without the logging of the secret)

lumigator/backend/backend/services/secrets.py
```diff
@@ -2,6 +2,7 @@
 import binascii
 import os
 
+import loguru
```
peteski22, line 5:
```suggestion
```

Unused

lumigator/backend/backend/services/secrets.py
```diff
@@ -38,6 +39,7 @@ def delete_secret(self, name: str) -> bool:
         :returns: ``True`` if the secret was deleted, ``False`` if the secret was not found, or not deleted.
         """
         return self._secret_repo.delete_secret(name)
+
```
peteski22, line 42:
```suggestion
```

lumigator/backend/backend/repositories/secrets.py
javiermtorres: Ok, removing duplicate sections (merge issues I presume)

javiermtorres:
> It would be cool if we could also add unit tests to isolate the changes in say the job service (not just the integration test).
> 
> I'll mark this as request changes as we're chatting about how this might work offline.

I can add a unit test to check that the job config is filled in with the secret, and another in the job itself to check that it retrieves the secret appropriately, sure.

peteski22 (approved):
❤️

Issue #811, opened 2025-02-05
Use Lumigator to run comparison of DeepSeek R-1 models with ACI-Bench Doctor-Patient summarization


njbrake:
Using Lumigator I would like to evaluate all of the DeepSeek R-1 models (including distilled ones) against the [ACI-Bench](https://github.com/wyim/aci-bench) dataset.

https://huggingface.co/collections/deepseek-ai/deepseek-r1-678e1e131c0169c0bc89728d

These models can be compared against GPT-4 or Claude 3.5 Sonnet V2 as a baseline.

Issue #1083, opened 2025-02-27
[FEATURE]: Dynamic list of mean metric scores


aittalam:
### Motivation

The current inference+eval workflow implementation copies a list of hardcoded metrics from the evaluation results file into a dictionary that will be saved together with all the `RunOutputs`.

We should make this list dynamic (e.g. by traversing the eval results dictionary for all the `*_mean` results) so we can compile the RunOutputs even if not all metrics are present (or if there are new ones).

### Have you searched for similar issues before submitting this one?

- [x] Yes, I have searched for similar issues

Issue #1082, opened 2025-02-27
[FEATURE]: Workflows accept a list of eval metrics as a parameter


aittalam:
### Motivation

The current workflow runs eval without passing any metrics, which makes it take the default list which is currently ["meteor", "rouge", "bertscore", "bleu"]. We should not impose this list to all workflows for the following reasons:

- Now that we are tackling different tasks we will soon have different list of default metrics, not just the one
- We might not want to run some metrics by default (especially if they have strong dependencies, e.g. g-eval needs a judge LLM to be available, possibly an API key too)
- Integration tests run workflows and are currently calculating every metric (and while this might be useful, we could have much shorter tests if we ran only easy-to-calculate metrics)

### Have you searched for similar issues before submitting this one?

- [x] Yes, I have searched for similar issues

PR #1091, opened 2025-02-28
Evaluating different flavors of DeepSeek R1.

njbrake:
# What's changing

These are all the code changes made to support running the evaluation to compare DeepSeek models.

* Support a workflow specifying a generation config
* Support a workflow specifying what metrics to run
* Support for extracting and saving reasoning informatin
* Support for saving an estimate of token lengths for ref and pred
* Jupyter notebook with explanation and results
* Bumped up memory of ray docker container because otherwise the cluster runs out of mem
* Bump up the timeout for ray jobs, because it takes ~25 minutes to run the 40 examples through the bigger models (reasoning models are not renowned for speed)
* Add metrics from inference jobs into the workflow metrics

Refs #811

# How to test it

Steps to test the changes:
1. Ask Nathan for the .env file that has the mapping of models to IP addresses, or contact Dimitris/Davide to get the IP addresses directly
2. Run the notebook! (You also need to deploy Lumigator with a DEEPSEEK_API_KEY)

# I already...


docker-compose.yaml
```diff
@@ -100,7 +100,7 @@ services:
       resources:
         limits:
           cpus: '2'
-          memory: '5g'
+          memory: '10g'
```
njbrake, line 103: When running 5 experiments together (even when inf was hosted outside the ray cluster) took more than 5g ram.

lumigator/backend/backend/services/workflows.py
```diff
@@ -122,12 +132,19 @@ async def _run_inference_eval_pipeline(
         # use the inference job id to recover the dataset record
         dataset_record = self._dataset_service._get_dataset_record_by_job_id(inference_job.id)
 
+        if request.metrics:
+            job_config = JobEvalConfig(
+                metrics=request.metrics,
+            )
+        else:
+            job_config = JobEvalConfig()
```
njbrake, line 140: Eventually this will be managed at the experiment level. However, to keep this PR from exploding even further than it already has, I elected to defer it to a later issue https://github.com/mozilla-ai/lumigator/issues/1105

lumigator/backend/backend/services/workflows.py
```diff
@@ -217,11 +239,11 @@ def create_workflow(self, request: WorkflowCreateRequest) -> WorkflowResponse:
 
         return workflow
 
-    def delete_workflow(self, workflow_id: str) -> WorkflowResponse:
+    def delete_workflow(self, workflow_id: str, force: bool) -> WorkflowResponse:
```
njbrake, line 242: Needed so that I can delete a hanging job https://github.com/mozilla-ai/lumigator/issues/1093

lumigator/schemas/lumigator_schemas/workflows.py
```diff
@@ -30,7 +31,10 @@ class WorkflowCreateRequest(BaseModel):
     system_prompt: str | None = None
     inference_output_field: str = "predictions"
     config_template: str | None = None
-    job_timeout_sec: PositiveInt = 60 * 10
+    generation_config: GenerationConfig = Field(default_factory=GenerationConfig)
+    job_timeout_sec: PositiveInt = 60 * 60
```
njbrake, line 35: 10 minutes when you're doing an inference job on an LLM for even a small/medium dataset is very fast. Bumping this up to a more reasonable number imo

lumigator/backend/backend/services/workflows.py
```diff
@@ -161,19 +190,12 @@ async def _run_inference_eval_pipeline(
...
             outputs = RunOutputs(
-                metrics={
-                    "rouge1_mean": round(eval_output.metrics["rouge"]["rouge1_mean"], 3),
-                    "rouge2_mean": round(eval_output.metrics["rouge"]["rouge2_mean"], 3),
-                    "rougeL_mean": round(eval_output.metrics["rouge"]["rougeL_mean"], 3),
-                    "rougeLsum_mean": round(eval_output.metrics["rouge"]["rougeLsum_mean"], 3),
-                    "bertscore_f1_mean": round(eval_output.metrics["bertscore"]["f1_mean"], 3),
-                    "bertscore_precision_mean": round(eval_output.metrics["bertscore"]["precision_mean"], 3),
-                    "bertscore_recall_mean": round(eval_output.metrics["bertscore"]["recall_mean"], 3),
-                    "meteor_mean": round(eval_output.metrics["meteor"]["meteor_mean"], 3),
```
njbrake, line 174: @khaledosman Is this change going to affect the frontend? If you access information via `workflow.rouge1_mean` it would now be at `workflow.rouge_rouge1_mean` since I now preprocess to flatten the array instead of manually specifying metrics like is happening here.

I see the "Metrics" type in `lumigator/frontend/src/types/Workflow.ts` which looks like it might impact, not sure. I can update that file in this PR if my understanding is correct.

lumigator/jobs/inference/tests/test_model_clients.py
njbrake, line 1: These tests were created with an assist from Github copilot, I'll come back here and consolidate the variables etc.

lumigator/backend/backend/services/workflows.py
```diff
@@ -56,6 +57,19 @@ def __init__(
         # TODO: rely on https://github.com/ray-project/ray/blob/7c2a200ef84f17418666dad43017a82f782596a3/python/ray/dashboard/modules/job/common.py#L53
         self.TERMINAL_STATUS = [JobStatus.FAILED.value, JobStatus.SUCCEEDED.value]
 
+    def _prepare_metrics(self, eval_output: JobResultObject) -> dict:
```
aittalam, line 60: Thanks for adding this! If we decide to converge on this specific solution for making the list of metrics that are brought out of the results file to tracking, please make sure your PR explicitly says it closes [this issue](https://github.com/mozilla-ai/lumigator/issues/1083) (@ividal this is one of the two issues we were talking about today!).

When thinking about the problem in general, I thought about two possible approaches: (1) being 100% explicit, i.e. adding which metric "paths" (=possible nested keys) we want to be brought to tracking directly as a workflow parameter; (2) being 50% explicit :-), i.e. having this stored somewherre in config; (3) being implicit, but general, i.e. something like you are doing. IMO the main advantage of 1 and 2 are that ppl know exactly what is brought over, but at the same time they'll need to update this if they add a new metric. The main advantage of 3 is that this extends to any new metric but the way it works is not explicitly documented, as it's just in code right now.

My 2 cents on this are: for the sake of having this code up and running in lumigator, I think 3 is the best as it adds zero overhead to people using it and lends itself to automatically including new metrics. Perhaps when start getting more contributions we'll see the need of being more explicit and then we can update. WDYT?

lumigator/backend/backend/services/workflows.py
aittalam: Cool! This also addresses https://github.com/mozilla-ai/lumigator/issues/1082 then (please ignore if you have already explicitly stated that this PR closes this issue too...)

lumigator/jobs/evaluator/eval_metrics.py
```diff
@@ -40,6 +40,10 @@ def __init__(self, metrics):
                 "method": functools.partial(self._g_eval, task="summarization"),
                 "requires": [EvaluationFields.GROUND_TRUTH, EvaluationFields.EXAMPLE],
             },
+            "token_length": {
```
aittalam, line 43: <3

lumigator/schemas/lumigator_schemas/jobs.py
```diff
@@ -79,6 +79,19 @@ class JobEvalConfig(BaseModel):
...
+    """Custom and limited configuration for generation.
+    Sort of a subset of HF GenerationConfig
+    https://huggingface.co/docs/transformers/en/main_classes/text_generation#transformers.GenerationConfig
+    """
+
+    model_config = ConfigDict(extra="forbid")
+    max_new_tokens: int = 1024
+    frequency_penalty: float = 0.0
+    temperature: float = 0.5
+    top_p: float = 0.5
```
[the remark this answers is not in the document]
aittalam: Ok! In any case, results will still be non-deterministic so we might want to add something in the documentation abou this. WDYT? Not needed for this PR specifically, in case you agree we can just capture it in an issue IMO

lumigator/schemas/lumigator_schemas/workflows.py
aittalam: Thank you! I had to do the same locally when doing my experiments 🙏 😄

lumigator/jobs/inference/inference.py
```diff
@@ -106,16 +106,38 @@ def run_inference(config: InferenceJobConfig) -> Path:
...
-        artifacts=InferenceJobOutput.model_validate(output),
-    )
+    artifacts = InferenceJobOutput.model_validate(output)
+    if all(p.metrics is not None for p in prediction_results):
+        avg_prompt_tokens = sum([p.metrics.prompt_tokens for p in prediction_results]) / len(prediction_results)
+        avg_total_tokens = sum([p.metrics.total_tokens for p in prediction_results]) / len(prediction_results)
+        avg_completion_tokens = sum([p.metrics.completion_tokens for p in prediction_results]) / len(prediction_results)
+        avg_reasoning_tokens = sum([p.metrics.reasoning_tokens for p in prediction_results]) / len(prediction_results)
+        avg_answer_tokens = sum([p.metrics.answer_tokens for p in prediction_results]) / len(prediction_results)
+        metrics = AverageInferenceMetrics(
```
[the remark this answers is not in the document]
njbrake: Good idea! Created https://github.com/mozilla-ai/lumigator/issues/1122 to track this.

lumigator/schemas/lumigator_schemas/jobs.py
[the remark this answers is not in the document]
njbrake: Good call. Created https://github.com/mozilla-ai/lumigator/issues/1123.

lumigator/backend/backend/services/workflows.py
njbrake: Ok, it's all ok because `lumigator/frontend/src/helpers/transformJobResults.ts` `transformJobResults` is taking the mean from the job, not the workflow, so the frontend doesn't need to be updated to be compatible with this change 👍

Issue #1101, opened 2025-03-03
[FEATURE]: API: Settings - allow deletion of a configured secret


peteski22:
### Motivation

Enable users to delete their encrypted secret data from Lumigator via an API endpoint.

### Have you searched for similar issues before submitting this one?

- [x] Yes, I have searched for similar issues

PR #1108, opened 2025-03-04
Settings: Allow deletion of a user created secret

peteski22:
# What's changing

Adds the ability to delete secrets created via the API.

Closes: https://github.com/mozilla-ai/lumigator/issues/1101

# How to test it

Steps to test the changes:

1. `make local-up`
2. [List secrets](http://localhost:8000/docs#/settings/list_secrets_api_v1_settings_secrets__get) - None
3. [Create a secret](http://localhost:8000/docs#/settings/upload_secret_api_v1_settings_secrets__secret_name__put)
4. [List secrets](http://localhost:8000/docs#/settings/list_secrets_api_v1_settings_secrets__get) - One
5. [Delete the secret](http://localhost:8000/docs#/settings/delete_secret_api_v1_settings_secrets__secret_name__delete)
6. [List secrets](http://localhost:8000/docs#/settings/list_secrets_api_v1_settings_secrets__get) - None

# I already...


lumigator/backend/backend/repositories/secrets.py
```diff
@@ -10,6 +10,18 @@ class SecretRepository(BaseRepository[SecretRecord]):
     def __init__(self, session: Session):
         super().__init__(SecretRecord, session)
 
+    def delete_secret(self, name: str) -> bool:
+        """Delete a secret identified by its name.
+
+        :param name: The name of the secret to be deleted
+        :returns: ``True`` if the secret was deleted, ``False`` if the secret was not found, or not deleted.
```
javiermtorres, line 17: Actually the only possibility is to have a deleted count of 0 that means that the secret was not found in the repo, right????

```suggestion
        :returns: ``True`` if the secret was deleted, ``False`` if the secret was not found.
```

javiermtorres (approved):

lumigator/backend/backend/repositories/secrets.py
```diff
@@ -10,6 +10,18 @@ class SecretRepository(BaseRepository[SecretRecord]):
...
+        """Delete a secret identified by its name.
+
+        :param name: The name of the secret to be deleted
+        :returns: ``True`` if the secret was deleted, ``False`` if the secret was not found.
+        """
+        secret = self.get_secret_by_name(name)
+        if not secret:
+            return False
+
+        return bool(self.delete(secret.id))
```
khaledosman, line 23: nit: is the query to get the secret before deleting it needed?

Might be a better option to just expect the caller to pass the secret id instead of its name, and do a delete operation directly instead of 2 sequential db operations

khaledosman (approved):

PR #1113, opened 2025-03-04
Show translation models in models endpoint

### LINKED ISSUE #967: List suggested translation models in models/ endpoint
Currently we have it only for summarization:

https://github.com/mozilla-ai/lumigator/blob/c290cae4fe59e741ab1116ca2cb3542cc4fb8e1c/lumigator/backend/backend/api/routes/models.py#L38-L42

HareeshBahuleyan:
# What's changing

With the inclusion of translation as a new use case that Lumigator supports, this PR updates the models/ endpoint.

* Previously a GET to the models endpoint required a task_name to be specified. This has been changed to tasks (a list) - so when queried with a list of tasks, it returns the models that support those.
* Updated models.yaml to reflect that LLM API models also currently support translation
* Updated backend and sdk unit tests for models/ endpoint

The UI can make a GET request to filter and display supported models per use-case in the experiment create page.

Closes #967 

# How to test it

Steps to test the changes:

1. make local-up
2. Go to Swagger UI
3. Query models/ endpoint with different values
a) single task - summarization
b) multiple tasks - summarization and translation

Or with CURL:
```
curl -X 'GET' \
  'http://localhost:8000/api/v1/models/?tasks=summarization&tasks=translation' \
  -H 'accept: application/json'
```

Returns:
```
{
  "total": 8,
  "items": [
    {
      "display_name": "facebook/bart-large-cnn",
      "provider": "hf",
      "base_url": null,
      "website_url": "https://huggingface.co/facebook/bart-large-cnn",
      "description": "BART is a large-sized model fine-tuned on the CNN Daily Mail dataset.",
      
...
```

TODO: 
- [x] Update Docs
- [ ] Check for breaking changes on UI

# I already...


HareeshBahuleyan:
@khaledosman - I think this may break the UI (Create Experiment) since the API call changes. If that is the case, would you like to add the FE changes to this PR?

lumigator/backend/backend/api/routes/models.py
```diff
@@ -19,30 +20,59 @@ def _get_supported_tasks(data: dict) -> list[str]:
...
+    return filtered_models
+
+
+@router.get("/")
+def get_suggested_models(
+    tasks: list[str] | None = DEFAULT_TASKS_QUERY,
+) -> ListingResponse[ModelsResponse]:
+    """Get a list of suggested models for the given tasks.
+
+    Usage: GET api/v1/models/?tasks=summarization&tasks=translation
```
khaledosman, line 45: its a bit redundant to have `?tasks=X&tasks=Y`, can't we set it as a comma separated list instead? i.e `tasks=summarization,translation` and we just need to split by `,`

khaledosman:
> @khaledosman - I think this may break the UI (Create Experiment) since the API call changes. If that is the case, would you like to add the FE changes to this PR?

I just pushed the change, it will now send a comma separated list of tasks, currently its only fetching for summarization to keep the old behavior and since the api doesn't support `tasks=X,Y` yet, but expects it to be `tasks=X&tasks=Y` which feels a bit weird

lumigator/backend/backend/api/routes/models.py
HareeshBahuleyan: I can do that, although the input then becomes a raw string which is expected to be comma-separated  (instead of the current list of strings) --> which has to be then processed to retrieve the individual tasks. I don't know if there is a cleaner way of doing this. 
https://github.com/mozilla-ai/lumigator/blob/6102322e6c14532869e0e3d8109e80f95bcb5a14/lumigator/backend/backend/api/routes/models.py#L40-L42

khaledosman: Yes, I think a comma separated string is usually more common for this use-case, its a bit nicer because it makes the url more compact and its easier to pass the params through the api clients.

But I'm also open if prefer the other approach from an api/sdk perspective

HareeshBahuleyan:
@khaledosman A related issue that we can handle in a different PR later: https://github.com/mozilla-ai/lumigator/issues/1127 - not all models may have support for all language pairs. So we may want to query supported models also based on the languages selected.

lumigator/backend/backend/api/routes/models.py
khaledosman: sounds good, i'll go with that approach then on the frontend

khaledosman (approved):

Issue #952, opened 2025-02-20
[Internal feature]: automated CW deployment


ividal:
Internally we are in need to make deployment/self-hosting of some models easier.

There was work put into the cli tool / source repo, but this is outdated. Internal contributors should be able to make a safe deployment of a model on our infrastructure.

Specifically, to support #811 , we'd like to make deploying several of those distilled models easy.

PR #1120, opened 2025-03-05
Deploy a Model with vLLM and Ray Serve

dpoulopoulos:
# What's changing

Introduce the Helm Chart and Python source code to deploy any vLLM-supported model on a Kubernetes cluster, using vLLM and Ray Serve.

Refs #952

# How to test it

Follow the corresponding operations guide, available in the PR.

# I already...

### DIFF
--- a/lumigator/rayserve/serve.py
+++ b/lumigator/rayserve/serve.py
@@ -1,0 +1,151 @@
+import logging
+import os
+
+from fastapi import FastAPI
+from ray import serve
+from starlette.requests import Request
+from starlette.responses import JSONResponse, StreamingResponse
+from vllm.engine.arg_utils import AsyncEngineArgs
+from vllm.engine.async_llm_engine import AsyncLLMEngine
+from vllm.entrypoints.logger import RequestLogger
+from vllm.entrypoints.openai.cli_args import make_arg_parser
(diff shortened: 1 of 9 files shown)


lumigator/rayserve/serve.py
```diff
@@ -0,0 +1,151 @@
+import logging
```
chainlink, line 1: Is this on an import path that will get included in lumigator code?

chainlink (approved):
Looks good, just one thing to check on import path. Also, out of curiosity, how come we don't need to specify GPUs for the ray serve job?

lumigator/rayserve/serve.py
dpoulopoulos: Sorry, I don't get that. This script will get pulled by Ray and executed in the Python environment you specify (e.g., the one we `pip install vllm`) just like any other Python script. It does not have any dependency to Lumigator.

chainlink: We have a bit of a non-standard way of including code in the main package See [here](https://github.com/mozilla-ai/lumigator/blob/main/Dockerfile#L8) but looks like this won't get included there. VLLM is not a portable import, it doesn't work on mac, so we need to be careful where this code is getting included (automatically or otherwise)

PR #1121, opened 2025-03-05
Notebook to Demo Translation Usecase via SDK

HareeshBahuleyan:
# What's changing

Added notebook to demo translation use case via SDK on [WMT24++](https://arxiv.org/abs/2502.12404v1)
The experiment contains example workflows with
1. LLM API model (gpt-4o-mini)
2. Local LLM
3. HF multilingual model
4. HF bilingual model 

# How to test it

Setup the env described in the notebook and then run the notebook cell by cell

# I already...


notebooks/translation_demo_wmt24pp.ipynb
```diff
@@ -0,0 +1,848 @@
...
+       "Helsinki-NLP/opus-mt-en-de             0.585  "
+      ]
+     },
+     "execution_count": 10,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "workflow_results = [client.workflows.get_workflow(workflow_id) for workflow_id in workflow_ids]\n",
```
aittalam, line 469: Just a suggestion here: I forgot to pull qwen2.5 and my ollama returned an error (100% my bad! 🙇 ). Everything else continues without issues (which is great), however when we gather results here I got a row of `null`s for qwen (fair enough) and the following cells returned an error (because there were no results to download for qwen).

If you want to make the code a bit more resilient to this kind of errors, I'd suggest something like the following:

```
workflow_results = [client.workflows.get_workflow(workflow_id) for workflow_id in workflow_ids 
                    if client.workflows.get_workflow(workflow_id).status!='failed']
```
... either here or in the following cell, if you still want to show the null values here to surface there was an issue. WDYT?

aittalam (approved):
Thank you @HareeshBahuleyan, this is great!
I tested the notebook and it worked for me. I particularly appreciate the pointers to documentation (for the python env setup) and on how to install qwen with ollama as a prerequisite for the notebook.

I left a minor comment on how to be more resilient if something breaks, but it is good to see how almost everything worked anyway even if I had forgotten to install qwen before trying your code 😇 

Pre-approved, thanks again!

notebooks/translation_demo_wmt24pp.ipynb
HareeshBahuleyan: Thanks @aittalam 
Good point, we can make that more resilient. 
Updated the cell to run a check every 60seconds to see if the workflows are success or failed and only then proceed to the next cells.

PR #1132, opened 2025-03-06
feat: toggle columns in DatasetViewer table

khaledosman:
# What's changing

Part of https://github.com/mozilla-ai/lumigator/issues/1169

This PR updates our new dynamic table component with the ability to select/deselect columns and have them added/removed from the table accordingly

# How to test it

1. View dataset
2. Toggle columns from the new multiselect field

image

While this might not be very helpful for view dataset at the moment, having this functionality in the new table implementation is good to have as the intention is to start reusing this table elsewhere wherever needed like potentially in experiment results, it will also be helpful when dealing with datasets that have multiple groundtruths/languages

# I already...


javiermtorres:
LGTM. I can select the columns from the field selector and I will only see the appropriate columns.
@HareeshBahuleyan is there a multilingual dataset that we could add to the repo (if it's not already there) to check more than 2 columns? Thanks in advance :slightly_smiling_face:

PR #1133, opened 2025-03-06
Rename provider `hf` => `huggingface`

peteski22:
# What's changing

Our provider names align with those used by LiteLLM, except Hugging Face, where we use `hf` instead of `huggingface`. 

Currently, this doesn't matter because we explicitly check if the provider `== "hf"` when generating the config and handle that case differently. But if we later decide to use LiteLLM for this too, we won't have the correct naming.

This will also tie into the UI (and docs) assisting with saving API key secrets using the format: `{provider}_api_key`.

# I already...


njbrake (changes requested):
"huggingface" is a LiteLLM provider, which is why we named it "hf" instead. Sorry, should have specified that in a comment in the code. 

https://docs.litellm.ai/docs/providers/huggingface

PR #1134, opened 2025-03-06
Refactor HuggingFaceModelClient to modular components

### LINKED ISSUE #1130: Refactor HuggingFaceModelClient to support future use cases
Currently HuggingFaceModelClient works with if-else logic but this can quickly grow as we add more tasks and different checks and conditions.

We can breakdown HuggingFaceModelClient to several subclasses such that each subclass handles a certain model type.

HareeshBahuleyan:
# What's changing

Currently logic for different tasks and different model types (seq2seq or causal LM) all reside in a single class. This PR breaks it down and applies the following changes to enable future use cases:

1. **Refactor `HuggingFaceModelClient` to Factory Class:**
    - Added method `create` which returns an instance of either `HuggingFaceSeq2SeqSummarizationClient` or `HuggingFaceCausalLMClient` based on the model configuration.

2. **Introduction of `HuggingFaceSeq2SeqModelClientMixin`:**
    - Contains common functionality for Seq2Seq models (could be used for seq2seq translation models too!).

3. **Creation of `HuggingFaceSeq2SeqSummarizationClient`:**

4. **Creation of `HuggingFaceCausalLMClient`:**

These changes enhance the modularity and maintainability of the code by separating concerns and encapsulating specific logic into dedicated classes.

Closes #1130 

# How to test it

Steps to test the changes:

1. Upload dataset
2. Test with HF models from the UI - BART or Falconsai

# I already...


lumigator/jobs/inference/model_clients.py
```diff
@@ -120,46 +120,53 @@ def predict(
...
-                trust_remote_code=config.hf_pipeline.trust_remote_code,
-                torch_dtype=config.hf_pipeline.torch_dtype,
-            )
-            # Store this so we can check if the input prompt is too long
-            self.tokenizer = AutoTokenizer.from_pretrained(
-                config.hf_pipeline.model_name_or_path,
-                use_fast=config.hf_pipeline.use_fast,
-                trust_remote_code=config.hf_pipeline.trust_remote_code,
-            )
+class HuggingFaceModelClient:
```
njbrake, line 123: Can we put Factory in the name?

njbrake (approved):
One minor comment about renaming the HuggingFaceModelClient, but other than that looks good. I guess one other question would be: does it make sense to split these into separate files, or do you prefer keeping everything inside the model_clients.py file?

lumigator/jobs/inference/model_clients.py
HareeshBahuleyan: Yup, updated to `HuggingFaceModelClientFactory` now 👍

HareeshBahuleyan: Yeah good point on splitting - it was getting quite big and there is more to come with translation. So, split model_clients.py into 3 files.

PR #1138, opened 2025-03-06
Move logging back into workflows service

### LINKED ISSUE #908: [BUG]: Logs in the UI do not appear anymore when job fails
### Description

When an evaluation workflow fails, logs for that job do not appear in the UI anymore. 
This only happens when a job in the workflow fails (logs are shown otherwise when the job completes).
The job correctly generates logs though, as they are present on the ray backend.

### Reproduction

- Start a new evaluation in the UI, making sure it will fail (e.g. choose a model that's too large for your compute).
- Wait for the job to be marked as FAILED in the UI.
- Click on the job row in results and then in the right pane on the "logs" link
- A new horizontal pane will be opened without any content

### Relevant log output

```shell

```

### Expected behavior

Especially when a job fails, we should be able to access its logs from the UI too.

### System Info

- Lumigator version: https://github.com/mozilla-ai/lumigator/commit/a9cf252b52be56aa3cdaceeec5e4e44250c4bd22

javiermtorres:
# What's changing

The handling of logs is now handled at the workflow service, leveraging on the Ray API and log storage within the lumigator DB. The workflow structure and linking specific jobs to a workflow is still handled by mlflow. Also, a job is linked to a workflow as soon as it is created regardless of its outcome, so its logs can always be retrieved regardless of whether it fails or succeeds.

Closes #908 

# How to test it

Steps to test the changes:

1. Start Lumigator
2. Start an experiment with two jobs, one that will succeed (e.g. one using an HF model) and another that will fail (e.g. one using an external API with no configured key)
3. When one of the job fails, click on it and click on the "Logs" link on the sidebar
4. The logs for the failed job should now be visible again

N/A

# I already...

  - No DB changes needed


peteski22 (approved):

peteski22:
Added `do-not-merge` while @javiermtorres is conducting some final tests on this. 😎

@javiermtorres will remove and merge when he's satisfied.

javiermtorres:
The downgrade from varchar to enum is too complex to be allowed. Only an upgrade will be possible.

PR #1144, opened 2025-03-06
Only run integration tests on lumigator code changes

chainlink:
# What's changing

Only run integration and unit tests on lumigator code changes (not helm or docs changes)

# How to test it
Tested by 
- Modifying a file in the root dir (Passed filter)
- Modifying a file in the helm chart (did not pass filter)

# I already...


.github/workflows/lumigator_pipeline.yaml
```diff
@@ -68,26 +81,42 @@ jobs:
...
 
+      - name: Check for modified paths
+        uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            test:
+              - '.github/**'
+              - 'lumigator/**'
+              - '*'
```
macaab26, line 92: Please, include also the "scripts" folder (those are used when spinning up the containers in the integration tests)

macaab26 (approved):

PR #1145, opened 2025-03-07
Explicit API key usage in jobs

### LINKED ISSUE #931: [FEATURE]: Jobs: Move from implicit to explicit API key usage in clients
### Motivation

Jobs need to be updated to ensure they supply the correct API key to their client which was passed into the Ray job (**NOT** use environment variables).

See `lumigator/jobs/inference/model_clients.py`:

* [OpenAI](https://github.com/mozilla-ai/lumigator/blob/c2bac98452a1295d71d4667a195f661ce843d139/lumigator/jobs/inference/model_clients.py#L96)
* [DeepSeek](https://github.com/mozilla-ai/lumigator/blob/c2bac98452a1295d71d4667a195f661ce843d139/lumigator/jobs/inference/model_clients.py#L121)
* [Mistral](
https://github.com/mozilla-ai/lumigator/blob/c2bac98452a1295d71d4667a195f661ce843d139/lumigator/jobs/inference/model_clients.py#L152)

Depends on https://github.com/mozilla-ai/lumigator/issues/930

Example of how implicit defaults are used in OpenAI: https://github.com/openai/openai-python/blob/7193688e364bd726594fe369032e813ced1bdfe2/src/openai/_client.py#L113

### Alternatives

_No response_

### Contribution

_No response_

javiermtorres:
# What's changing

Inference and evaluation* jobs now use an explicit API key instead of relying on environment variables provided by the deployment (Docker etc).

*Evaulation jobs previously didn't require API keys, however since the introduction of G-Eval, the OpenAI API key is required so this is manually 'linked' to the API key decrypted and sent over from Lumigator.

Closes #931 

# How to test it

Steps to test the changes:

1.
2.
3.

N/A

# I already...


lumigator/jobs/inference/model_clients/huggingface_clients.py
```diff
@@ -51,6 +51,7 @@ def _initialize_model_and_tokenizer(self, config):
             tokenizer=self.tokenizer,
             revision=config.hf_pipeline.revision,
             device=config.hf_pipeline.device,
+            use_auth_token=config.api_key,
```
peteski22, line 54:
```suggestion
            token=config.api_key,
```

`use_auth_token` appears to be deprecated:

https://github.com/huggingface/transformers/blob/6bc0fbcfa7acb6ac4937e7456a76c2f7975fefec/src/transformers/pipelines/__init__.py#L762

lumigator/jobs/inference/model_clients/huggingface_clients.py
```diff
@@ -161,6 +162,7 @@ def __init__(self, config: InferenceJobConfig):
         # HF pipeline task overwritten to 'text-generation' since these causalLMs are not task-specific models
         pipeline_config = config.hf_pipeline.model_dump()
         pipeline_config["task"] = TaskType.TEXT_GENERATION
+        pipeline_config["use_auth_token"] = config.api_key,
```
peteski22, line 165:
```suggestion
        pipeline_config["token"] = config.api_key,
```

`use_auth_token` appears to be deprecated:

https://github.com/huggingface/transformers/blob/6bc0fbcfa7acb6ac4937e7456a76c2f7975fefec/src/transformers/pipelines/__init__.py#L762

lumigator/jobs/inference/tests/test_huggingface_clients.py
```diff
@@ -46,6 +49,7 @@ def test_initialization(self, mock_pipeline, mock_automodel, mock_tokenizer, moc
         mock_pipeline_instance = MagicMock()
         mock_pipeline_instance.model = mock_model
         mock_pipeline_instance.tokenizer = mock_tokenizer_instance
+        mock_pipeline_instance.use_auth_token=API_KEY_VALUE,
```
peteski22, line 52:
```suggestion
        mock_pipeline_instance.token=API_KEY_VALUE,
```

peteski22 (commented):
Thanks @javiermtorres 👍🏼 

Added suggestions to use `token` for HF pipelines.

❓ Could we update the tests for when the `api_key` is `None` too?

javiermtorres:
> Can you set the temporary env var for evaluation in the `eval_command` or somewhere high up and noticeable?

Sure, but I think that's part of a previous PR?

lumigator/jobs/evaluator/evaluator.py
```diff
@@ -114,6 +115,11 @@ def eval_command(config: str) -> None:
     config_dict = json.loads(config)
     logger.info(f"{config_dict}")
     config_model = EvalJobConfig(**config_dict)
+
+    # NOTE: Temporary solution to avoid API key issues in G-Eval which defaults to calling OpenAI.
+    if "g_eval_summarization" in config_model.evaluation.metrics and (api_key := os.environ.get("api_key")):
+        os.environ["OPENAI_API_KEY"] = api_key
```
njbrake, line 121: Can we log an error or something? Seems like it may be worth having something as an "else" here.

lumigator/jobs/inference/model_clients/huggingface_clients.py
```diff
@@ -13,7 +13,7 @@ class HuggingFaceModelClientFactory:
     """Factory class that creates the appropriate specialized client"""
 
     @staticmethod
-    def create(config: InferenceJobConfig) -> BaseModelClient:
+    def create(config: InferenceJobConfig, api_key: str | None = None) -> BaseModelClient:
```
njbrake, line 16: Is it necessary to have the `api_key: str | None = None` syntax everywhere, or could you have that in one place and then enforce that everyone else gets the api_key as `api_key: str | None`.

aka can we have the default constructor value be specified in a single location only?

njbrake (approved):
I don't see any blocking issues, but made a few comments about suggested streamlining, if they look useful.

lumigator/jobs/evaluator/evaluator.py
peteski22: In the current code `main`, I think the eval stuff would just conk on its own somewhere. 

We can log something at `error` for sure, I'm not super familiar with the job code (even though I've been tinkering recently) so I don't know if you'd just bail out there, or run eval and be able to get the other metrics at least?

lumigator/jobs/inference/model_clients/huggingface_clients.py
peteski22: I get you, I suppose it's just because you don't have to use the static factory method to get the client, so I dunno if there's any harm in the default `None`. 🤔

Issue #976, opened 2025-02-21
Support for HF seq2seq models for translation


HareeshBahuleyan:
Instead of using f"translate {config.source_language} to {config.target_language}:" as default prompt, we should use it as prefix:
https://huggingface.co/docs/transformers/en/tasks/translation

HareeshBahuleyan:
HF Seq2Seq multilingual models for translation vary quite a bit in terms of usage, depending on the architecture. 

For example,
* models based on mBART architecture → Requires src_lang/tgt_lang `pipeline("translation", src_lang="en_XX", tgt_lang="de_DE")`
* models based on mT5 architecture → Work with prefix engineering ["translate English to German: This is a test sentence.”]

Furthermore, language codes used may be different:
- German
- de
- deu
- deu_Latn
- de_DE

Considering these differences in model approaches, we will begin with support for a few model families and extend the supported approaches based on user interest and requirements.

PR #1147, opened 2025-03-07
Support for HF seq2seq multilingual models for translation: mt5-based model architectures with prefix

HareeshBahuleyan:
# What's changing

To support translation with HF seq2seq models, we [begin](https://github.com/mozilla-ai/lumigator/issues/976#issuecomment-2710689718) with models that are based on mt5 architecture, trained to support many-to-many translation.
These models work based on "prefix" to the input sentence.

The main changes are:
* Addition of a new HuggingFacePrefixTranslationClient class for handling translation tasks using specific model prefixes
With this PR, we would support two variants, namely `google/flan-t5-*` and `bigscience/mt0-*`

* Introduction of `langcodes` library to handle different ways of mentioning the language - as a code "en" vs. as the name "English"

Support for other model families like mbart-* and mmt100-* will come in later PRs.

Refs #976 

# How to test it

Steps to test the changes:

Currently, support is only through the API/SDK and not via the UI yet.

1. Upload en to de sample dataset
2. Create an experiment:
```
curl -X 'POST' \
  -d '{
  "name": "translation",
  "dataset": "82096978-17a6-40e7-8698-67fc7338320c",
  "max_samples": 2,
  "task_definition": {
    "task": "translation",
    "source_language": "English",
    "target_language": "German"
  }
}'
```

3. Create and run workflow:
```
curl -X 'POST' \
  -d '{
  "name": "tranz",
  "experiment_id": "24",
  "task_definition": {
    "task": "translation",
    "source_language": "English",
    "target_language": "German"
  },
  "model": "google/flan-t5-base",
  "provider": "hf",
  "dataset": "82096978-17a6-40e7-8698-67fc7338320c",
  "max_samples": 2
}'
```

4. Check UI to see if the generation is as expected.

Currently, we don't have a way to check if the model specified supports the language that the user has provided. 

Some mt5 based models are only pre-trained variants and do not support zero-shot translation out of the box. Users are required to fine-tune these models with their necessary language pairs [[1](https://stackoverflow.com/questions/76040850/can-mt5-model-on-huggingface-be-used-for-machine-translation)].
  

TODO:
- [ ] Add tests

# I already...


lumigator/backend/backend/services/jobs.py
```diff
@@ -408,6 +408,8 @@ def create_job(
         runtime_env = {
             "pip": job_settings.pip_reqs,
             "working_dir": job_settings.work_dir,
+            # FIXME move to a constant
```
njbrake, line 411: If it's not being fixed in this PR, is it worth creating a github issue and linking it here?

lumigator/jobs/inference/definition.py
```diff
@@ -30,13 +28,14 @@ def generate_config(self, request: JobCreate, record_id: UUID, dataset_path: str
...
             ),
+            task_definition=request.job_config.task_definition,
             system_prompt=request.job_config.system_prompt,
         )
         if request.job_config.provider == "hf":
             # Custom logic: if provider is hf, we run the hf model inside the ray job
             job_config.hf_pipeline = HuggingFacePipelineConfig(
                 model_name_or_path=request.job_config.model,
-                task=request.job_config.task_definition.task,
+                task=request.job_config.task_definition.task.value,
```
njbrake, line 38: Why .value here?

lumigator/jobs/inference/model_clients/huggingface_clients.py
```diff
@@ -23,6 +27,20 @@ def create(config: InferenceJobConfig) -> BaseModelClient:
             logger.info(f"Running inference with HuggingFaceSeq2SeqSummarizationClient for {model_name}")
             return HuggingFaceSeq2SeqSummarizationClient(config)
 
+        elif task == TaskType.TRANSLATION and model_config.is_encoder_decoder:
```
njbrake, line 30: This may be completely irrelevant but just in case it is in any way useful, HF has an example implementation of loading translation models which may be useful.

https://github.com/huggingface/transformers/tree/main/examples/pytorch/translation

This is likely how many companies will be fine-tuning their own translation models, so it may give some insight into how best to load and manage them, not sure.

lumigator/jobs/inference/model_clients/huggingface_clients.py
```diff
@@ -82,7 +101,9 @@ def _set_seq2seq_max_length(self):
         if len(matched_params):
             max_pos_emb = matched_params[0]
         else:
-            raise ValueError(
+            # If none of the plausible fields are found (e.g. google/mt5 model family), use a default value
+            max_pos_emb = DEFAULT_MAX_POSITION_EMBEDDINGS
```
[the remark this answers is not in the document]
njbrake: Side-note don't think we need to support mt5-base/small/large checkpoints themselves since those models aren't fine-tuned. But the error might propogate to models fine-tuned based off those models so it's a valid issue.

https://huggingface.co/google/mt5-base 
"Therefore, this model has to be fine-tuned before it is useable on a downstream task."

Instead of setting a default, could we throw an error or log a warning saying "this may not work"?

I don't think setting a default like this is a stable solution, because then it assumes that the model being used supports whatever that DEFAULT_MAX is. I think it might better to throw an error saying something "the model your using doesn't have a max position embedding set, so you need to tell us what to do".

lumigator/jobs/inference/model_clients/translation_models.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - google/flan-t5-large
+  - google/flan-t5-xl
+  - google/flan-t5-xxl
+
+  # BigScience mT0 models (multilingual T5 variants)
+  - bigscience/mt0-small
+  - bigscience/mt0-base
+  - bigscience/mt0-large
+  - bigscience/mt0-xl
+  - bigscience/mt0-xxl
```
njbrake, line 17: What if the custom model doesn't have these names? I'm not sure we can safely create a definitive list here, we may need to rely on the user to say whether or not the model needs a prefix, though we can maybe log that maybe it does if it's this list of models.

lumigator/jobs/inference/schemas.py
```diff
@@ -50,7 +43,7 @@ class GenerationConfig(BaseModel):
 
 class HuggingFacePipelineConfig(BaseModel, arbitrary_types_allowed=True):
     model_name_or_path: str
-    task: TaskType
+    task: str
```
njbrake, line 46: Can this go back to being TaskType?

lumigator/jobs/inference/model_clients/translation_models.yaml
HareeshBahuleyan: Your observation is on point - there are different families of models and they have different approaches to how they work, I also noted this https://github.com/mozilla-ai/lumigator/issues/976#issuecomment-2710689718 

Supporting "any" seq2seq model for translation out-of-the-box will not be feasible. 

With this first PR, my idea was to support a limited few models that work based on "prefix" (mt5 based architectures). This will be followed by [m2m100](https://huggingface.co/facebook/m2m100_1.2B) model family - which work with params src_lang and tgt_lang instead of prefix - with a different model client class ... and then we keep expanding depending on user needs.

Hence, the hard-coded list of supported models, the list will expand, but unlikely to support each and every case.
Open to suggestions if you feel we should tackle it differently.

lumigator/jobs/inference/schemas.py
HareeshBahuleyan: TaskType is an Enum and this throws a warning when I change the string from "translation" to "translation_en_to_de" (needed for some model families). 
If you feel that we need to stick to TaskType, I can think about a work around whereby the task name is modified only [just before](https://github.com/mozilla-ai/lumigator/blob/5df2915187efdfb41467c6faf6110a65f69f2619/lumigator/jobs/inference/model_clients/huggingface_clients.py#L68) it is fed to the HFpipeline

lumigator/jobs/inference/model_clients/huggingface_clients.py
HareeshBahuleyan: Thanks for the link! Yeah quite a few differences in how different model families input src and tgt language

lumigator/jobs/inference/model_clients/huggingface_clients.py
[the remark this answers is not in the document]
HareeshBahuleyan: Yes, that is right - mt5 is only pre-trained. 

>But the error might propogate to models fine-tuned based off those models so it's a valid issue.

Correct, most fine-tuned variants of mt5 also face this problem. And therefore, without this default-setting we won't be able to a number of support prefix-based-mt5 models. 
In the current supported model list, I have flan-t5 and bigscience/mt0-* models. The latter fails to run due to the missing field. I will update the warning message saying that we are use the default so that it is transparent in the logs. Let me know if that works.

HareeshBahuleyan:
Hey @njbrake thanks for the prompt review and feedback. I have replied to them, happy to discuss if you think further changes are needed.

lumigator/jobs/inference/model_clients/translation_models.yaml
njbrake: Sounds good! my only thought would be if we have an optional input parameter like "prefix" that the client has to pass in, instead of us trying to intelligently set the prefix inside the job.

lumigator/jobs/inference/schemas.py
njbrake: My only reservation would be that it might add some confusion since the name "task" might refer to TaskType or a string, idk if it makes sense to rename the variable or not.

HareeshBahuleyan: Okay sure, will keep it as TaskType.

lumigator/jobs/inference/model_clients/translation_models.yaml
HareeshBahuleyan: I think there are pros and cons. Pro is that it gives flexibility to the user. But then it may add friction - since the user may have to research on their own whether to use "translate from en to de" or "translate from English to German", etc. 
For the moment, both of the supported models `google/flan-t5-*` and `bigscience/mt0-*` are based on the same prefix set inside the job. I would say, later when the need arises for different prefix options, we can address it via user input. 
What do you think?

njbrake: Sounds good to me! Totally agree, for right now this should work fine, if someone reports an issue later we can adjust the approach.

lumigator/jobs/inference/model_clients/huggingface_clients.py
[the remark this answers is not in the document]
njbrake: Sounds good. These days most models support >512 sequence length so it does seem like it probably won't be a problem in practice, unless someone is evaluating an older model.

lumigator/backend/backend/services/jobs.py
HareeshBahuleyan: Sure, done: https://github.com/mozilla-ai/lumigator/issues/1170

lumigator/jobs/inference/schemas.py
HareeshBahuleyan: Done!

njbrake (approved):
LGTM. Thanks!

PR #1165, opened 2025-03-10
Move back to setting the API key in the runtime env

peteski22:
# What's changing

In a previous [PR](https://github.com/mozilla-ai/lumigator/pull/1090) we wired up passing the API key to jobs, however it seems that supplying this via config opens us up to exposing the values in even more places than anticipated (rather than just to folks that log into the Ray dashboard).

It turns up via:

* Our API for Ray logs (as our current inference and eval jobs log the config)
* `results.json` as params
* MLFlow mentions it when meddling with results
* who knows where else

This change moves to using the runtime env vars supplied to Ray to hold the API key. While it is still visible in the Ray dashboard, it limits the scope of where we see it (unless someone decides to log the env vars in their job, in which case it will appear in the Ray logs - which are difficult to parse on the fly to redact, as they're just one single string for the entire log).

Once the PR is merged, I will contribute to: https://github.com/mozilla-ai/lumigator/pull/1145 to update how things are working

Apologies to @javiermtorres who was intending to follow this approach initially before I insisted on using the config. 😳 

Refs https://github.com/mozilla-ai/lumigator/issues/930

# I already...

### DIFF
--- a/lumigator/backend/backend/services/jobs.py
+++ b/lumigator/backend/backend/services/jobs.py
@@ -375,12 +375,16 @@ class JobService:
         dataset_s3_path = self._dataset_service.get_dataset_s3_path(request.dataset)
         job_config = job_settings.generate_config(request, record.id, dataset_s3_path, self.storage_path)
 
+        # Build runtime ENV for workers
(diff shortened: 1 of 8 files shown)


agpituk (approved):
looks good to me, once tests are green happy with it

Issue #1169, opened 2025-03-11
Implement a reusable table component


khaledosman:
We have many different places in the app where we display a table: View dataset, datasets table, ground truth jobs, experiments, workflow results and experiment results.. Right now each of these is a different component with a custom implementation.

Ideally we should implement a generic table component that can be reused for different usecases to avoid code duplication, style inconsistencies and pain when maintaining/changing styles or anything else in any of the implementations.

That way any new feature we want to add to the table implementation can be made available for all of the usecases. The datasets/experiments tables maybe a special case because they're not showing data directly but have custom implementations with the "..." options and custom badge for "status" field, but even then if we can reuse it for workflow results, experiment results and view dataset that's already a win.

Issue #851, opened 2025-02-12
Refactor component architecture


khaledosman:
### Motivation

Alot of our components aren't really properly encapsulated or reusable and have too much state or duplicated logic

Goals:
1. Split bigger components into smaller ones
1. Avoid over reliance on store composables that are shared everywhere making all components unnecessarily stateful
1. Make sure each component is truly encapsulated & reusable by taking its inputs as props and emitting back events to be handled by parent components that are connected to store or apis
1. Make sure all `<style>` tags are properly `<scoped>` to avoid bleeding into other components or libraries in the application
1. Move duplicate code into reusable components and/or composables

### Have you searched for similar issues before submitting this one?

- [x] Yes, I have searched for similar issues

PR #1172, opened 2025-03-11
refactor: migrate results component to use the new table

khaledosman:
# What's changing

Closes https://github.com/mozilla-ai/lumigator/issues/1169
Part of https://github.com/mozilla-ai/lumigator/issues/851

Should be merged after https://github.com/mozilla-ai/lumigator/pull/1132 & https://github.com/mozilla-ai/lumigator/pull/1140

Gets rid of custom component implementations for showing results `LWorkflowResults` & `LExperimentResults` and reuse the new reusable table component instead to render both

image

image

# How to test it

Steps to test the changes:

1. Run an experiment
2. Click on an experiment -> View results
3. Expand an experiment from the table, Click on a workflow -> View results

# Additional Notes to reviewers

- Tooltips are missing (do we care?)
- ~Having too many columns in the multiselect field along with design changes are handled in this PR https://github.com/mozilla-ai/lumigator/pull/1132~ should now be sync'd
- Right now filtering/searching is only enabled for the top level table, the nested table has its own controls which are hidden to avoid bloating the ui which would look something like this:
image
Although I can see some benefit in having it with datasets with a large number of samples

# I already...


lumigator/frontend/src/helpers/getExperimentResults.ts
```diff
@@ -1,25 +1,114 @@
...
+    bleu: data.metrics.bleu.bleu_mean.toFixed(2),
+    ...(model &&
+      model.info && {
+        'model size': model.info.model_size.replace(/(\d+(?:\.\d+)?)([a-zA-Z]+)/g, '$1 $2'),
+        parameters: model.info.parameter_count.replace(/(\d+(?:\.\d+)?)([a-zA-Z]+)/g, '$1 $2'),
+      }),
+    // runTime: undefined, //getJobRuntime(results.id),
+    subRows: transformWorkflowResults(data),
+  }
+}
```
khaledosman, line 76: In the old implementation `results.artifacts.model` used to have the actual model object, now its just the model name, I adjusted the implementation to find the model via the other `/models` endpoint to get the model info to be able to display the size & params.

I also noticed that some models like gpt-4o & gpt-4o-mini have `model.info` set as null so we can't display the model size or params for these, just wanted to double check with you if thats intentional or if any adjustments are needed. cc @javiermtorres @peteski22 @HareeshBahuleyan

peteski22: The `info` property currently can only represent the following:

```yaml
info:
    parameter_count:
    tensor_type:
    model_size:
```

If the data isn't stored in Lumigator (`lumigator/backend/backend/models.yaml`) then it won't be available via the API, and you'll currently see `null` in the responses.

If you're asking: can we omit the `info` property entirely if it would be `null`? Sure!

We could also default the `tasks` to show the property name and an empty object `{}` vs. `null` too if that makes the code more friendly?

It sometimes looks like this:

```yaml
  tasks:
    - summarization:
        max_length: 200
        min_length: 30
        length_penalty: 2.0
        early_stopping: true
        no_repeat_ngram_size: 3
        num_beams: 4
```

And sometimes like this:

```yaml
  tasks:
    - summarization:
    - translation:
```

What do you think?

khaledosman: > If you're asking: can we omit the info property entirely if it would be null? Sure!

not necessarily, I think the structure doesn't matter much, its ok to have it `null` as is, I was just wondering if in these cases whether it makes sense to render the `model size` and `parameters` columns in the results table.

With the current implementation on the frontend it would depend on whether the first row would have them or not, which may not be very accurate since if you choose other models within the experiment that have them we might want to display them 🤔

hasangzl (approved):
LGTM

Issue #1087, opened 2025-02-28
[FEATURE]: /health endpoint returns detailed status of lumigator services


aittalam:
### Motivation

Currently /health returns "ok" by default 😅 , but it'd be great if we could return the status of the different services behind lumigator. This would e.g. allow the front-end to start regardless of all the containers being ready and allow different functionalities depending on the status of the services.

A practical example is the new model preloading feature: it currently blocks the other services from starting until the models are downloaded. If, instead, we start the UI and limit the available functionalities to those that don't require running models until they have been cached, users would already be able to e.g. upload datasets (or see past experiments even if ray is not up yet).

### Have you searched for similar issues before submitting this one?

- [x] Yes, I have searched for similar issues

PR #1173, opened 2025-03-11
Improve Ray startup time

### LINKED ISSUE #1129: [BUG]: Ray takes more than two minutes to start after `make local-up`
### Description

I recently noticed that ray was taking more time to start (e.eg. when the rest of the system is already up, one can e.g. upload a dataset but not run jobs because Ray is not reachable yet).
I added some `date` commands between different the different entrypoint lines and found that the `chmod` operation on `/tmp/ray_pip_cache` takes more than two minutes on my laptop (using a lot of computation too, which makes me think this could be even worse on other hardware).

### Reproduction

- Run `make local-up`
- Wait for ray to be up
- Check for CPU usage in the meantime (you can also run `top` from within the ray container, you'll see `chmod` running for a while)

### Relevant log output

```shell

```

### Expected behavior

A boot that runs chmod only if necessary, only on the strictly required directories.

Why is chmod needed, if the cache volumes are already mounted with `ray` as an owner and all processes run under the `ray` user? Moreover, why it is needed at every single run, even when the permissions are already set?

My suggestion is to (1) verify if and when `rwx` permissions are required (for instance, I re-set the original permissions on my laptop's `.cache/huggingface` dir and model download works anyway), and (2) if there is any need to chmod any of the dirs (especially the pip-cache, which is what is taking most of the time), avoid to do that recursively and at every execution.

### System Info

- OS: Mac OS
- Lumigator version: https://github.com/mozilla-ai/lumigator/commit/6c56e5fae0e94ffa6b019910cb739f740070cb0d

javiermtorres:
# What's changing

If Ray is started in "local mode", where the huggingface cache is shared with the host, permissions are changed unnecessarily causing a large amount of time to be traversed in that operation. This is now avoided. Likewise, the pip cache is now changed ownership instead of permissions, for a faster startup.

Closes #1129 

# How to test it

Start lumigator using `make local-up` and check that Ray takes seconds, and not minutes, to start.

N/A

# I already...


cache/download_models.py
```diff
@@ -14,6 +14,8 @@
 models = [model.strip() for model in models_env.split(",")]
 model_total = len(models)
 
+os.umask(0o000)
```
aittalam, line 17: Rather than defaulting to rwx, I'd use 022 here and make sure the user is the appropriate one

docker-compose.yaml
```diff
@@ -83,12 +83,17 @@ services:
       - -c
       - |
           set -eaux
-          mkdir -p /home/ray/.cache/ && mkdir -p /tmp/ray_pip_cache
-          sudo chmod -R 777 /home/ray/.cache/ && sudo chmod -R 777 /tmp/ray_pip_cache/ || true
-          RAY_JOB_ALLOW_DRIVER_ON_WORKER_NODES=1 RAY_REDIS_ADDRESS=redis:6379 ray start --head --dashboard-port=${RAY_DASHBOARD_PORT} --port=6379 --dashboard-host=0.0.0.0 --ray-client-server-port 10001
+          mkdir -p /home/ray/.cache/huggingface && mkdir -p /tmp/ray_pip_cache
```
aittalam, line 86: isn't this redundant? when you mount the directories are already present in the system

docker-compose.yaml
```diff
@@ -83,12 +83,17 @@ services:
...
+          ls -la /home/ray/.cache/huggingface
```
aittalam, line 87: remember to remove this before merge ;-)

docker-compose.yaml
```diff
@@ -83,12 +83,17 @@ services:
...
           # If the file was mounted in a volume instead of
-          # a shared dir, permissions need to be setup
+          # a shared dir, ownership needs to be changed
           # ... || true allows this to fail (-e is set)
+          sudo chmod 777 /home/ray/.cache/huggingface/hub || true
+          ls -la /home/ray/.cache/huggingface
+          ls -la /tmp/ray_pip_cache
+          sudo chmod 777 /tmp/ray_pip_cache || true
+          ls -la /tmp/ray_pip_cache
```
aittalam, line 95:
```suggestion
```
(see above)

docker-compose.yaml
```diff
@@ -83,12 +83,17 @@ services:
...
-          mkdir -p /home/ray/.cache/ && mkdir -p /tmp/ray_pip_cache
-          sudo chmod -R 777 /home/ray/.cache/ && sudo chmod -R 777 /tmp/ray_pip_cache/ || true
-          RAY_JOB_ALLOW_DRIVER_ON_WORKER_NODES=1 RAY_REDIS_ADDRESS=redis:6379 ray start --head --dashboard-port=${RAY_DASHBOARD_PORT} --port=6379 --dashboard-host=0.0.0.0 --ray-client-server-port 10001
+          mkdir -p /home/ray/.cache/huggingface && mkdir -p /tmp/ray_pip_cache
+          ls -la /home/ray/.cache/huggingface
           # If the file was mounted in a volume instead of
-          # a shared dir, permissions need to be setup
+          # a shared dir, ownership needs to be changed
           # ... || true allows this to fail (-e is set)
+          sudo chmod 777 /home/ray/.cache/huggingface/hub || true
```
aittalam, line 91: I would strongly suggest to use chown here:
- `chmod` changes permissions in the host environment, without the user being aware of that
- `chown` allows access to the directories, without changing anything in the host environment (both if a directory exist and if a new one is created, the owner of the dir is the user that ran the docker application, which is a sound assumption)

docker-compose.yaml
```diff
@@ -83,12 +83,17 @@ services:
...
+          ls -la /home/ray/.cache/huggingface
+          ls -la /tmp/ray_pip_cache
+          sudo chmod 777 /tmp/ray_pip_cache || true
```
aittalam, line 94: see above for chown vs chmod

cache/download_models.py
javiermtorres: In this case we have a separate container that preloads this data. I'm not sure if we can safely infer the user and pwd used so that it makes sense to all contexts involved (ray, preload, host if shared). This container has no knowledge if the ray user in the ray container, plus it may not know the user id in the host volume. But any idea as to how handle this is definitely welcome.

docker-compose.yaml
javiermtorres: IIRC there was some directory that was not present, either because it was a subdir or because it was share with the host and it had not been created there. But we've changed assumptions a couple of times, so I'll check this again.

docker-compose.yaml
javiermtorres: Wop :-S Thanks!

javiermtorres: I thought about leaving it as a sort of debug for this kind of changes, wdyt????

docker-compose.yaml
javiermtorres: Ok, I'll try it out.

javiermtorres:
@aittalam I followed your recommendations and tests are passing, thanks a lot 😊 Could you check if it works as expected in your env?

agpituk (approved):
just a minor comment, tested it locally and it's way faster now :) thanks

.github/workflows/lumigator_pipeline.yaml
```diff
@@ -233,6 +233,9 @@ jobs:
         if: steps.filter.outputs.test == 'true' || contains(github.ref, 'refs/tags/')
         run: ENABLE_FIRST_TIME_CACHE=false make start-lumigator-build
 
+      - name: Output Ray logs
+        run: docker logs lumigator-ray-1
```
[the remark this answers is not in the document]
javiermtorres: It's for debugging purposes. Usually we download only the jobs logs, not the output of the docker container. It's not a big deal to remove it, or maybe better still, to move to the "if failure()" section of the action.

javiermtorres:
> I opened another PR to ensure the backend only runs/accepts requests once ray is ready to prevent errors, might be worth adjusting the health check interval/timeout there accordingly #1211

We have https://github.com/mozilla-ai/lumigator/issues/1087 pending which should help with this.
I'll check #1211 👍

aittalam:
> @aittalam I followed your recommendations and tests are passing, thanks a lot 😊 Could you check if it works as expected in your env?

Tested and works great, thank you! I tried it both without and with the required directories on my system, it created them with the right (host) user and permissions and then accessed them without issues.

aittalam (approved):
LGTM, thank you @javiermtorres! Feel free to merge after updating with the main branch and the tests complete successfully 🙏

Issue #924, opened 2025-02-19
[EPIC] Improve settings


agpituk:
TL;DR:
Make it easier for developers to install, configure, and manage settings in Lumigator by introducing a settings system that allows version-controlled configurations, API-based updates, and better integration with job creation.

Problem Statement 
Currently, configuring Lumigator requires managing environment variables manually and lacks a clear versioning or update system. This creates friction for developers who need to adjust settings, makes it harder to track changes, and increases the risk of misconfigurations across environments.

By introducing a structured settings function, developers will be able to store configurations in version control, update settings via API, and streamline installations, reducing setup time and errors.

**User Stories**
As a developer, I want a version-controlled configuration file, so that I can track and manage settings changes easily.
As a developer, I want an API endpoint to update settings dynamically, so that I don’t have to restart services for every config change.
As a developer, I want the SDK/UI to support configuration updates, so that I can make changes programmatically.
As a developer, I want job creation to automatically include configuration keys, so that settings are applied correctly in Ray without manual intervention.
As a developer, I want to remove reliance on environment variables in Docker Compose, so that my configurations are more maintainable.

User Goals:
Easily manage settings without manually handling environment variables.
Quickly update configurations via API without redeploying.
Ensure consistency across different environments with version-controlled settings.

Non-Goals:
Redesigning the entire build system.

In Scope:
- Have a configuration file that's  in version control with a build system
- Create an endpoint in our API that allow settings update
- Update SDK/UI to be able to call the API endpoint
- Update job creation to include the keys in the entrypoint in ray
- Remove env vars from docker comopse

PR #1176, opened 2025-03-11
UI: Pass secret key names for each workflow based on the provider it is targetting

peteski22:
# What's changing

Now that jobs and workflows can be given the secret key name to grab from Lumigator and supply to underlying jobs, the UI needs to support sending in the right secret name key based on the provider.

UI has strong opinions on the name of settings saved via the UI following the format: `{provider}_api_key` and this PR continues that.

Refs: https://github.com/mozilla-ai/lumigator/issues/924

One thing we might decide later is to include some copy/content on the site explaining which settings need to be saved before you run the experiments.

# How to test it

Steps to test the changes:

1.
2.
3.

# I already...

### DIFF
--- a/lumigator/frontend/src/sdk/experimentsService.ts
+++ b/lumigator/frontend/src/sdk/experimentsService.ts
@@ -101,6 +101,7 @@ export async function createExperimentWithWorkflows(
         experiment_id: experimentId,
         model: model.model,
         provider: model.provider,
+        secret_key_name: `${model.provider}_api_key`,
         base_url: model.base_url,
       }),
     ),
--- a/lumigator/frontend/src/sdk/experimentsService.ts
+++ b/lumigator/frontend/src/sdk/experimentsService.ts
@@ -101,7 +101,7 @@ export async function createExperimentWithWorkflows(
         experiment_id: experimentId,
         model: model.model,
         provider: model.provider,
-        secret_key_name: `${model.provider}_api_key`,
+...(model.requirements.includes('api-key') && {secret_key_name: `${model.provider}_api_key`}),
         base_url: model.base_url,
       }),
     ),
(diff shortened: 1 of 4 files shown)


lumigator/frontend/src/sdk/experimentsService.ts
```diff
@@ -101,6 +101,7 @@ export async function createExperimentWithWorkflows(
         experiment_id: experimentId,
         model: model.model,
         provider: model.provider,
+        secret_key_name: `${model.provider}_api_key`,
```
khaledosman, line 104: what about models that don't require an api key, would this break something in the backend? maybe we can only add it to the payload only if its an api key one

maybe something like this?

```suggestion
...(model.requirements.includes('api-key') && {secret_key_name: `${model.provider}_api_key`}),
```

khaledosman (approved):

lumigator/frontend/src/sdk/experimentsService.ts
khaledosman: on another note I also think its a bit leaky to expect the sdk clients or frontend to specify which api key to use for each model, can't we just use the same naming convention on the backend so we don't have to pass it through the api? afterall the frontend can't just create new keys, but only one of the 4 ones we currently support.

it sounds like we could have some immutable property or something on each key that tells the backend which model its for, or some relationship between models and the secrets tables

peteski22: Hey, thanks @khaledosman. I like the suggestion to check if the requirements say an API key is needed! 

RE: Being leaky:

Some folks could create secrets which don’t follow our UI (and soon SDK) naming, so we have to be flexible in allowing the caller to say which secret key contains the API key they need to be passed through to the job (even via the workflow).

khaledosman: wouldn't passing a model id when creating the secret do the trick? that way we can infer the model via the secret and vice versa.

PR #1178, opened 2025-03-12
Support for language code based HF translation models like facebook/m2m100* family

HareeshBahuleyan:
# What's changing

Add `HuggingFaceLanguageCodeTranslationClient` to support m2m100-* models that support translation where the language pairs are specified in the pipeline as iso codes. Example:
```
pipeline('translation', 'facebook/m2m100_418M', src_lang='en', tgt_lang="de")
```

Note, previously we added `HuggingFacePrefixTranslationClient` to support models that work with prefix before the input sentence in #1147 

Closes https://github.com/mozilla-ai/lumigator/issues/1192

# How to test it

Steps to test the changes:

1. Upload en to de sample dataset
2. Create an experiment:
```
curl -X 'POST' \
  -d '{
  "name": "translation",
  "dataset": "82096978-17a6-40e7-8698-67fc7338320c",
  "max_samples": 2,
  "task_definition": {
    "task": "translation",
    "source_language": "English",
    "target_language": "German"
  }
}'
```

3. Create and run workflow:
```
curl -X 'POST' \
  -d '{
  "name": "tranz",
  "experiment_id": "24",
  "task_definition": {
    "task": "translation",
    "source_language": "English",
    "target_language": "German"
  },
  "model": "facebook/m2m100_418M",
  "provider": "hf",
  "dataset": "82096978-17a6-40e7-8698-67fc7338320c",
  "max_samples": 2
}'
```

4. Check UI to see if the generation is as expected.
image

Currently, we don't have a way to check if the model specified supports the language that the user has provided. 

TODO:
- [x] Add jobs unit tests 

# I already...


lumigator/jobs/inference/model_clients/huggingface_clients.py
```diff
@@ -36,10 +37,15 @@ def create(config: InferenceJobConfig, api_key: str | None = None) -> BaseModelC
             # Load the supported translation model families configuration
             translation_config = load_translation_config()
 
-            prefix_models = translation_config.get("prefix_translation_models", [])
+            prefix_models = translation_config.get("prefix_based", [])
```
aittalam, line 40: nit: I understand the `get` to make sure there is something returned even if the keys disappear from the input file. WDYT about bringing this logic in the `load_translation_config` method instead?
One advantage is that you would not need for the presence of those keys in here, and at the same time you could issue some kind of warning if the file does not contain the keys you are expecting (or error, if you think the application should break if those keys are not present)

aittalam:
Hi Hareesh, many thanks for this PR! 
Re: "we don't have a way to check if the model specified supports the language that the user has provided", is it something we just need to implement or we do not have a way to get it from HF? What happens if the user provides the wrong pair? Alternatively, is there information we can gather about the models we are supporting that we could encode in a structured way (e.g. in the yaml file holding the translation models) explicitly stating which language pairs are available?

lumigator/jobs/inference/model_clients/huggingface_clients.py
HareeshBahuleyan: Good idea, I moved the logic to load_translation_config() and included issuing a warning 👍

HareeshBahuleyan:
Yes Davide, regarding the language compatibility one way we can achieve this is by calling the hf_api to get the model tags
https://huggingface.co/docs/huggingface_hub/en/package_reference/hf_api#huggingface_hub.HfApi.get_model_tags

And validate if the language is present in the list (not sure if it is a fool-proof way to check for all models, since it depends on the contributors adding those tags): 
image

We could build a functionality for this, at the moment I think words from unsupported languages may be treated as unknown tokens.

aittalam:
> Yes Davide, regarding the language compatibility one way we can achieve this is by calling the hf_api to get the model tags https://huggingface.co/docs/huggingface_hub/en/package_reference/hf_api#huggingface_hub.HfApi.get_model_tags
> 
> And validate if the language is present in the list (not sure if it is a fool-proof way to check for all models, since it depends on the contributors adding those tags)

That's a good point! Still better than nothing though, perhaps instead of having a breaking error we could give a warning to the user that the language is not officially supported so they can decide... The only thing I'd like to avoid is people using unsupported languages from the UI and just seeing bad results with no apparent reason (which btw might happen even if the languages are supported 😅). WDYT?

If that sounds good to you, I think this could be a follow-up so we can keep this PR small and easy. We might even choose to capture this in the UI directly to allow users make a decision before the job is submitted, but I'll let you and @khaledosman decide whether it makes sense and how to implement it (e.g. direct call to HF API vs storing languages info ourselves in the yaml file and making it available via the API).

aittalam (approved):
Thanks @HareeshBahuleyan! Looks good to me, I left a comment re: trapping potentially unsupported languages early but I think it's out of scope wrt this PR. Feel free to merge this when all checks are good and choose how to move forward.

HareeshBahuleyan:
@aittalam Sounds good. Created an issue to track this: https://github.com/mozilla-ai/lumigator/issues/1207

> The only thing I'd like to avoid is people using unsupported languages from the UI and just seeing bad results with no apparent reason (which btw might happen even if the languages are supported 😅).

That is quite true!
For now, we have language selection from the drop-down (just 5 options): 
image
I ll ensure that the HF models that are displayed in the UI support them (according to their model cards on HF)

And I think this problem I think goes beyond HF models - since it also happen with less powerful LLMs (for example local models):
- User may ask to translate to a very-low resource language and the model is unable to do so

And also applies to summarization models. User may upload a summarization dataset in German and try to use it along with BART! (less likely but plausible scenario 😅)

PR #1189, opened 2025-03-13
feat(models): replaced deprecated mistral-7B with ministral -8B

### LINKED ISSUE #569: Update Mistral model version and API
# Issue

We currently use [Mistral API endpoints](https://github.com/mozilla-ai/lumigator/blob/5744518333b2493adc1eda6026f6066ee89e404c/lumigator/python/mzai/backend/backend/api/routes/completions.py#L18) for an inference choice. 

Mistral is in the process of [deprecating several models](https://docs.mistral.ai/getting-started/models/models_overview/#model-deprecation) from API availability. 

> Mistral 7B, Mistral 8x7B, Mistral 8x22B, Mistral Medium, and Mistral Large 1 will be retired from la Plateforme soon.

We [currently make use of](https://github.com/mozilla-ai/lumigator/blob/5744518333b2493adc1eda6026f6066ee89e404c/docs/source/get-started/suggested-models.md?plain=1#L82C16-L82C31) `open-mistral-7b`, which we'll need to replace with `ministral-8b-latest` by March 30. 

# Solution

For the model: 
1.  Do a full replace (https://github.com/search?q=repo%3Amozilla-ai%2Flumigator%20%20open-mistral-7b&type=code)
2. End-to-end testing

For the API: 
1.  Do a diff between our version of the API (https://github.com/mozilla-ai/lumigator/blob/5744518333b2493adc1eda6026f6066ee89e404c/lumigator/python/mzai/backend/pyproject.toml#L12) and the current (1.2.6) https://github.com/mistralai/client-python 
2. Bump version and see if there are significant changes at the completions level.

ividal:
# What's changing

At the end of this month, Mistral is retiring one of the models we offer via API (Mistral-7B). Ministral-8B is its replacement. This PR simply does the switch and updates info + tests.

Closes #569

# How to test it
There are tests targeting this model. In addition, any custom script that was using M-7B should just as easily use Ministral-8B instead.

# I already...

- [ X] Tested the changes in a working environment to ensure they work as expected
- [X ] Added some tests for any new functionality
- [X ] Updated the documentation (both comments in code and [product documentation](https://mozilla-ai.github.io/lumigator) under `/docs`)
- [ NA] Checked if a (backend) DB migration step was required and included it if required

### DIFF
--- a/lumigator/backend/backend/models.yaml
+++ b/lumigator/backend/backend/models.yaml
@@ -78,11 +78,11 @@
     - summarization:
     - translation:
 
-- display_name: open-mistral-7b
-  model: open-mistral-7b
+- display_name: ministral-8b-latest
+  model: ministral-8b-latest
   provider: mistral
   website_url: https://mistral.ai/technology/#models
-  description: Mistral's 7B model, hosted by mistral
+  description: Mistral's 8B model, hosted by mistral
   requirements:
     - api_key
   tasks:
--- a/lumigator/backend/backend/models.yaml
+++ b/lumigator/backend/backend/models.yaml
@@ -78,7 +78,7 @@
     - summarization:
     - translation:
 
-- display_name: ministral-8b-latest
+- display_name: ministral-8b
   model: ministral-8b-latest
   provider: mistral
   website_url: https://mistral.ai/technology/#models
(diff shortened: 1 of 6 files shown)


lumigator/backend/backend/models.yaml
```diff
@@ -78,11 +78,11 @@
     - summarization:
     - translation:
 
-- display_name: open-mistral-7b
-  model: open-mistral-7b
+- display_name: ministral-8b-latest
```
dpoulopoulos, line 81:
```suggestion
- display_name: ministral-8b
```

Should we just go with `ministral-8b` for the display name? For example, with DeepSeek the model is called `deepseek-chat` but the display name is `deepseek-v3`.

dpoulopoulos (approved):
Looks good. I just added a small comment.

PR #1202, opened 2025-03-13
Workflows: improve validation for workflows and jobs

peteski22:
# What's changing

In order to be able to provide more timely feedback to callers of our API validation of workflow and job secrets is carried out earlier and an error returned if there's a clear issue. (We're also double checking the experiment exists before creating the workflow).

Updates to the UI to handle some workflows erroring while some work.

Refs #...
Closes #...

# How to test it

Steps to test the changes:

1. `make local-up`
2. Don't set any settings in the UI
3. Use the UI to upload a dataset and create an experiment targeting a hosted model, and an API model (or 2, or 3)
4. See the errors

Bonus:

Create an experiment with only API models (that don't have settings configured) and the experiment should be deleted too.

# I already...


lumigator/frontend/src/components/experiments/LExperimentForm.vue
```diff
@@ -231,30 +231,50 @@ async function handleRunExperimentClicked() {
     system_prompt: experimentPrompt.value || defaultPrompt.value,
   }
 
-  const workflows = await experimentsService.createExperimentWithWorkflows(
+  const [experimentId, workflowResults] = await experimentsService.createExperimentWithWorkflows(
```
khaledosman, line 234: nit: `workflowResults` can be misinterpreted as the real workflow results

```suggestion
  const [experimentId, createWorkflowResponses] = await experimentsService.createExperimentWithWorkflows(
```

lumigator/frontend/src/components/experiments/LExperimentForm.vue
```diff
@@ -231,30 +231,50 @@ async function handleRunExperimentClicked() {
...
     experimentPayload,
     modelSelection.value.selectedModels,
   )
+  // Separate successful workflows and errors
+  const workflows = workflowResults.filter((r) => r.status === 'fulfilled').map((r) => r.value)
```
khaledosman, line 239: nit: `r.status` can be misinterpreted as the workflow status instead of the promise's

```suggestion
  const successfulResponses = createWorkflowResponses.filter((promise) => promise.status === 'fulfilled').map((promise) => promise.value)
```

lumigator/frontend/src/components/experiments/LExperimentForm.vue
```diff
@@ -231,30 +231,50 @@ async function handleRunExperimentClicked() {
...
+  const failures = workflowResults
+    .filter((r) => r.status === 'rejected')
+    .map((r) => r.reason?.response?.data?.detail)
```
khaledosman, line 242:
```suggestion
  const failures = createWorkflowResponses
    .filter((promise) => promise.status === 'rejected')
    .map((promise) => promise.reason.response?.data?.detail)
```

We could technically use `getAxiosError` to capture all kinds of errors aswell including network/server down errors, not just the ones returned by the api in the body

```suggestion
  const failures = createWorkflowResponses
    .filter((promise) => promise.status === 'rejected')
    .map((promise) => {
      const errorMessage = getAxiosError(r.reason);
      try {
        const json = JSON.parse(errorMessage) // if its a json string, that means it returned `response.data`
        return json.detail
      } catch {
         return errorMessage // otherwise its just an error message / string
      }
    })
```

lumigator/frontend/src/components/experiments/LExperimentForm.vue
```diff
@@ -231,30 +231,50 @@ async function handleRunExperimentClicked() {
...
+  failures.forEach((msg: string) => {
+    toast.add({
+      severity: 'error',
+      summary: `Workflow failed to start`,
+      detail: `${msg}`,
+      messageicon: 'pi pi-exclamation-triangle',
+      group: 'br',
+      life: 6000,
+    } as ToastMessageOptions & { messageicon: string })
+  })
```
khaledosman, line 253: do we want to create multiple toasts for each workflow, or just one with the failure messages combined via a `.join` or similar?

peteski22: I don't mind, but I'll leave that for later if possible as I wanted to make sure 'it works' in the UI after the frontend changes. I guess ideally you don't have loads of failing ones so maybe having one per error makes sense (1 error => 1 action to fix it). Which ever way it goes 👍🏼 🤷🏼 😄

khaledosman (approved):

Issue #1210, opened 2025-03-14
[BUG]: Backend should wait for ray to start before accepting requests


khaledosman:
### Description

When running lumigator using `make local-up` there's a period of time where the backend server has started and is already accepting requests, but ray is not ready yet leading to server errors when opening the frontend.

The backend docker container should wait for / depend on the ray one, so it only starts once ray is ready

<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/672215a4-98a1-421d-94c7-b7c34cb9f321" />

### Reproduction

1. run `make local-up`
2. go to `localhost:80` or `localhost:3000` (if you're running `npm dev` from the frontend folder)
3. the server is throwing errors until ray connection is ready

### Relevant log output

```shell

```

### Expected behavior

Backend docker image should wait until ray is ready

### System Info

Mac OS
latest lumigator version on main

### Have you searched for similar issues before submitting this one?

- [x] Yes, I have searched for similar issues

PR #1211, opened 2025-03-14
fix: wait for ray to start before accepting server requests

khaledosman:
# What's changing

Closes https://github.com/mozilla-ai/lumigator/issues/1210

Adds a health check in docker for ray and ensures the backend waits for ray to be ready before starting to prevent server errors

# How to test it

Steps to test the changes:

1. run `make local-up`
2. try to make requests to the backend before & after this change

# I already...


agpituk:
Thanks for this @khaledosman , I'd wait  until #1173 is merged to retest this. They are independent but I think 1173 will improve a lot how the startup time behaves and I want to ensure we test this one after that one!

agpituk (approved):

PR #1231, opened 2025-03-17
doc: architecture diagram with secrets

### LINKED ISSUE #229: Pin all libraries
Some of the libs we use (e.g. HF evaluate) depend on other libs which are not direct deps for us without pinning them (in the specific case of nltk, evaluate does not even include it as a dep but will just warn you to install it if you try to use e.g. meteor without having it installed). As a result, we cannot guarantee stability (the https://github.com/nltk/nltk/issues/3293 case made this evident).

To fix this, we should test our code to make sure our current set of dependencies is stable and then pin all dependencies, even indirect ones.

### LINKED ISSUE #1229: Update architecture diagrams (system, backend)
Diagram for the backend (API, services, jobs) needs to include secrets.

ividal:
# What's changing

Small update of the diagram to include secrets and give it visibility in the docs.

Closes #1229 

# How to test it

N/A

# I already...

- [NA] Tested the changes in a working environment to ensure they work as expected
- [NA ] Added some tests for any new functionality
- [ X] Updated the documentation (both comments in code and [product documentation](https://mozilla-ai.github.io/lumigator) under `/docs`)
- [NA ] Checked if a (backend) DB migration step was required and included it if required


dpoulopoulos (commented):
Looks good to me. But why does this PR closes #229 (which is closed already)? Is this a typo?

dpoulopoulos (approved):

ividal:
> Looks good to me. But why does this PR closes #229 (which is closed already)? Is this a typo?

It is, sorry! It was #1229 - updated that.

PR #1233, opened 2025-03-18
Remove Alchemy DB env var in Makefile

### LINKED ISSUE #928: [BUG]: Datasets and annotation jobs are not persisted on restart
### Description

When running `make local-up` only the experiments & workflows from previous sessions are persisted, but the datasets and groundtruth jobs go away, it sounds like they're not stored in the docker volume.

### Reproduction

1. run `make local-up`
1. Upload dataset, generate ground truth, run experiments, etc..
1. Close the docker process and run make local-up again

### Relevant log output

```shell

```

### Expected behavior

1. Datasets and annotation jobs from previous docker session should be persisted through a volume
2. The datasets and annotation jobs are returned from the api and shown in the UI

### System Info

- MacOS
- Latest/Main branch

javiermtorres:
# What's changing

The wrong location for the SQL Alchemy env var was being set in the `Makefile`, overriding the defaults present in the `.default.conf` file.

Closes #928 

# How to test it

Steps to test the changes:

1. Start Lumigator
2. Perform some annotation job or an experiment
3. Close Lumigator and start it again
4. Jobs from step 2 should still be visible

N/A

# I already...

  - No DB changes needed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,6 @@ RAY_WORKER_GPUS ?= 0
 RAY_WORKER_GPUS_FRACTION ?= 0.0
 GPU_COMPOSE :=
 MODEL_CACHE_COMPOSE :=
-export SQLALCHEMY_DATABASE_URL ?= sqlite:////tmp/local.db
 
 DEBUGPY_ARGS :=
 ifneq ($(shell echo $(DEBUGPY) | grep -i '^true$$'),)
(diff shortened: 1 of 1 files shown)


javiermtorres:
Yes. Previously this setting was exported to every make target, including the container targets, but it only applies to pytest based tests. Otherwise we'll lose data persistence.

peteski22 (approved):

PR #1248, opened 2025-03-19
Removed cache container and update Readme accordingly

agpituk:
# What's changing

Removing cache-inference container and all of it's setup. It's currently causing more trouble than helping new comers to the repo. 

# How to test it

Steps to test the changes:

1. Run lumigator in any of its mode and it still should work
2. Local cache for models should still work (but not pre-warmed). Even after deleting volumes.
3.

# I already...


.default.conf
```diff
@@ -19,7 +19,6 @@ NVIDIA_VISIBLE_DEVICES=all
 GPU_COUNT=0
 # Cache for HuggingFace models and artifacts
 HF_HOME=${HOME}/.cache/huggingface
-ENABLE_FIRST_TIME_CACHE=false
 MODELS_CACHED="facebook/bart-large-cnn,roberta-large" #ensure values are comma separated
```
peteski22, line 22:
```suggestion
```

peteski22 (approved):
Just added a suggestion for the models list as I don't think it will be used anymore

agpituk:
> Just added a suggestion for the models list as I don't think it will be used anymore

Thanks! I missed this one :)

ividal:
This is sad news - it'll mean going back to a 9-minute warm up time while the model is downloaded on the first run.

Let's reopen the issue on that then. I understand how inconvenient it is to have HF model download fail intermittently, but that warm up time is also painful.

Are we doing automatic retries?

PR #1249, opened 2025-03-19
Support for Opus-MT HF translation models

### LINKED ISSUE #1197: Support for bilingual models from opus-MT family
These models are trained on single language pairs and due to the explicit training, they may be better that general multi-lingual many-to-many models
https://huggingface.co/models?other=marian

HareeshBahuleyan:
# What's changing

Support for Opus-MT models[[1](https://huggingface.co/docs/transformers/en/model_doc/marian)][[2](https://huggingface.co/Helsinki-NLP#models)] via UI and API/SDK.

Closes #1197 

# How to test it

Steps to test the changes:

### Test via UI
1. Upload dataset
2. In the experiments, select use-case as translation, select `Helsinki-NLP/opus-mt`
3. After experiment is complete, ensure that the correct default model for the language pair is used. Example `Helsinki-NLP/opus-mt-en-de` for english to german
image

### Test via API 
User can bring any custom fine-tuned model (_belonging to the same model family_). Example: `Helsinki-NLP/opus-mt-tc-bible-big-mul-deu_eng_fra_por_spa` for a multi-lingual model
1. Upload dataset
2. Create experiment
```
curl -X 'POST' \
  -d '{
  "name": "Opus",
  "dataset": "922a0b14-491a-452e-9a36-98a4e9bb7ba6",
  "max_samples": -1,
  "task_definition": {
    "task": "translation",
    "source_language": "English",
    "target_language": "Spanish"
  }
}'
```
3. Trigger workflow
```
curl -X 'POST' \
  -d '{
  "name": "Opus custom",
  "description": "model",
  "experiment_id": "54",
  "model": "Helsinki-NLP/opus-mt-tc-bible-big-mul-deu_eng_fra_por_spa",
  "provider": "hf",
  "batch_size": 3
}'
```
Check if the specified model was used.

# I already...


lumigator/jobs/inference/model_clients/huggingface_clients.py
```diff
@@ -13,9 +13,8 @@
 
 
 def is_encoder_decoder(model_name: str) -> bool:
-    """Check if the model is an encoder-decoder model"""
-    model_config = AutoConfig.from_pretrained(model_name)
-    return model_config.is_encoder_decoder
+    """Check if the model is an encoder-decoder model."""
+    return model_name.startswith("Helsinki-NLP/opus-mt") or AutoConfig.from_pretrained(model_name).is_encoder_decoder
```
aittalam, line 17: Is this because opus-mt does not identify itself as an encoder-decoder? Would it be useful to open an issue in transformers for this?

lumigator/jobs/inference/model_clients/huggingface_clients.py
```diff
@@ -205,3 +207,81 @@ def predict(self, examples: str | list[list[dict[str, str]]]) -> list[Prediction
...
+        # https://huggingface.co/Helsinki-NLP/opus-mt-tc-bible-big-mul-deu_eng_fra_por_spa#how-to-get-started-with-the-model
+        self.target_language_prefix_string = f">>{self.target_language_alpha3_code}<< "
+
+    @staticmethod
+    def is_model_type_marianmt(model_name_or_path: str) -> bool:
+        """Check if the model is a MarianMT model."""
+        if model_name_or_path.startswith("Helsinki-NLP/opus-mt"):
+            return True
+        config = AutoConfig.from_pretrained(model_name_or_path)
+        return config.model_type == "marian"
```
aittalam, line 240: Same as the encoder-decoder note (nothing required specifically for this PR!)

aittalam (approved):
Thank you @HareeshBahuleyan !
I left a couple of questions but they are not blockers, so please go ahead and merge after the code is in sync with main and all tests succeed

lumigator/jobs/inference/model_clients/huggingface_clients.py
HareeshBahuleyan: Good question. 
Not exactly, so opus-mt models are indeed encoder-decoder models. 

However, it has to appear along with the language pair: `Helsinki-NLP/opus-mt-{src}-{tgt}` 

I added the generic `Helsinki-NLP/opus-mt` as an option to select from the UI and then based on the src and tgt language that the user provides, we load the appropriate model from HF model hub under the hood. (This way we can avoid explicitly enlisting the 500+ models with all possible language pairs on the UI).

lumigator/jobs/inference/model_clients/huggingface_clients.py
HareeshBahuleyan: Same comment as above, since the generic `Helsinki-NLP/opus-mt` model name does not exist, but we use it for simplifying our UI.

lumigator/jobs/inference/model_clients/huggingface_clients.py
aittalam: Ah that makes sense, you can only know if _after_ you include the language pair in the model name, but all this is information you need in advance. Many thanks for the clarification!

PR #1250, opened 2025-03-19
1230 integrate deepevals ollama support into evaluator

### LINKED ISSUE #1230: Integrate deepeval's ollama support into evaluator
As a follow-up to our [investigations](https://github.com/mozilla-ai/lumigator/issues/1013), we decided to provide ollama support, which is already built into deepeval, into evaluator.

This boils down to adding extra parameters to eval so people can specify the model they want to use as a judge, and saving these parameters in a `.deepeval` configuration file as described [here](https://docs.confident-ai.com/docs/metrics-introduction#using-ollama).

aittalam:
# What's changing

- integrate deepeval's ollama support into evaluator
- minor fixes

Closes #1230 

# How to test it

1. pull gemma3:27b on ollama
1. run an evaluation job with the following job_config:

```
                job_config: {
                    "job_type": "evaluator",
                    "metrics": [
                      "meteor",
                      "rouge",
                      "bertscore",
                      "g_eval_summarization"
                    ],
                    "llm_as_judge": {
                        "model_name": "gemma3:27b",
                        "model_base_url": "http://localhost:11434"
                    }
```

Note this might raise issues if one tries to run parallel jobs, some of which relying on GPS while others rely on ollama. We have talked about this in a meeting and decided I'll raise an issue in deepeval's repo

# I already...


lumigator/jobs/evaluator/definition.py
```diff
@@ -1,18 +1,21 @@
...
 from evaluator.schemas import DatasetConfig, EvalJobConfig, EvaluationConfig
+from loguru import logger
 from lumigator_schemas.jobs import JobCreate, JobType
 
 from backend.services.job_interface import JobDefinition
 
 
 class JobDefinitionEvaluation(JobDefinition):
     def generate_config(self, request: JobCreate, record_id: UUID, dataset_path: str, storage_path: str):
+        logger.info(request)
```
HareeshBahuleyan, line 12: Not sure if this line was for debugging purposes 🤔

lumigator/schemas/lumigator_schemas/jobs.py
```diff
@@ -105,11 +105,18 @@ class BaseJobConfig(BaseModel, ABC):
     )
 
 
+class DeepEvalLocalModelConfig(BaseModel):
+    model_name: str
+    model_base_url: str
+    model_api_key: str | None = "ollama"
```
HareeshBahuleyan, line 111: Trying to understand what the `model_api_key` field represents? Does Ollama actually require api keys?

lumigator/jobs/evaluator/schemas.py
```diff
@@ -11,13 +11,15 @@ class DatasetConfig(BaseModel):
     model_config = ConfigDict(extra="forbid")
 
 
-class ModelConfig(BaseModel):
-    path: str
-    model_config = ConfigDict(extra="forbid")
+class DeepEvalLocalModelConfig(BaseModel):
```
HareeshBahuleyan, line 14: I think we can import this directly from 
```
from lumigator_schemas.jobs import DeepEvalLocalModelConfig
```
This way we won't have to duplicate and maintain it at 2 places.

This is possible thanks to being able to specify the ray py_modules to include lumigator_schemas:
https://github.com/mozilla-ai/lumigator/blob/22692469e6f59aa95163ba524ed58a46b4aaaf24/lumigator/backend/backend/services/jobs.py#L448

Currently this is how it is being done in inference schemas for task definition:
https://github.com/mozilla-ai/lumigator/blob/22692469e6f59aa95163ba524ed58a46b4aaaf24/lumigator/jobs/inference/schemas.py#L6-L7

HareeshBahuleyan (commented):
Thanks @aittalam great step towards reducing dependency on closed models 👏 
Added a few comments/clarification questions.

Q: Would this also support vLLM models hosted remotely by changing the base_url? Or only ollama at the moment?

lumigator/schemas/lumigator_schemas/jobs.py
aittalam: So... Ollama _does not_ require one, but deepeval _uses_ it to choose ollama instead of other engines (see [here](https://github.com/confident-ai/deepeval/blob/main/deepeval/metrics/utils.py#L286)).
It is not ideal, but I thought there might be perhaps a way to actually use this for other kind of engines - perhaps sending a PR to deepeval itself - so I decided to leave this as an optional parameter instead of just hardcoding "ollama" in.

lumigator/jobs/evaluator/definition.py
aittalam: Oh yes, forgot it there... my bad!

aittalam:
> Thanks @aittalam great step towards reducing dependency on closed models 👏 Added a few comments/clarification questions.
> 
> Q: Would this also support vLLM models hosted remotely by changing the base_url? Or only ollama at the moment?

In theory [yes](https://docs.confident-ai.com/docs/metrics-introduction#other-local-providers) (this could also be a reason for keeping the API key as an optional parameter)

lumigator/jobs/evaluator/schemas.py
aittalam: Addessed in https://github.com/mozilla-ai/lumigator/pull/1250/commits/5ba34c111cadf6036222c71742ee349e45d05f47

HareeshBahuleyan (approved):
Thanks for the changes, Davide! LGTM

PR #986, opened 2025-02-23
UI: Fix - only show ground truth jobs created by the UI under 'Ground Truth Jobs' tab

peteski22:
# What's changing

Partial/temporary fix to prevent inference jobs associated with experiments from appearing in the Ground Truth Jobs tab.

# How to test it

Steps to test the changes:

1. Create a ground truth job
2. View Ground Truth Jobs panel (1 job)
3. Create an experiment
4. Allow to complete
5. View Ground Truth Jobs panel (still 1 job - previously 2)

# Additional notes for reviewers

Anything you'd like to add to help the reviewer understand the changes you're proposing.

# I already...

- [ ] Tested the changes in a working environment to ensure they work as expected
- [ ] Added some tests for any new functionality
- [ ] Updated the documentation (both comments in code and [product documentation](https://mozilla-ai.github.io/lumigator) under `/docs`)
- [ ] Checked if a (backend) DB migration step was required and included it if required

PR #1253, opened 2025-03-20
Correct the job type when submitting annotation jobs (#986)

dpoulopoulos:
# What's changing

Currently, when running ground truth generation jobs, the job type is immediately overwritten to `inference`.

Identify all ground truth jobs with the job type `annotate` and fix the "Ground Truth Jobs" tab in the UI.

Refs #986

# How to test it

Create an annotation job, either from the UI or the API, and try to retrieve it in the UI, or programmatically by running:

```bash
curl -X 'GET' \
  'http://localhost:8000/api/v1/jobs/?job_types=annotate' \
  -H 'accept: application/json'
```

Note that, currently, ground truth generation is limited to summarization tasks from the UI.

However, an annotation job is a special type of inference job that is used to annotate a dataset with predictions from a model. The predictions are stored in the dataset as a new field called `ground_truth`.

Thus, `JobAnnotateConfig` can inherit from `JobInferenceConfig` but fix the following fields, using the [`SkipJsonSchema`](https://docs.pydantic.dev/latest/concepts/json_schema/#skipjsonschema-annotation) type to prevent them from being included in the JSON schema:
- `job_type` to `annotate`
- `output_field` to `ground_truth`

It also sets sensible defaults for the following fields:
- `store_to_dataset`: `True`
- `model`: `facebook/bart-large-cnn`
- `provider`: `hf`

Users can change the model and provider fields but cannot change the `job_type` or `output_field` fields. So, while the UI only permits someone to run a summarization annotation job, the API can do anything.

# I already...


lumigator/backend/backend/services/jobs.py
```diff
@@ -392,7 +393,10 @@ def create_job(
         # This includes both the command that is going to be executed and its
         # arguments defined in eval_config_args
         try:
-            job_settings = job_settings_map[job_type]
+            if job_type == JobType.ANNOTATION:
```
ividal, line 396: I see the problem you ran into with how job_settings_map is defined. Since it's still useful to have different tasks use the same job to avoid copy-pasting, wouldn't it be preferable to add annotation to job_settings_map [here](https://github.com/mozilla-ai/lumigator/pull/1253/files#diff-59885862ae7b5c8dd1f02df9d48befc70d96c68b03e5308d24a45990f0e5bb22R57)?

Something like:

```
job_settings_map = {
    job_module.definition.JOB_DEFINITION.type: job_module.definition.JOB_DEFINITION for job_module in job_modules
}
job_settings_map[JobType.ANNOTATION] = job_settings_map[JobType.INFERENCE]
# if someone brings something else that can be achieved with an existing job instead of needing a new module, they would add their type here as well.
```

and get rid of this if?

javiermtorres: I definitely agree. An even cleaner alternative would be to allow a definition to register to several `JobType`s instead of a single one, in https://github.com/mozilla-ai/lumigator/blob/989-job-type-annotation/lumigator/jobs/inference/definition.py#L71
```python
JOB_DEFINITION: JobDefinition = JobDefinitionInference(
    command=INFERENCE_COMMAND,
    pip_reqs=INFERENCE_PIP_REQS,
    work_dir=INFERENCE_WORK_DIR,
    config_model=InferenceJobConfig,
    type=[JobType.INFERENCE, JobType.ANNOTATION],
)
```

lumigator/schemas/lumigator_schemas/jobs.py
```diff
@@ -151,10 +151,37 @@ class JobInferenceConfig(BaseJobConfig):
...
+
+    Users can change the model and provider fields but cannot change the
+    job_type or output_field fields.
+
+    Note that, currently, ground truth generation is limited to summarization
+    tasks from the UI. Users can run any ground truth generation task from the
+    API.
+    """
+
+    job_type: SkipJsonSchema[Literal[JobType.ANNOTATION]] = JobType.ANNOTATION
```
javiermtorres, line 180: If we take the approach outlined above by Irina, I'd also avoid modifying the schema, since it will be the job's responsibility to adapt the incoming config to the config used by the job.

An alternative is to use a redefinition of the Inference config that uses optional, literal fields https://docs.pydantic.dev/1.10/usage/types/#literal-type

In this way, the user doesn't specify a value, so it takes the default value; and the user _cannot_ explicitly specify another value than the default one.

The SkipJsonSchema option feels fairly hackish to me. Even if it's not present in the schema, we should verify whether the code may still build the entity from the existing json data.

lumigator/backend/backend/services/jobs.py
javiermtorres: There is one itch that bothers me, which is that we have 3 inconsistent "names" for things which are essentially the same:

* JobType annotation
* JobType inference, task summarization
* JobType inference, task translation

I'd vote for either:

* Making translation a JobType in itself (and mapping all 3 job types into the same inference job)
* Making annotation a task

Of course as part of a separate PR :)

 WDYT?

javiermtorres (approved):
LGTM!

lumigator/schemas/lumigator_schemas/jobs.py
javiermtorres: Let's do it like this for the moment, we can open this discussion in a later PR.

PR #1254, opened 2025-03-20
Job Logs: Fix bug where logs were not updated after first check

### LINKED ISSUE #1088: [BUG] logs endpoint for ground truth generation only returns the first log message
@khaledosman Testing this branch now and noticing that logs in Ray are not parsing fully to the UI.

Both of these ss were taken right after one another:

Screenshot 2025-02-28 at 11 38 54 am
Screenshot 2025-02-28 at 11 39 03 am

_Originally posted by @hasangzl in https://github.com/mozilla-ai/lumigator/issues/1078#issuecomment-2690434852_
            

GT job is finished now but the logs are still not showing fully:
Screenshot 2025-02-28 at 11 44 28 am

peteski22:
# What's changing

This PR fixes a bug where the first call to `jobs/{job_id}/logs` retrieved all current logs for the job, but subsequent calls returned the same value as the first call. 

We now try to get updated logs from Ray when possible, and fallback to returning what we have stored in the DB if Ray returns us errors.

Closes #1088 

# How to test it

Steps to test the changes:

1. `make local-up`
2. Upload a dataset and create an experiment
3. Click on one of the jobs in the experiment, and click logs from the right hand side panel... watch and wait for logs

# I already...

### DIFF
--- a/lumigator/backend/backend/services/jobs.py
+++ b/lumigator/backend/backend/services/jobs.py
@@ -310,17 +310,19 @@ class JobService:
             raise JobNotFoundError(job_id) from None
 
         try:
-            job_logs = self._retrieve_job_logs(job_id)
-        except JobUpstreamError:
+            ray_job_logs = self._retrieve_job_logs(job_id)
+        except JobUpstreamError as e:
             # If we have logs stored, just return them to support 'offline' Ray
             if job.logs:
+                loguru.logger.error("Unable to retrieve job logs from Ray, returning stored DB logs: {}", e)
                 return JobLogsResponse(logs=job.logs)
             raise
 
         # Update the database with the latest logs
-        self._update_job_record(job_id, logs=job_logs.logs)
+        if job.logs != ray_job_logs.logs:
+            self._update_job_record(job_id, logs=ray_job_logs.logs)
 
-        return job_logs
+        return ray_job_logs
 
     def _retrieve_job_logs(self, job_id: UUID) -> JobLogsResponse:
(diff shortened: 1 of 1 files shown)


lumigator/backend/backend/services/jobs.py
```diff
@@ -297,17 +297,32 @@ def get_upstream_job_status(self, job_id: UUID) -> str:
...
+        if not job:
+            raise JobNotFoundError(job_id) from None
+
+        try:
+            job_logs = self._retrieve_job_logs(job_id)
+        except JobUpstreamError:
+            # If we have logs stored, just return them to support 'offline' Ray
+            if job.logs:
+                return JobLogsResponse(logs=job.logs)
+            raise
```
khaledosman, line 318: should we log this error?

lumigator/backend/backend/services/jobs.py
```diff
@@ -297,17 +297,32 @@ def get_upstream_job_status(self, job_id: UUID) -> str:
...
+
+        # Update the database with the latest logs
+        self._update_job_record(job_id, logs=job_logs.logs)
```
khaledosman, line 321: nit: This will do a db request to update the record every time `get_job_logs` gets called which might not be necessary since the logs may have not changed, might be better to check if the logs are different from `job.logs` first to avoid unnecessary requests/load on the db

I don't have enough context, but I'm under the impression that this is a bit sketchy architecture wise because the method name is "get_job_logs" so I would expect it to read only instead of updating the db/job after getting it, not sure if there's a better way to separate the reads from the writes, since you don't necessarily want to write to the db every time someone wants to fetch the logs.

khaledosman (approved):
tested the same code snippet locally, seemed to work fine

PR #1260, opened 2025-03-20
Update schemas, align job results S3 paths, refactor inference metrics calc

peteski22:
# What's changing

* Fixes the schemas: `InferenceJobOutput`, `InferenceMetrics`, `PredictionResult` and `JobOutput` within `jobs/`
* Removes the extra list comprehension in metrics calculation, extracts average calculation to function and adds tests
* Align S3 paths for jobs, try to re-use functions to create URIs and pass in fully qualified storage paths when submitting jobs
* Sanitize S3 filename to prevent forward slashes and other awkward characters
* Inference and Evaluation jobs updated (mostly based on S3 tweaks)

# How to test it

Steps to test the changes:

1.
2.
3.

# I already...

### DIFF
--- a/lumigator/jobs/inference/inference.py
+++ b/lumigator/jobs/inference/inference.py
@@ -1,9 +1,10 @@
 """python job to run batch inference"""
 
 import argparse
-import json
 import os
+import re
 from pathlib import Path
+from uuid import UUID
 
 import s3fs
 from dataset import create_dataloader
@@ -19,6 +20,10 @@ from utils import timer
 
 from schemas import AverageInferenceMetrics, InferenceJobOutput, JobOutput, PredictionResult
 
+SAFE_JOB_NAME_REGEX = re.compile(r"[^\w\-_.]")
+
+JOB_NAME_REPLACEMENT_CHAR = "-"
+
 
 @timer
 def predict(dataloader: DataLoader, model_client: BaseModelClient) -> list[PredictionResult]:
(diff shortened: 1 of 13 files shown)


lumigator/jobs/inference/inference.py
```diff
@@ -120,12 +120,18 @@ def run_inference(config: InferenceJobConfig, api_key: str | None = None) -> Pat
...
-        avg_reasoning_tokens = sum([p.metrics.reasoning_tokens for p in prediction_results]) / len(prediction_results)
-        avg_answer_tokens = sum([p.metrics.answer_tokens for p in prediction_results]) / len(prediction_results)
+        total_results = len(prediction_results)
+
+        avg_prompt_tokens = sum(p.metrics.prompt_tokens for p in prediction_results) / total_results
+        avg_total_tokens = sum(p.metrics.total_tokens for p in prediction_results) / total_results
+        avg_completion_tokens = sum(p.metrics.completion_tokens for p in prediction_results) / total_results
+        # Provide a default for optional fields so we don't affect the average.
+        avg_reasoning_tokens = sum((p.metrics.reasoning_tokens or 0) for p in prediction_results) / total_results
+        avg_answer_tokens = sum((p.metrics.answer_tokens or 0) for p in prediction_results) / total_results
```
njbrake, line 133: Why is the '0' being added? This seems risky since then if 1 of the results has an error or a None for some reason, this could subtly lower the average. Maybe need some way to assert that if 1 result has an answer_token of 0, all answer_token values need to be 0?

peteski22: I may have misunderstood the intention of the code 😞 

The previous code would (I think) raise a `TypeError` if it encountered a `None` in the list comprehension for`p.metrics.reasoning_tokens` as it can't add `int`s and `NoneType`s together.

Is the intention for the code to error out in some way if one of the reasoning tokens isn't present? (if so I will change the code).

If it's not the intention, I think this change might fix a bug, but doesn't alter the average you'd get since it just adds `0` to the sum total that you're dividing for the average.

```python

# 5 elements(where the 5th element was previously None)
input1 = [1, 2, 3, 4]
total1 = sum([x for x in input1])
avg1 = total1 / 5
print(avg1) # 2.0

# 5 elements (with a 0 default)
input2 = [1, 2, 3, 4, 0]
total2 = sum([x for x in input1])
avg2 = total2 / 5
print(avg2) # 2.0
```

Just let me know if I got the wrong idea? 😄

njbrake: The math checks out 😆 . However, yeah I'm struggling to think of a scenario where it would be ok for 4 out of the 5 to have reasoning_tokens set. Why would 1 of those 5 not have reasoning tokens?

Raising an error is what I would want to happen, I think.

peteski22: Would it be better to update the `InferenceMetrics` schema itself to not allow the optional `int`? (same question for answer_tokens).

https://github.com/mozilla-ai/lumigator/blob/50d95ff0e4b29beae15f6f19a1ec7b0653ed7147/lumigator/jobs/inference/schemas.py#L74-L75

njbrake: Hmm, Good question. I guess we need some validator so that we assert that either all are filled out or none are filled out, and disallow a mixture of int and None in the list. Does that seem like a could approach?

peteski22: Adjusted the schemas to disallow 'None' since this is the place we deal with the model. Also added some safety code to check that we have all or none, but not some.

https://github.com/mozilla-ai/lumigator/pull/1260/commits/93c3d0a0afd994da2ddd9604d603f7ada1ad09fd

lumigator/jobs/inference/schemas.py
peteski22, line 1: `InferenceMetrics`:

`reasoning_tokens` is set [here](https://github.com/mozilla-ai/lumigator/blob/64bdaf10049330520eaa614329ae2805d846f44b/lumigator/jobs/inference/model_clients/external_api_clients.py#L89) and will raise an exception if it was `None` (`0` is fine).

`answer_tokens` is set [here](https://github.com/mozilla-ai/lumigator/blob/64bdaf10049330520eaa614329ae2805d846f44b/lumigator/jobs/inference/model_clients/external_api_clients.py#L100) and we'll have an error if `usage` is `None` (the object/dict we try to check `completion_tokens` on, or if `completion_tokens` is `None` as we'd attempt to do `None - 0`.

lumigator/jobs/inference/inference.py
```diff
@@ -112,6 +112,8 @@ def run_inference(config: InferenceJobConfig, api_key: str | None = None) -> Pat
     if config.job.output_field in dataset.column_names:
         logger.warning(f"Overwriting {config.job.output_field}")
 
+    prediction_results: list[PredictionResult]
+    inference_time: float
```
njbrake, line 116: Woah I had no idea you could do this, nice.

lumigator/jobs/inference/inference.py
```diff
@@ -121,16 +123,21 @@ def run_inference(config: InferenceJobConfig, api_key: str | None = None) -> Pat
 
     artifacts = InferenceJobOutput.model_validate(output)
 
+    all_metrics_present = all(p.metrics for p in prediction_results)
+
+    if any(p.metrics for p in prediction_results) != all_metrics_present:
+        raise ValueError("Prediction result 'metrics' must be present in ALL results or NONE, but not in SOME.")
+
+    results = JobOutput(artifacts=artifacts, parameters=config)
+
```
njbrake, line 132: This comment is almost certainly scope creep, but I wonder if the creation of this avg_results and artifact json file should get extracted out to a new class or function..

lumigator/jobs/inference/inference.py
```diff
@@ -139,16 +146,8 @@ def run_inference(config: InferenceJobConfig, api_key: str | None = None) -> Pat
             avg_reasoning_tokens=avg_reasoning_tokens,
             avg_answer_tokens=avg_answer_tokens,
         )
```
njbrake, line 148: What if this Pydantic model has some sort of __init__ logic that creates the average metrics or throws an error if there are any issues? idk

lumigator/jobs/inference/schemas.py
njbrake: Much better design. thank you 🙏

njbrake (approved):
A few scope-creepy type comments but overall looks great. Thank you for the improvements!

lumigator/jobs/inference/inference.py
peteski22: I've pulled out the average calculations

PR #1262, opened 2025-03-20
Testing: Attempting to review wait_for_x code to see if we can see any issues

peteski22:
TESTING - WIP.

Trying to see if this makes any difference to the GHA runners erroring out.

### DIFF
--- a/lumigator/backend/backend/services/jobs.py
+++ b/lumigator/backend/backend/services/jobs.py
@@ -1,6 +1,7 @@
 import asyncio
 import csv
 import json
+import time
 from http import HTTPStatus
 from io import BytesIO, StringIO
 from pathlib import Path
@@ -61,7 +62,7 @@ job_settings_map = {
 
 DEFAULT_SKIP = 0
 DEFAULT_LIMIT = 100
-DEFAULT_POST_INFER_JOB_TIMEOUT_SEC = 10 * 60
+DEFAULT_POST_INFER_JOB_TIMEOUT_SEC = 5 * 60
 JobSpecificRestrictedConfig = type[JobEvalConfig | JobInferenceConfig]
 
 
@@ -135,8 +136,8 @@ class JobService:
             return True
 
         try:
-            status = await self.wait_for_job_complete(job_id, max_wait_time_sec=10)
-        except JobUpstreamError as e:
+            status = await self.wait_for_job_complete(job_id, timeout_seconds=10)
+        except TimeoutError as e:
             loguru.logger.error("Failed to stop job {}: {}", job_id, e)
             return False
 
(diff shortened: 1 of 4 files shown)


lumigator/backend/backend/services/jobs.py
```diff
@@ -381,8 +398,13 @@ async def handle_annotation_job(self, job_id: UUID, request: JobCreate, max_wait
         dataset_filename = self._dataset_service.get_dataset(dataset_id=request.dataset).filename
         dataset_filename = Path(dataset_filename).stem
         dataset_filename = f"{dataset_filename}-annotated.csv"
+        job_status = ""
+
+        try:
+            job_status = await self.wait_for_job_complete(job_id, max_wait_time_sec)
+        except TimeoutError as e:
```
javiermtorres, line 405: If there's a timeout error, I guess we cannot have a value of succeeded after that????

peteski22: 👍🏼 I wasn't sure if we wanted to let the flow carry on to the next conditional check so we also got the same log written saying the results weren't going to be stored in the DB.

lumigator/backend/backend/services/jobs.py
```diff
@@ -61,7 +62,7 @@
 
 DEFAULT_SKIP = 0
 DEFAULT_LIMIT = 100
-DEFAULT_POST_INFER_JOB_TIMEOUT_SEC = 10 * 60
+DEFAULT_POST_INFER_JOB_TIMEOUT_SEC = 5 * 60
```
javiermtorres, line 65: Sometimes it's hard to pin the right time. I'd suggest using the extra param to set the waiting time in tests and avoid modifying the default, but let's wait until we get feedback from the people running their own jobs.

lumigator/backend/backend/services/jobs.py
```diff
@@ -339,34 +340,50 @@ def _retrieve_job_logs(self, job_id: UUID) -> JobLogsResponse:
         except json.JSONDecodeError as e:
             raise JobUpstreamError("ray", f"JSON decode error from {resp.text or ''}") from e
 
-    async def wait_for_job_complete(self, job_id, max_wait_time_sec):
+    async def wait_for_job_complete(
+        self,
+        job_id: UUID,
+        timeout_seconds: int = 300,
```
javiermtorres, line 346: No defaults at this place. Defaults only at top level, please. Also, I'm not sure we need a complex backoff scheme, but I don't have firm arguments against it at the moment :-/

lumigator/backend/backend/services/jobs.py
javiermtorres: Let's give this place a check, after this PR. The except should probably cover a few following lines.

lumigator/backend/backend/tests/integration/api/routes/test_api_workflows.py
```diff
@@ -432,19 +433,49 @@ def test_job_non_existing(local_client: TestClient, dependency_overrides_service
     assert response.json()["detail"] == f"Job with ID {non_existing_id} not found"
 
 
-def wait_for_workflow_complete(local_client: TestClient, workflow_id: UUID):
+def wait_for_workflow_complete(
+    local_client: TestClient,
+    workflow_id: UUID,
+    timeout_seconds: int = 300,
```
javiermtorres, line 439: No defaults at this level, again.

lumigator/jobs/inference/requirements.txt
```diff
@@ -1,12 +1,13 @@
...
+numpy==1.26.3
+pandas==2.2.3
+pydantic==2.10.6
+python-box==7.3.2
+s3fs==2024.2.0
 sentencepiece==0.2.0
-torch==2.5.1
-transformers==4.46.3
+torch==2.6.0
+transformers==4.49.0
```
javiermtorres, line 13: @ividal can the ML team please check these versions just in case?

javiermtorres (commented):
Only comment: please leave defaults at the top level, most probably at the interface with external settings.

lumigator/backend/backend/tests/integration/api/routes/test_api_workflows.py
peteski22: The previous (test) code already had magic numbers further down the method (`300` iterations and `1` second sleeps), I don't think this change actually makes things worse as it makes the values and purpose clear in the method signature.

javiermtorres: Yes, that doesn't mean it was ok :)

PR #1265, opened 2025-03-21
Check type fields in evaluator

### LINKED ISSUE #646: [BUG]: Empty predictions cause the format to be lost
### Description

If an inference job that adds predictions to an existing dataset uses all empty strings for the prediction values, the HF datasets package assumes the format is `float_64`, as shown in the following excerpt of a `dataset_info.json` file:

```json
{
  "builder_name": "csv",
  "dataset_name": "csv",
  "dataset_size": 3640,
  ...,
  "features": {
    "predictions": {
      "dtype": "float64",
      "_type": "Value"
    },
    "examples": {
      "dtype": "string",
      "_type": "Value"
    },
    "ground_truth": {
      "dtype": "string",
      "_type": "Value"
    }
  },
  ...
}
```

A following evaluation job will crash due to finding an incorrect format.

### Reproduction

In the backend integration tests, use the following model when running an inference job previous to an evaluation job:

```python
    infer_payload = {
        "name": "test_run_hugging_face",
        "description": "Test run for Huggingface model",
        "model": "hf://hf-internal-testing/tiny-random-T5ForConditionalGeneration",
        "dataset": ...,
        "max_samples": 10,
        "output_field": "predictions",
        "store_to_dataset": True,
    }
```

The next evaluation job will fail with the following error:

```python
  File "/tmp/ray/session_2025-01-15_23-52-28_884602_1/runtime_resources/working_dir_files/_ray_pkg_cc3766f409bcf1be/eval_metrics.py", line 47, in _meteor
    evals["meteor"].append(ev.compute(predictions=[p], references=[r])["meteor"])
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/ray/session_2025-01-15_23-52-28_884602_1/runtime_resources/pip/31906f6f1883f7fb62d83ad16365f2311e2b5116/virtualenv/lib/python3.11/site-packages/evaluate/module.py", line 455, in compute
    self.add_batch(**inputs)
  File "/tmp/ray/session_2025-01-15_23-52-28_884602_1/runtime_resources/pip/31906f6f1883f7fb62d83ad16365f2311e2b5116/virtualenv/lib/python3.11/site-packages/evaluate/module.py", line 514, in add_batch
    self.selected_feature_format = self._infer_feature_from_batch(batch)
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/ray/session_2025-01-15_23-52-28_884602_1/runtime_resources/pip/31906f6f1883f7fb62d83ad16365f2311e2b5116/virtualenv/lib/python3.11/site-packages/evaluate/module.py", line 596, in _infer_feature_from_batch
    return self._infer_feature_from_example(example)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/ray/session_2025-01-15_23-52-28_884602_1/runtime_resources/pip/31906f6f1883f7fb62d83ad16365f2311e2b5116/virtualenv/lib/python3.11/site-packages/evaluate/module.py", line 616, in _infer_feature_from_example
    raise ValueError(error_msg) from None
ValueError: Predictions and/or references don't match the expected format.
Expected format:
Feature option 0: {'predictions': Value(dtype='string', id='sequence'), 'references': Sequence(feature=Value(dtype='string', id='sequence'), length=-1, id='references')}
Feature option 1: {'predictions': Value(dtype='string', id='sequence'), 'references': Value(dtype='string', id='sequence')},
```

### Relevant log output

```shell

```

### Expected behavior

An evaluation job should get the right format in the input dataset regardless of any other consideration.

### System Info

N/A

javiermtorres:
# What's changing

Columns with empty values will be incorrectly identified as type `float64` by HF `datasets` library. This PR updates the type to `string`and converts `None` values to `''`.

Closes #646 

# How to test it

An automated case is provided. Some testing models will yield empty fields, causing this error.

N/A

# I already...

  - No DB migration required.

### DIFF
--- a/lumigator/jobs/evaluator/evaluator.py
+++ b/lumigator/jobs/evaluator/evaluator.py
@@ -5,6 +5,7 @@ from pathlib import Path
 import click
 import s3fs
 from datasets import load_from_disk
+from datasets.features.features import Value
 from eval_metrics import EvaluationMetrics
 from loguru import logger
 from utils import timer
@@ -76,7 +77,27 @@ def run_eval(config: EvalJobConfig) -> Path:
 
     # Load dataset given its URI
     dataset = load_from_disk(config.dataset.path)
+    # Cast non-string fields into strings. If it cannot be done, then
+    # the dataset is most probably broken.
+    # All-empty fields are interpreted by the loader as 'float64' by default,
+    # which causes issues when an empty 'predictions'
+    for fname in dataset.features:
+        feature = dataset.features[fname]
+        if not (isinstance(feature, Value) and feature.dtype == "string"):
+            logger.warning(
+                f"Found column '{fname}' with non-string type '{feature}', "
+                "converting type to string and None values to ''"
+            )
+            dataset = dataset.cast_column(fname, Value("string"))
+
     logger.info(f"Retrieving {config.dataset.path} for evaluation")
+
+    def replace_none_with_empty(row):
+        if row["predictions"] is None:
+            row["predictions"] = ""
+        return row
+
+    dataset = dataset.map(replace_none_with_empty)
 
     # Check for required fields
     # If any of the G-Eval metrics are in config.evaluation.metrics,
--- a/lumigator/backend/backend/tests/unit/api/routes/test_jobs.py
+++ b/lumigator/backend/backend/tests/unit/api/routes/test_jobs.py
@@ -138,6 +138,5 @@ def test_missing_api_key_in_job_creation(
     with patch("backend.api.deps.JobSubmissionClient"):
         with patch.object(job_service, "create_job", side_effect=SecretNotFoundError("test error msg")):
             response = app_client.post("/jobs/inference/", headers=POST_HEADER, json=infer_payload)
-            loguru.logger.critical(f"response text: {response.text}")
             assert response is not None
             assert response.status_code == status.HTTP_400_BAD_REQUEST
--- a/lumigator/jobs/evaluator/tests/test_eval_metrics.py
+++ b/lumigator/jobs/evaluator/tests/test_eval_metrics.py
@@ -1,6 +1,13 @@
+import shutil
+from pathlib import Path
+
 import numpy as np
 import pytest
+from datasets import load_dataset, load_from_disk
 from eval_metrics import EvaluationMetrics
+from evaluator import run_eval
+
+from schemas import DatasetConfig, EvalJobConfig, EvaluationConfig
 
 
 @pytest.fixture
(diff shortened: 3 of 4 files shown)


lumigator/jobs/evaluator/evaluator.py
```diff
@@ -76,8 +77,28 @@ def run_eval(config: EvalJobConfig) -> Path:
...
+            logger.warning(
+                f"Found column '{fname}' with non-string type '{feature}', "
+                "converting type to string and None values to ''"
+            )
+            dataset = dataset.cast_column(fname, Value("string"))
+
     logger.info(f"Retrieving {config.dataset.path} for evaluation")
 
+    def replace_none_with_empty(row):
+        if row["predictions"] is None:
```
HareeshBahuleyan, line 96: Would it be an issue hard-coding the column name "predictions" since this can be set by the user to a different name, or no?

lumigator/jobs/evaluator/evaluator.py
```diff
@@ -76,8 +77,28 @@ def run_eval(config: EvalJobConfig) -> Path:
...
+            row["predictions"] = ""
```
HareeshBahuleyan, line 97: What happens if examples or ground_truth is `None`? Do we get the same error?

Maybe `None` in examples or ground_truth would need to be handled differently - like should those instances be dropped from the dataset itself? WDYT? 
Its a different issue than what you are addressing here, so we can tackle it else where :)

lumigator/jobs/evaluator/evaluator.py
javiermtorres: Hmmm, I thought it was already hardcoded. If it can be set to another name, I'll use that instead of hardcoding, sure!

lumigator/jobs/evaluator/evaluator.py
javiermtorres: Well, all instances apparently need to be `None` for the field to be interpreted as `float64` instead of string. But I'm not so sure, so I'll make another test item and find it out.

> Maybe None in examples or ground_truth would need to be handled differently

You tell me :D I'm just ensuring that the job doesn't fail, but the input data may not be valid in the first place. What would be less surprising and more informative to the user?

lumigator/jobs/evaluator/evaluator.py
javiermtorres: Note that `"examples"` and `"ground_truth"` are still hardcoded.

HareeshBahuleyan: Hey @javiermtorres thanks for the clarification. I am trying to see whether this check would happen correctly incase the user overrides `output_field` in their job config. The default is `predictions` but I guess they can specify a different value. 
https://github.com/mozilla-ai/lumigator/blob/8f0a86af7d91efd138af9816732b0320bc8bd619/lumigator/schemas/lumigator_schemas/jobs.py#L146

I may be missing something, or let me know if these are unrelated 😅

HareeshBahuleyan: >I'll extract this to an optional field defaulted to "predictions"

Thanks!

lumigator/jobs/evaluator/evaluator.py
HareeshBahuleyan: Yep, lets create an issue to revisit later. Thank you

lumigator/jobs/evaluator/evaluator.py
javiermtorres: > I may be missing something, or let me know if these are unrelated 😅

_Somewhat_ unrelated, but relevant. You're pointing at an inference config, whereas I'm pointing at the evaluation config. Strictly speaking, this is configurable in inference and not evaluation, which would obviously cause a failure (so that's why I changed it to be configurable as well in evaluator 😅). So basically you're right 😋

HareeshBahuleyan: Got it, thanks for `config.predictions_field` - looks good !

PR #1269, opened 2025-03-21
Schema: Add schema util that allows merging of Pydantic schema models

peteski22:
# What's changing

Adds `merge_models` from `model_operations` as a schema util that allows you to merge two of the same Pydantic models (including a deep merge). 

It returns the resulting merged model ... along with some sets identifying:

* Overwritten keys that appeared in both models and the values were different
* Un-merged keys that only appeared in the base model and not the overlay (they are present in the final model)
* Skipped keys that existed in both, but already had the same value

Added tests to cover the functionality.

# How to test it

Unit tests in the PR cover testing.

This can be used to merge job results in the MLFlow tracking client when it attempts to generate the compiled results for a workflow.

# I already...

### DIFF
--- a/lumigator/schemas/lumigator_schemas/tests/conftest.py
+++ b/lumigator/schemas/lumigator_schemas/tests/conftest.py
@@ -1,0 +1,39 @@
+import json
+from pathlib import Path
+
+import pytest
+
+
+def load_json_from_file(file_path: Path) -> dict:
+    """Load JSON data from a file path and return it as a dictionary."""
+    with Path.open(file_path) as file:
+        return json.load(file)
+
+
+@pytest.fixture(scope="session")
+def resources_dir() -> Path:
+    return Path(__file__).parent / "data"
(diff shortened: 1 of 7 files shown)


njbrake:
Possibly you will hate this comment: At this stage of Lumigator, do we need advanced logic to handle the unexpected merge keys, as this code does? Or should we have a more "feature engineered" approach that says "these keys are ok to be in both, assert that they. have the same value so we know the overwrite is safe", otherwise through an exception? I expect there to be duplicate `prediction`, `examples`, and `ground_truth` columns, but nothing else. My gut reaction is that I would rather handle the keys we know are ok to be in both, and otherwise throw an exception. I am wary of having code that tries to keep things working instead of raising an error to tell us that something changed that we didn't expect.

Also possible that I'm confused here, please let me know if I am missing the point 😆

javiermtorres (approved):

PR #1285, opened 2025-03-25
Tracking client: update MLFlow run status when we update the status tag

peteski22:
# What's changing

* Updates the status of the run in MLFlow as well as the tag we create for status.

# How to test it

Steps to test the changes:

1.
2.
3.

This is the 'status' shown in MLFlow when we don't come back to update it at the end:

image

Now the status no longer says 'Running'

image

So we can see what was successful/failed in the MLFlow dashboard too:

image

# I already...

### DIFF
--- a/lumigator/backend/backend/services/workflows.py
+++ b/lumigator/backend/backend/services/workflows.py
@@ -91,16 +91,34 @@ class WorkflowService:
         """Handle a workflow failure by updating the workflow status and stopping any running jobs."""
         loguru.logger.error("Workflow failed: {} ... updating status and stopping jobs", workflow_id)
 
-        # Mark the workflow as failed.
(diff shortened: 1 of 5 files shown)


njbrake (approved):
LGTM! I'm not sure whether we set update_workflow_status to take advantage of those workflowstatus' that you add, but I trust that you're aware of the plan and this is a step in the process. Thanks :)

peteski22:
> LGTM! I'm not sure whether we set update_workflow_status to take advantage of those workflowstatus' that you add, but I trust that you're aware of the plan and this is a step in the process. Thanks :)

😄 I added some screenshot receipts to help explain that it's fixing (the Status not being updated in MLFlow), although we were tracking the 'status' as a tag we added manually, it now lines up.

PR #1321, opened 2025-04-09
1312 enable llm as judge for translation

### LINKED ISSUE #1312: Add deepeval criteria and evaluation steps for the translation task
Similarly to what we did for summarization, we should also add criteria and evaluation steps to perform llm-as-judge evaluation of a translation task.

At the moment there are many open questions about how this would work, e.g.:

- how well a multilingual LLM can evaluate the quality of a translation in a given language
- whether the prompt language affects the quality of the evaluation
- how good English evaluation is wrt other languages
- whether we should use different, language-specific models for the evaluation

The current task mostly focuses on (1) extending the current code so it will run llm-as-judge for translation and (2) provide a relatively simple way to play with different prompts, so we can evaluate how well this works.

aittalam:
# What's changing

- Added new metrics for g-eval-based evaluation of translations, both with and without relying on reference translations.
- Added unit test to address the case where a new metric is added to the code but not in the schemas (usually not raising error, but not returning any output either)
- Patched Makefile that was breaking on locally-run unit tests 

Closes #1312 

# How to test it

Given a dataset containing translations, run an evaluator job with `g_eval_translation` and/or `g_eval_translation_noref` as metric names. Optionally you can add an `llm_as_judge` section with parameters to connect to an ollama server as follows:

```
{
  "name": "eval_name",
  "dataset": "dataset_id",
  "job_config": {
    "secret_key_name": "<required if you use OAI>",
    "job_type": "evaluator",
    "metrics": [ "g_eval_translation", "g_eval_translation_noref" ],
    "llm_as_judge": {
      "model_name": "gemma3:27b",
      "model_base_url": "http://localhost:11434",
    }
  }
}
```

# I already...


lumigator/jobs/evaluator/evaluator.py
```diff
@@ -183,7 +183,7 @@ def sanitize_job_name(job_name: str) -> str:
     else:
         config = EvalJobConfig.model_validate_json(args.config)
         # NOTE: Temporary solution to avoid API key issues in G-Eval which defaults to calling OpenAI.
-        if "g_eval_summarization" in config.evaluation.metrics and (api_key := os.environ.get("api_key")):
+        if any(s.startswith("g_eval_") for s in config.evaluation.metrics) and (api_key := os.environ.get("api_key")):
             os.environ["OPENAI_API_KEY"] = api_key
```
ividal, line 187: Possible nit: #1230 gave us an OS alternative for GEval, which is an understandable choice for those who do not have or want an OpenAI key. So a few lines below, we're looking at LOCAL_MODEL_API_KEY to handle GEval model in that case. Are we making both point at the same key (if so, how and could we explain this in the doc)?

I feel this key handling could lead to a bit of confusion (but not sure, other than documenting how to proceed in each case openai/local) how to handle it.

lumigator/jobs/evaluator/tests/test_eval_metrics.py
```diff
@@ -80,3 +82,43 @@ def test_empty_fields_cast_as_float64():
...
+       name with a specific task (and possible extra params).
+    - `EvalJobMetrics` in `schemas` which specifies the format of the new evaluation
+       results and where they will appear in the resulting dictionary.
+
+    This test automatically verifies, for each metric called geval_*, that there is
+    actually an output with the same name inside the resulting dictionary.
+    """
+    # we cannot access _supported_metrics statically, so we instantiate
+    # the EvaluationMetrics class with a known metric
+    em = EvaluationMetrics(["g_eval_summarization"])
```
ividal, line 107: I may be missing something... why summarization instead of translation? Or was it a random choice, merely to be able to access supported metrics via any EM object?

ividal (commented):
Nice to have this in translation too! 🥳 🙌 
A question and a request for doc... mostly looking to reduce potential for confusion  😅

lumigator/jobs/evaluator/evaluator.py
aittalam: Great point! I agree this could raise some confusion. Despite thte fact that deepeval _could_ support other models, we have tested it with ollama alone and the way deepeval deals with it is that it [uses an `ollama` API key to specify it will use ollama](https://github.com/confident-ai/deepeval/blob/d805336d714bed66223a16a455042dcfe16b95de/deepeval/metrics/utils.py#L298)... 🤔 

So, one thing we could do is use only the secret_key_name parameter as follows:
- if no `llm_as_judge` section is present in the config, we assume we use the default deepeval model which is GPT so we expect the key to be OPENAI_API_KEY
- if the `llm_as_judge` section is present, we either pass the provided key (for servers which are not ollama, which we are not supporting rn but we could at least document it) or `ollama` if no key is present (so users can just configure their ollama server without having to know deepeval quirks).

WDYT?

I think in the longer term we might want to support API keys differently, as while we run a single model for inference the assumption we'll use a single one for eval is a bit too strong... We are already using different models now, we are just lucky they are not all APIs :-)

lumigator/jobs/evaluator/tests/test_eval_metrics.py
aittalam: The latter... It's a bit ugly, but that attribute holds information about the class methods and I could not make it static. I used a g_eval_* variable as it's a g_eval_* test, but it  could be meteor or anything else as the instance is thrown away immediately after getting the list of metrics.

lumigator/jobs/evaluator/evaluator.py
aittalam: I tried to implement the above in commit b9c1a852da1121864c8dae523a1822fcc406038d, lmk what you think 🙏

aittalam: Aaand added documentation in commit db4be72e2246f15ed98178f7f0390a90a045af0a :-)

HareeshBahuleyan (approved):
Thanks Davide for updating the PR to include `llm_as_judge` parameter to be available at the workflow level as well based on offline discussions! 
Tested that it works with the correct api keys when using a different model (mistral) for inference and open_ai for evaluation/llm-as-judge 🥳

